PR #60, opened 2021-10-18
PER-8094: Make zip icons less hideous

meisekimiu:
## Description
This PR makes ZIP icons less hideous. This is done in a similar way to how folder icons are displayed in the UI. Folders actually have a "true" thumbnail associated with them in the back end (these days it's just a weird grid of grey squares), but the web app ignores that thumbnail and instead displays the folder icon. This fix does the same for ZIP icons, using a much nicer vector icon instead of the compressed pixelized nightmare that's associated with the file in the backend.

## Steps to Test
- Look at how ZIP files look in the UI under various circumstances:
    - In list view
    - In grid view
    - in the sidebar editor
    - in the fullscreen editor
    - On a mobile browser

The exact styling of this is still under QA review, but I might as well open this for code review sooner rather than later.


meisekimiu:
@jasonaowen care to re-review the code real quick? I've added FontAwesome icons to be closer to the mockup design and that has a few small code changes.

cecilia-donnelly (approved):
Thanks @meisekimiu !  Makes sense to me.  I see that we still use the material icons in a lot of places (well, more specifically, I see that other stylesheets still have the material-icons class).  Is your plan to eventually replace those with FontAwesome everywhere?

meisekimiu:
I think right now the blend of icons we use is fine! I believe with the way I have installed FontAwesome, we're only importing the single icon we're using (I was previously worried about importing an entire icon library for just one icon). I think one reason why we wanted to go with the FontAwesome one was specifically so it'd look distinct compared to the other icons we're using so people wouldn't confuse ZIP files with folders.

PR #72, opened 2021-11-08
Disable Angular CLI Analytics

meisekimiu:
Our Angular configuration file had a string ID for it's `cli.analytics` property. I believe this means that analytics is enabled on this project and we might be sending anonymous usage data to Google every time we run the Angular CLI within this project. This PR makes sure we're not doing that, since it seems against the philosophy of Permanent... and is also just creepy.

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -206,6 +206,6 @@
     }
   },
   "cli": {
-    "analytics": "0f7fbf7e-c8eb-4198-b099-2c1d5308c974"
+    "analytics": false
   }
 }


jasonaowen (approved):
Thank you for finding and fixing this, @meisekimiu!

I didn't find much first party documentation on this setting, but did find [this SO answer](https://stackoverflow.com/a/60778004) which matches.

Testing now, but that'll take a few minutes, so in the meantime +1 and thank you again!

jasonaowen:
Aha, here is the [configuration documentation](https://angular.io/guide/workspace-config#cli-configuration-options):

Property | Description | Value Type
-- | -- | --
`analytics` | Share anonymous usage data with the Angular Team. | `boolean` \| `ci`

So yeah, exactly as you said, putting a UUID there was already incorrect but was probably evaluating to `true`.

jasonaowen:
I was able to successfully build the project on this branch, and saw nothing about analytics in either direction... just like previous builds. I am not set up to easily see all network traffic that happens during a build, so I can't confirm that no analytics are being sent, but given how strongly their docs emphasize the opt-in nature of these analytics, I am reasonably confident that with this change we are no longer opting in.

PR #76, opened 2021-12-01
PER-8591: Remove user context from Sentry events

meisekimiu:
## Description
This PR removes the code that adds user information (e-mail, account/archive numbers) to Sentry events. This means that no more personally identifying information is being attached to these events.

## Steps to Test
Since Sentry is disabled in local, the only way to test this is to deploy to another environment. This is a pretty simple code change so I'm confident that once it's reviewed we can merge it and check Sentry in the coming weeks to see that new events don't have user tags on them anymore.

(Otherwise we can deploy to dev and force an error there to test.)

### DIFF
--- a/src/app/shared/services/account/account.service.ts
+++ b/src/app/shared/services/account/account.service.ts
@@ -71,7 +71,7 @@ export class AccountService {
 
     // set account data on Sentry scope
     Sentry.configureScope(scope => {
-      scope.setUser({id: this.account.accountId, email: this.account.primaryEmail});
+      scope.setUser({id: this.account.accountId});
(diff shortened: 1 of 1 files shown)


jasonaowen:
Hey @meisekimiu, sorry for not documenting the decision about what to include and what to exclude better! I just commented on the ticket as well, but: let's please keep `archiveId` and `accountId`, but exclude email.

meisekimiu:
Updated branch to only remove e-mail now!

jasonaowen (approved):
Awesome, thank you @meisekimiu!

PR #77, opened 2021-12-09
Fix whitespace

slifty:
This PR fixes some tiny whitespace mistakes in our package.json because better now than later.

I should note: I didn't actually make these changes -- npm did when running `npm install`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,8 +1,8 @@
 {
   "name": "mdot",
   "version": "2.13.5",
-  "engines" : { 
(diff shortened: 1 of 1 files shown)


meisekimiu:
Somehow unit tests are broken? I don't think it is from this change (how would it even break the tests?), so while I think this is fine to merge we should investigate what broke in the tests.

meisekimiu (approved):
Unit test issues are resolved, going to go ahead and approve this!

PR #78, opened 2021-12-09
Unit test fixes

meisekimiu:
These fixes do a few things:
- Force the node version to 14 (due to Karma just not working in newer node versions)
- Disable loading the actual external Google Maps API in our tests.

This combination of fixes seems to get the unit tests to pass. The Google Maps failures were triggering errors on my local machine with a downgraded node version, but CI was continuing to have errors I wasn't detecting due to it also loading an incompatible version of node.

(The name of this branch hints at what I thought the initial solution was)

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -12,6 +12,10 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '14'
+
       - uses: actions/cache@v2
         with:
           path: node_modules
--- a/src/app/core/services/edit/edit.service.ts
+++ b/src/app/core/services/edit/edit.service.ts
@@ -112,6 +112,11 @@ export class EditService {
   }
 
   loadGoogleMapsApi() {
+    if (window['doNotLoadGoogleMapsAPI']) {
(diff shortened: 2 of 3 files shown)


meisekimiu:
When testing with Google Maps enabled, I get a bunch of warnings and error messages. I'm getting many variations of the following error:
```
ERROR LOG: 'Unhandled Promise rejection:', 'Cannot read properties of undefined (reading 'i')', '; Zone:', '<root>', '; Task:', 'Promise.then', '; Value:', TypeError: Cannot read properties of undefined (reading 'i')
TypeError: Cannot read properties of undefined (reading 'i')
    at https://maps.googleapis.com/maps/api/js?key={my local key}&callback=__gmapsLoaded&libraries=places:187:16
```

Which specifically points to the external Maps JS we're importing (should we be loading it in testing anyway?). I swear I remember seeing this in the Github Action logs as well, although in looking at everything now it does seem to just be related the node version error.

Also I get this error which shows that we're at least not handling the Maps JS properly in testing:
```
ERROR: 'You have included the Google Maps JavaScript API multiple times on this page. This may cause unexpected errors.'
```

jasonaowen (approved):
Ah, weird! Thanks for the context; I haven't seen those errors, but you're right that probably we don't need to load Maps during tests.

PR #81, opened 2021-12-21
PER-8672: Allow empty profile to be private.

meisekimiu:
## Description
There is an issue with setting a user's profile visibility when their archive's profile is mostly empty. This is because the profile's visibility status is actually bound to the various `ProfileItemVO`s that belong the that profile, using the `publicDT` field.

Essentially, the web-app goes through <del>all</del> <ins>some of</ins><sup> 1</sup> the `ProfileItemVO`s on an Archive and checks the `publicDT` field of each. If there is at least one non-null `publicDT` field, the profile is considered public. Otherwise, if they are all null, the profile is considered private. However, as a default case, if there are no profile fields at all<sup>1</sup> the profile is considered public. This is what is causing the current error; on an empty profile there are no `ProfileItemVO`s to set the `publicDT` field to null.

<sup>1: Any fields that are considered "ALWAYS_PUBLIC" are filtered out of these checks and are not part of this check process.</sup>

This PR will add a dummy `ProfileItemVO` with a null `publicDT` to the user's Archive in the event the previously mentioned error would occur, allowing them to save an empty profile as private. I've set this dummy `ProfileItemVO` to hopefully be recognizable and readable in the database in case we ever need to debug things further.

Ideally, this fix would involve restructuring the way we store a profile's public status on the back end, but for now this is a quick fix that takes advantage of the back end's... loose way of structuring archive profile information.

## Steps to Test
* Open the "Archive Profile" page on an archive with a completely empty profile (it doesn't matter if it's newly created or a pre-existing archive you just cleared out).
* Set the profile to private.
* Refresh the page and note how the profile remains set to private.
* Make sure nothing else is broken.
  * Technically, setting an empty profile to private shouldn't actually change how the Public Archive Profile page looks since... well the page is already empty.

Resolves PER-8672.

### DIFF
--- a/src/app/models/profile-item-vo.ts
+++ b/src/app/models/profile-item-vo.ts
@@ -18,7 +18,8 @@ export type FieldNameUI =
 'profile.location' |
 'profile.phone_nbr' |
 'profile.social_media' |
-'profile.timezone';
+'profile.timezone' |
+'profile.FORCE_PUBLIC_UPDATE';
 
 export type FieldNameUIShort =
 'basic' |
--- a/src/app/shared/services/profile/profile.service.ts
+++ b/src/app/shared/services/profile/profile.service.ts
@@ -237,7 +237,19 @@ export class ProfileService {
     });
 
     if (!minItems.length) {
-      return;
+      if (this.checkProfilePublic() === setPublic) {
+        return;
+      } else {
+        const currentArchive = this.account.getArchive();
+        const publicDummy: ProfileItemVOData = {
+          archiveId: currentArchive.archiveId,
+          fieldNameUI: 'profile.FORCE_PUBLIC_UPDATE',
+          string1: setPublic ? 'true' : 'false',
+          type: 'type.widget.string',
+          publicDT: setPublic ? new Date().toISOString() : null
+        };
+        minItems.push(publicDummy);
+      }
     }
 
     try {


jasonaowen:
> the profile's visibility status is actually bound to the various `ProfileItemVO`s that belong the that profile, using the `publicDT` field

Oof, this model... I knew it was bad on the back-end, but I hadn't realized how closely the front-end model matches.

> if there are no profile fields at all the profile is considered public

Is that just the case on the front-end? How does the back-end handle this? Without looking yet, I would naively hope that an unfilled archive profile defaults to private, and that clients cannot request details on private archives without the appropriate credentials & permissions.

Which I guess would mean that an unfilled profile cannot be made public... which doesn't seem like the worst idea, but is slightly surprising.

meisekimiu:
> Is that just the case on the front-end? How does the back-end handle this?

All of this code reading and setting profile visibility is specifically part of the front end. I'm not even sure if the back-end actually has the same notion of full archive profile visibility. Rather I think it only sees visibility on a field-by-field basis. I believe things on the back end are private by default though (which is almost the reverse of the front-end).

jasonaowen:
Hmm... how does this work? Do you have reproduction steps on how to set an archive as public? That is turning out to be more difficult for me to do than I expected.

jasonaowen:
Aha, I think the backend determines whether or not the archive is visible *at all* by if there is anything in the public workspace; then, the public/private profile preference affects the not-`ALWAYS_PUBLIC` items. Confusing, but I think I see how it got there.

And, in turn, why the bug at hand existed in the first place: the toggle is misleading; it does not determine whether or not the archive is visible, it determines if *some* of the details on the profile page are visible.

meisekimiu:
I can confirm that what you've described is how the system works and was also confusing for me!

cecilia-donnelly (approved):
Thank you for explaining this so clearly and for this workaround, @meisekimiu !  LGTM.  (Though you're right that we should come up with a better overall design for the profile items.)

PR #84, opened 2022-01-11
PER-8503: Preserve newlines in archive description

meisekimiu:
This PR preserves newlines in Description field of a Public Archive Profile. This is done through pure CSS. Later fixes may implement functionality to include links or use markdown. Also, the only other multiline profile field (The description to Milestones) does not need this fix since it already preserves newlines.

To test, just use newlines in an archive profile. This fix preserves newlines, but does not preserve extra spaces/indents, similar to how Markdown works.

Picture of the fix:


cecilia-donnelly (approved):
I was able to see the newlines preserved!  Thanks @meisekimiu.  This is a nice stopgap solution. :)

PR #89, opened 2022-02-02
Manage tags fixes

meisekimiu:
Commit 1 uncomments out the code that allows managers and owners to both view the manage tags page.

Commit 2 adds a small delay when refreshing the tags list on the archive settings page.

We may only want to merge commit 1.


mithuna:
With the changes in the commit1 of this PR, I was able to test https://github.com/PermanentOrg/back-end/pull/25. For anyone other than managers and owners, it was showing a no permission message. So this looks great.

cecilia-donnelly (approved):
I like it!  The archive settings modal behaved as expected for an owner, manager, and curator of an archive.

I didn't notice any slowness on resetting tags locally with the new setTimeout, so that's good.  I think this is fine and am going to approve; will comment if I manage to simulate some back end slowness.

PR #90, opened 2022-02-09
PER-8255: Change error message

cecilia-donnelly:
We use "connection error" as our default error message, which is
confusing to users.  Change the name of the variable to better reflect
the problem and the message to users similarly.

### DIFF
--- a/src/app/core/components/upload-progress/upload-progress.component.html
+++ b/src/app/core/components/upload-progress/upload-progress.component.html
@@ -16,10 +16,10 @@
         <strong>{{fileCount.completed}}</strong> of <strong>{{fileCount.total}}</strong>files uploaded 
     </div>
   </div>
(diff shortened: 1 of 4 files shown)


meisekimiu (approved):
Looks good! The JIRA ticket doesn't explain very much so please correct me here, but the idea is to show this message even when a connection error occurs, right? Otherwise, do we still want to keep the `ConnectionError` value in the `UploadSessionStatus` enum along with its error message?

cecilia-donnelly:
Yes, since the only specific error we look for is the out of storage error in `upload.session.ts`, right?  I didn't see anywhere that we differentiate between a connection error and any other error (e.g. a zero byte file being uploaded, which is how I tested this).  If I missed it, though, then I can definitely change this.

I originally opened the ticket in response to a concern from the experience side that users would see "connection error" and try to fix their connection (reasonable!) even though really something else was going on that made a file fail.

PR #94, opened 2022-02-15
Remove unused non-owner constant

jasonaowen:
As far as we can tell, this constant was never used. Remove it.

See also: https://github.com/PermanentOrg/back-end/pull/31

Thanks to @cecilia-donnelly for finding the original issue, and fixing it on the back-end!


cecilia-donnelly (approved):
w00t w00t! Thanks @jasonaowen for finishing this.

PR #96, opened 2022-02-16
PER-8463: Onboarding Wire Up

meisekimiu:
This PR hooks up the onboarding flow in two ways:
- Adds route guards and redirects to ensure that the onboarding flow is triggered when a user has no archives.
    - In addition, locks down the routing for the auth and onboarding sections of the app as well, so the user can never navigate to the wrong place.
- Hooks up additional API requests in the onboarding flow (specifically, getting archive member invitations).

Two small hacks are present right now:
- For testing purposes, I've added a feature where QA can add a query parameter to /auth/signup to always create an archive or never create an archive. The full a/b testing code will be handled in parallel with the next ticket.
- I wanted to show a popup message when there was an error accepting an archive invitation but... the Message Service and its component were difficult to get working and I didn't want to waste time on this for something that could be fixed in the next PR.

### DIFF
--- a/src/app/auth/components/login/login.component.ts
+++ b/src/app/auth/components/login/login.component.ts
@@ -71,10 +71,17 @@ export class LoginComponent implements OnInit {
             window.location.assign(`/app/public?cta=timeline`);
           }
         } else {
-          this.router.navigate(['/'], { queryParamsHandling: 'preserve'})
-            .then(() => {
-              this.message.showMessage(`Logged in as ${this.accountService.getAccount().primaryEmail}.`, 'success');
-            });
+          const archives = this.accountService.getArchives().filter(
+            (archive) => !archive.isPending()
+          );
+          if (archives.length > 0) {
+            this.router.navigate(['/'], { queryParamsHandling: 'preserve'})
+              .then(() => {
+                this.message.showMessage(`Logged in as ${this.accountService.getAccount().primaryEmail}.`, 'success');
+              });
+          } else {
+            this.router.navigate(['/app/onboarding']);
+          }
         }
       })
       .catch((response: AuthResponse) => {
--- a/src/app/auth/components/signup/signup.component.ts
+++ b/src/app/auth/components/signup/signup.component.ts
@@ -118,39 +118,7 @@ export class SignupComponent implements OnInit {
         } else {
           this.accountService.logIn(formValue.email, formValue.password, true, true)
             .then(() => {
-              this.message.showMessage(`Logged in as ${this.accountService.getAccount().primaryEmail}.`, 'success');
-
-              if (this.route.snapshot.queryParams.eventCategory) {
-                this.ga.sendEvent({
-                  hitType: 'event',
-                  eventCategory: this.route.snapshot.queryParams.eventCategory,
-                  eventAction: 'signup'
-                });
-              }
-
-              if (this.route.snapshot.queryParams.shareByUrl) {
-                this.router.navigate(['/share', this.route.snapshot.queryParams.shareByUrl]);
-              } else if (this.route.snapshot.queryParams.cta === 'timeline') {
-                if (this.device.isMobile() || !this.device.didOptOut()) {
-                  this.router.navigate(['/public'], { queryParams: { cta: 'timeline' }});
-                } else {
-                  window.location.assign(`/app/public?cta=timeline`);
-                }
-              } else if (!this.isForShareInvite) {
-                if (this.device.isMobile() || !this.device.didOptOut()) {
-                  this.router.navigate(['/']);
-                } else {
-                  window.location.assign('/app');
-                }
-              } else if (this.shareItemIsRecord) {
-                setTimeout(() => {
-                  this.router.navigate(['/shares', 'withme']);
-                }, 500);
-              } else {
-                setTimeout(() => {
-                  this.router.navigate(['/shares', 'withme', this.shareItem.archiveNbr, this.shareItem.folder_linkId]);
-                }, 500);
-              }
+              this.redirectUserFromSignup();
             });
         }
       })
@@ -158,5 +126,41 @@ export class SignupComponent implements OnInit {
         this.message.showError(response.getMessage(), true);
         this.waiting = false;
       });
+  }
+
+  public redirectUserFromSignup() {
+    this.message.showMessage(`Logged in as ${this.accountService.getAccount().primaryEmail}.`, 'success');
+
+    if (this.route.snapshot.queryParams.eventCategory) {
+      this.ga.sendEvent({
+        hitType: 'event',
+        eventCategory: this.route.snapshot.queryParams.eventCategory,
+        eventAction: 'signup'
+      });
+    }
+
+    if (this.route.snapshot.queryParams.shareByUrl) {
+      this.router.navigate(['/share', this.route.snapshot.queryParams.shareByUrl]);
+    } else if (this.route.snapshot.queryParams.cta === 'timeline') {
+      if (this.device.isMobile() || !this.device.didOptOut()) {
+        this.router.navigate(['/public'], { queryParams: { cta: 'timeline' }});
+      } else {
+        window.location.assign(`/app/public?cta=timeline`);
+      }
+    } else if (!this.isForShareInvite) {
+      if (this.device.isMobile() || !this.device.didOptOut()) {
+        this.router.navigate(['/app', 'onboarding']);
+      } else {
+        window.location.assign('/app/onboarding');
+      }
+    } else if (this.shareItemIsRecord) {
+      setTimeout(() => {
+        this.router.navigate(['/shares', 'withme']);
+      }, 500);
+    } else {
+      setTimeout(() => {
+        this.router.navigate(['/shares', 'withme', this.shareItem.archiveNbr, this.shareItem.folder_linkId]);
+      }, 500);
+    }
   }
 }
--- a/src/app/auth/guards/auth.guard.ts
+++ b/src/app/auth/guards/auth.guard.ts
@@ -1,0 +1,35 @@
+import { Injectable } from '@angular/core';
+import { CanActivate, ActivatedRouteSnapshot, Router, RouterStateSnapshot, UrlTree } from '@angular/router';
+import { Observable } from 'rxjs';
+
+import { AccountService } from '@shared/services/account/account.service';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class AuthGuard implements CanActivate {
+  constructor(
+    private account: AccountService,
+    private router: Router,
+  ) {}
+
+  canActivate(
+    route: ActivatedRouteSnapshot,
+    state: RouterStateSnapshot): Observable<boolean | UrlTree> | Promise<boolean | UrlTree> | boolean | UrlTree {
+    if (this.account.getAccount()?.accountId) {
+      this.account.refreshArchives().then((archives) => {
+        const ownArchives = archives.filter(
+          (archive) => !archive.isPending()
+        );
+        if (ownArchives.length > 0) {
+          this.router.navigate(['/app/myfiles']);
+        } else {
+          this.router.navigate(['/app/onboarding']);
+        }
+      });
+      return false;
+    }
+    return true;
+  }
+
+}
--- a/src/app/auth/guards/auth.guard.ts
+++ b/src/app/auth/guards/auth.guard.ts
@@ -15,19 +15,15 @@ export class AuthGuard implements CanActivate {
 
   canActivate(
     route: ActivatedRouteSnapshot,
-    state: RouterStateSnapshot): Observable<boolean | UrlTree> | Promise<boolean | UrlTree> | boolean | UrlTree {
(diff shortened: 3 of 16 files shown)


cecilia-donnelly:
Code makes sense to me, @meisekimiu !

Things I'm going to test:
- Signing up with the new flow
- Signing up with the old flow
- Trying to navigate to log in page while logged in
- Signing up after being invited to be a member of an archive
- Signing up after being invited to join Permanent (but not being invited to a specific archive)

cecilia-donnelly (approved):
I tested all the cases I could think of and they all look awesome!  Way to go!  Do you want to remove the debugging page in this PR or wait to do that with the styling work?

I notice that you're checking whether a person has non-pending archives in a few different places.  Is there a way to refactor that or would that be Way Too Hard?  This might really be a question about why we need a specific guard for onboarding as opposed to the regular AuthGuard.  What's going on there?

meisekimiu:
The `OnboardingAuthGuard` is basically acting as the opposite of the `AuthGuard` in Auth. It specifically makes sure that you can only access the Onboarding area if you are supposed to.

I think there could possibly be a functional way of refactoring out the shared functionality though.

src/app/auth/guards/auth.guard.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  canActivate(
+    route: ActivatedRouteSnapshot,
+    state: RouterStateSnapshot): Observable<boolean | UrlTree> | Promise<boolean | UrlTree> | boolean | UrlTree {
+    if (this.account.getAccount()?.accountId) {
+      this.account.refreshArchives().then((archives) => {
+        const ownArchives = archives.filter(
+          (archive) => !archive.isPending()
+        );
+        if (ownArchives.length > 0) {
```
jasonaowen, line 24: Ah, this is the second time I've seen this snippet of code! It appears to be used for the same purpose; maybe my above comment about extracting a helper should lead to a more widely-visible helper than I was originally imagining.

src/app/onboarding/components/onboarding/onboarding.component.html
```diff
@@ -4,11 +4,14 @@
   </div>
 </div>
 <pr-debugger [currentScreen]="screen" (setState)="setState($event)"></pr-debugger>
-<div class="onboarding">
+<div class="onboarding" *ngIf="!skipOnboarding ; else loading">
```
jasonaowen, line 7: I find that positively-worded boolean variable names are much easier to read and reason about than negatively-worded ones, particularly when we then need to negate it to use! So maybe: `showOnboarding` instead?

src/app/onboarding/components/onboarding/onboarding.component.ts
```diff
@@ -43,6 +67,9 @@ export class OnboardingComponent implements OnInit {
     if (correspondingRoute) {
       this.location.go('app/onboarding/' + correspondingRoute.path);
     }
+    if (screen === OnboardingScreen.done) {
+      this.router.navigate(['/app']);
```
jasonaowen, line 71: What is the difference between `location.go` and `router.navigate`?

src/app/onboarding/guards/onboarding.auth.guard.ts
```diff
@@ -0,0 +1,35 @@
...
+})
+export class OnboardingAuthGuard implements CanActivate {
+  constructor(
+    private account: AccountService,
+    private router: Router,
+  ) {}
+
+  canActivate(
+    route: ActivatedRouteSnapshot,
+    state: RouterStateSnapshot): Observable<boolean | UrlTree> | Promise<boolean | UrlTree> | boolean | UrlTree {
```
jasonaowen, line 18: The other advantage to wrapping the closing paren, return type, and opening brace onto its own line is it serves as a clearer separator between the parameter list and the body of the function.

jasonaowen (commented):
I have a bunch of beginner questions! Wanted to get these out before I start testing.

src/app/auth/components/signup/signup.component.ts
```diff
@@ -159,4 +127,40 @@ export class SignupComponent implements OnInit {
...
+    }
+
+    if (this.route.snapshot.queryParams.shareByUrl) {
+      this.router.navigate(['/share', this.route.snapshot.queryParams.shareByUrl]);
+    } else if (this.route.snapshot.queryParams.cta === 'timeline') {
+      if (this.device.isMobile() || !this.device.didOptOut()) {
+        this.router.navigate(['/public'], { queryParams: { cta: 'timeline' }});
+      } else {
+        window.location.assign(`/app/public?cta=timeline`);
+      }
```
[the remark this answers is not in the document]
meisekimiu: I asked about these last week, and it seems like both the Beta Participation and Timeline CTA functionality are currently unused and can probably be deleted. I just decided to keep them in for now and hope that we can delete it some other time.

src/app/auth/guards/auth.guard.ts
```diff
@@ -0,0 +1,35 @@
...
+})
+export class AuthGuard implements CanActivate {
+  constructor(
+    private account: AccountService,
+    private router: Router,
+  ) {}
+
+  canActivate(
+    route: ActivatedRouteSnapshot,
+    state: RouterStateSnapshot): Observable<boolean | UrlTree> | Promise<boolean | UrlTree> | boolean | UrlTree {
```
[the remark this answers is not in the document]
meisekimiu: This is how `ng generate` made the code, so that's why I just ignored any formatting issues, but I can split these lines.

src/app/auth/guards/auth.guard.ts
```diff
@@ -0,0 +1,35 @@
...
+    if (this.account.getAccount()?.accountId) {
+      this.account.refreshArchives().then((archives) => {
+        const ownArchives = archives.filter(
+          (archive) => !archive.isPending()
+        );
+        if (ownArchives.length > 0) {
+          this.router.navigate(['/app/myfiles']);
+        } else {
+          this.router.navigate(['/app/onboarding']);
```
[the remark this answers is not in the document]
meisekimiu: I mainly just copied the approach that the standard `AuthGuard` we have in the `CoreModule` works. The UrlTree might sound like a better solution though!

src/app/onboarding/components/onboarding/onboarding.component.ts
meisekimiu: I believe `location.go` is some kind of `history.pushState` type thing where as `router.navigate` is a more "hard" navigation... if that makes sense?

jasonaowen: Ah, I'm afraid it does not. :S I might be missing some background!

When would you use one or the other? Or, how concerned should I be that we're using multiple different types of navigation?

I don't think this is a blocker for deploying this, but I would like to understand!

src/app/auth/components/signup/signup.component.ts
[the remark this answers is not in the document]
jasonaowen: Yeah of course, definitely not suggesting changes here! Just getting the lay of the land.

src/app/onboarding/components/onboarding/onboarding.component.ts
meisekimiu: `location.go` sets the URL in the browser window without actually refreshing the page, so we can keep the current component state intact while we go through the different onboarding screens. `router.navigate` actually causes a real browser navigation event to happen, which causes the page to reload. I believe this is the difference between the two calls.

meisekimiu:
Oh yeah, I definitely have realized I've trapped myself on the onboarding page and needed to delete cookies a few times over. I think I'll add it as part of the UI work!

src/app/auth/guards/auth.guard.ts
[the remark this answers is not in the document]
meisekimiu: Update on this is that UrlTree does work better! Though to create `UrlTree` types you need to actually still depend on the `Router` dependency, just not calling the `navigate()` method. Still going to commit it as part of the Guard refactoring.

src/app/onboarding/components/onboarding/onboarding.component.ts
jasonaowen: Aha! That is excellent context, thank you.

So here, after we finish onboarding, we want the page to reload so that the app asks for the default archive and populates all the state it needs as on first login?

src/app/auth/guards/auth.guard.ts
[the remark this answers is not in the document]
meisekimiu: It's definitely simpler, clearer code. This might just be my mind placebo-ing things, but the actual in-browser redirects feel cleaner as well.

src/app/onboarding/components/onboarding/onboarding.component.ts
meisekimiu: Yes! Also `location.go` doesn't actually cause Angular to reload any modules. Just as a sanity check I replaced `router.navigate` with `location.go` and while it set the URL correctly, I'm still stuck on the onboarding page.

Note the browser URL:

jasonaowen: Awesome - thanks again for explaining!

src/app/auth/guards/auth.guard.ts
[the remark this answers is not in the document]
jasonaowen: Thank you!

src/app/auth/components/login/login.component.ts
```diff
@@ -71,10 +71,17 @@ export class LoginComponent implements OnInit {
...
           }
         } else {
-          this.router.navigate(['/'], { queryParamsHandling: 'preserve'})
-            .then(() => {
-              this.message.showMessage(`Logged in as ${this.accountService.getAccount().primaryEmail}.`, 'success');
-            });
+          const archives = this.accountService.getArchives().filter(
+            (archive) => !archive.isPending()
+          );
+          if (archives.length > 0) {
```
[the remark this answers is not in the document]
jasonaowen: Should this use the new `hasOwnArchives` function, too? or is not having a `this.account` make that complicated enough to not be worthwhile?

jasonaowen (approved):
Thanks for the refactoring, @meisekimiu! This looks good to me - anything you're worried about?

src/app/core/guards/auth.guard.ts
```diff
@@ -18,7 +18,9 @@ export class AuthGuard implements CanActivate {
       if (isSessionValid && this.account.isLoggedIn()) {
         return true;
       } else {
-        this.account.refreshAccount();
+        if (isSessionValid !== this.account.isLoggedIn()) {
+          this.account.clear();
```
jasonaowen, line 22: Why do this only conditionally?

meisekimiu: The logic here is only to clear out the account if it's in the invalid state though I realize now it probably isn't necessary since they're logged out otherwise.

jasonaowen: Yep, that was what I was getting at! :slightly_smiling_face: It looks like `isLoggedIn` doesn't actually make a network call, which is good, but if we're gonna log them out might as well do so regardless, right?

src/app/onboarding/components/onboarding/onboarding.component.ts
```diff
@@ -74,8 +74,13 @@ export class OnboardingComponent implements OnInit {
 
   public setNewArchive(archive: ArchiveVO): void {
     this.currentArchive = archive;
-    this.account.setArchive(archive);
-    this.setScreen(OnboardingScreen.done);
+    const updateAccount = new AccountVO({defaultArchiveId: archive.archiveId});
+    this.account.updateAccount(updateAccount).then(() => {
+      this.account.setArchive(archive);
+      this.api.archive.change(archive).then(() => {
+        this.setScreen(OnboardingScreen.done);
```
jasonaowen, line 81: Ahh, I think I see! Let me make sure I understand the problem this is fixing:

- Archive-invited accounts have their default archive set to the pending invited archive
- Creating a new archive as part of the new onboarding flow did not update the default archive
- We were previously calling `setArchive` during onboarding, which meant that things worked for a little while - the duration of the session, basically
- But, if they logged out and back in, they were stuck with a default archive they couldn't (yet) access, and a full non-default archive prevented the onboarding from rerunning

Is that an accurate summary, @meisekimiu?

So the fix is to update the default after creating the new archive. Simple enough!

meisekimiu: I'm also calling `api.archive.change` here because while `setArchive` informs the `AccountService` what the current archive is, we still need to actually change to the new default archive on the back end as well (otherwise the back end still tries to fetch the pending archive).

(`account.setArchive` sets the Archive in the `AccountService` which for some reason is not automatically set by the API request)

jasonaowen: Oh ugh, thanks I missed that! It is always, always, always the things I think I understand but actually don't that get me; I didn't understand the distinction between `account.setArchive` and `api.archive.change`. I appreciate the explanation!

PR #98, opened 2022-02-23
Onboarding A/B testing

jasonaowen:
Set up A/B testing for new signups to use the new onboarding workflow, and set the random chance of so doing to 0 for now. Move the hacky query parameter check into the main signup component, and set all the secondary signup components to opt out of the new workflow.

I'm currently targeting PR #96 with this PR, but I believe if that gets merged to `main` the base branch of this PR will automatically update. If this PR is approved before #96 is merged, then I'll likely push my changes to that branch so as to avoid a merge conflict; let me know if you want to do something different, @meisekimiu!

I have not yet tested this extensively, but I wanted to get it up for review soonest.

[PER-8725 Start A/B testing signups](https://permanent.atlassian.net/browse/PER-8725)

### DIFF
--- a/src/app/shared/services/api/account.repo.ts
+++ b/src/app/shared/services/api/account.repo.ts
@@ -35,7 +35,11 @@ export class AccountRepo extends BaseRepo {
       passwordVerify: passwordConfirm
     });
 
-    const data = [{ AccountVO: accountVO, AccountPasswordVO: accountPasswordVO, SimpleVO: null }];
+    const data = [{
+      AccountVO: accountVO,
+      AccountPasswordVO: accountPasswordVO,
+      SimpleVO: null,
+    }];
 
     // HACK: This should be replaced with a function parameter
     if (window.location.search.includes('createArchive')) {
--- a/src/app/shared/services/api/account.repo.ts
+++ b/src/app/shared/services/api/account.repo.ts
@@ -20,6 +20,7 @@ export class AccountRepo extends BaseRepo {
     optIn: boolean,
     phone: string,
     inviteCode: string,
+    createDefaultArchive: boolean,
   ) {
     const accountVO = new AccountVO({
       primaryEmail: email,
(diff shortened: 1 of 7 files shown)


jasonaowen:
> I'm currently targeting PR https://github.com/PermanentOrg/web-app/pull/96 with this PR

Hm, seems we currently have CI configured to only run on pull requests that target `main`. Manually running for now: https://github.com/PermanentOrg/web-app/runs/5310008310?check_suite_focus=true

jasonaowen:
Which have [failed](https://github.com/PermanentOrg/web-app/runs/5310008310?check_suite_focus=true#step:6:1335) with what appears to be an unrelated error:

```
1) can cancel out of renaming a tag
     ManageTagsComponent #manage-tags
     Error: Uncaught (in promise): Error: Manage Tags: Error accessing delete endpoint
```

Any insight, @meisekimiu? I'm not entirely sure where to start with that.

src/app/shared/services/api/account.repo.ts
```diff
@@ -29,17 +36,14 @@ export class AccountRepo extends BaseRepo {
...
-    // HACK: This should be replaced with a function parameter
-    if (window.location.search.includes('createArchive')) {
-      data[0].SimpleVO = new SimpleVO({key: 'createArchive', value: true});
-    } else if (window.location.search.includes('noArchive')) {
-      data[0].SimpleVO = new SimpleVO({key: 'createArchive', value: false});
-    }
+    const data = [{
+      AccountVO: accountVO,
+      AccountPasswordVO:
+      accountPasswordVO,
```
meisekimiu, line 42: Extreme nitpick here, but why is this broken into two lines, when the other key / value pairs here are all on one line?
```suggestion
      AccountPasswordVO: accountPasswordVO,
```

meisekimiu (approved):
This looks good, and is exactly what I was imagining the implementation would look like! Thanks for also adding the edge-cases for the various other routes to signup.

As for the failing test cases, I believe that's something that somehow got past some of the onboarding changes, maybe? I think I have it fixed in my styling branch.

src/app/shared/services/api/account.repo.ts
jasonaowen: Because I wasn't paying enough attention and made a mistake! Thank you for catching it.

PR #99, opened 2022-02-23
PER-8790: Style Onboarding Page

meisekimiu:
This is the Onboarding Style Change PR!

Notes:
- Debugging component is still here for initial PR review so we can test things faster. It's been moved to the bottom of the screen now to be less disruptive. Going to delete before final merge.
- Wrote brand new test suite for the Create Archive section of the onboarding screen. Initially I extracted out the main app's "create archive" screen into its own component, but because of style changes it wasn't really reusable anymore so I had to add that functionality back into the main component.
- I did not actually end up using the MessageService due to me not being able to get Angular liking that (for some reason the MessageService and MessageComponent are dependent on each other but are split across multiple modules in a very annoying way). I ended up just emulating the functionality message component to show error messages to users on this screen.
- This has fixes to the broken unit tests in #98... I think?


meisekimiu:
Github isn't showing it, but unit tests do indeed pass: https://github.com/PermanentOrg/web-app/runs/5311409934

jasonaowen:
Looking now, @meisekimiu!

Would you mind pointing out the test fix? I don't immediately see anything obvious.

meisekimiu:
In `onboarding.spec.ts`, the mocked API/ArchiveServices did not have all the methods being called by the component, and that was erroring out. I'm not sure how it passed tests to begin with, but when I started on this branch that was causing tests to fail until I fixed those lines.

src/app/shared/components/archive-small/archive-small.component.ts
```diff
@@ -29,10 +29,13 @@ export class ArchiveSmallComponent implements OnInit, OnChanges {
   @Output() editClick = new EventEmitter<any>();
 
   @Input() showAccept = false;
+  @Input() showAcceptMobile = false;
   @Input() acceptText = 'Accept';
   @Input() acceptIcon = 'check';
   @Output() acceptClick = new EventEmitter<any>();
 
+  @Input() showDefault = false;
```
jasonaowen, line 37: This is a change in behavior, right? If I understand correctly, the default should either be `true`, or we should update the non-onboarding call sites of this component to pass `showDefault = true`.

src/app/onboarding/components/create-new-archive/create-new-archive.component.html
```diff
@@ -1,7 +1,91 @@
...
+            <span class="archive-type-description">an organizational entity, such as a company or nonprofit</span>
+          </div>
+        </ng-container>
+      </ng-container>
+    </div>
+    <p><strong>What is your new archive called?</strong> This is generally the legal or official name of the person, family, or organization the archive is about &mdash; for example, “The Katherine Johnson Archive.”</p>
+    <p>You can always edit your archive’s name later.</p>
+    <div class="name-field">
+      <label for="archive-name">The</label>
+      <input type="text" id="archive-name" class="form-control" (input)="archiveName = $event?.target?.value" [value]="archiveName" [placeholder]="getArchiveNamePlaceholder()" />
```
jasonaowen, line 82: Can we use the [HTML `required` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/required)? That'd allow us to lean on the browser for validation, and I believe that instead of needing the input event we could read the value during `onSubmit`.

src/app/onboarding/components/create-new-archive/create-new-archive.component.scss
```diff
@@ -0,0 +1,124 @@
+@import '../welcome-screen/welcome-screen.component.scss';
+
+.two-buttons {
+  button {
+    @media only screen and (max-width: 950px) {
```
jasonaowen, line 5: I believe SASS supports developer-defined [variables](https://sass-lang.com/documentation/variables); `950px` seems like a good candidate for such a variable, as is maybe this entire `@media only screen and (max-width: 950px)`, since we use it half a dozen times in this file.

src/app/onboarding/components/create-new-archive/create-new-archive.component.spec.ts
```diff
@@ -4,25 +4,86 @@ import { CreateNewArchiveComponent } from './create-new-archive.component';
...
+let createdArchive: ArchiveVO | null;
+const mockApiService = {
+  archive: {
+    create: async (a: ArchiveVO) => {
+      calledCreate = true;
+      createdArchive = a;
+      return {
+        getArchiveVO: () => {
+          return a;
+        }
```
jasonaowen, line 19: This can be inlined:

```suggestion
        getArchiveVO: () => a,
```

jasonaowen (commented):
First set of comments, before I take a break!

The progress bar appears not to be rendering - it's showing up in the dev tools as having a zero height. I haven't been able to use the dev tools to force it to be visible yet.

Screen Shot 2022-02-23 at 16 51 53

meisekimiu:
Addressed most of the issues so far! The only one I don't understand is the progress bar thing. I think this might be `ngClass` failing to even assign the proper classes to the progress bar? I've tested in both Firefox and Chromium, so I'm not entirely sure what's happening here.

I'm also going to be working on dialogs that will display after the onboarding process in the main app.

meisekimiu:
Realized I left a pretty big issue in the native form validation, working on fixing that now!

src/app/core/components/manage-tags/manage-tags.component.spec.ts
```diff
@@ -117,7 +117,8 @@ describe('ManageTagsComponent #manage-tags', () => {
   });
 
   it('should not delete a tag if an error happens', async () => {
-    const { find, fixture, outputs } = await defaultRender();
+    const { find, fixture, instance, outputs } = await defaultRender();
+    instance.throwErrors = false;
```
jasonaowen, line 121: This wasn't the test that failed on #98; that was `can cancel out of renaming a tag`, below.

...which, now that I think about it, is kind of multiply concerning - isn't part of the point of that test that the delete endpoint does not get called?

jasonaowen (commented):
I'm not sure I follow the manage tags test fix, and I'm not sure that fix should be in this PR. Let's defer that for now, accept that we might have to merge without a fully passing test suite (or might have to rerun the tests a few times, since it appears to be inconsistently passing), and focus on onboarding!

src/app/core/components/manage-tags/manage-tags.component.ts
```diff
@@ -53,7 +54,9 @@ export class ManageTagsComponent implements OnInit {
       }).catch(() => {
         // Let's add the tag back to UI to show it isn't deleted.
         this.tags.push(tag);
-        throw new Error('Manage Tags: Error accessing delete endpoint');
+        if (this.throwErrors) {
```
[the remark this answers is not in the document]
meisekimiu: For some reason that's still not working? Karma still fails the test when an error is thrown even when passing it in a function and expecting it to throw.

meisekimiu:
Pushed change with new dialogs. Right now there is a bit of code duplication between two dialogs, which ideally would not happen and can be cleaned up a bit.

I specifically made the `welcome` and `welcome-invitation` dialogs separate so I could bind each to separate routes with the way the dialog system works in our application.

jasonaowen:
[Manage tags test failed](https://github.com/PermanentOrg/web-app/runs/5324217401?check_suite_focus=true#step:6:1225), as expected:

```
1) should be able to delete a tag
     ManageTagsComponent #manage-tags
     Error: Uncaught (in promise): Error: Manage Tags: Error accessing delete endpoint
```

Rerunning in case that helps, but as I said previously I'm fine with ignoring that test failure in this PR.

src/app/onboarding/components/create-new-archive/create-new-archive.component.html
```diff
@@ -1,7 +1,99 @@
...
+      </label>
+
+      <label>
+        <div [ngClass]="getArchiveTypeClasses('type.archive.family')">
+          <img src="assets/svg/onboarding/archive-group.svg" alt="Group Archive" />
+          <div>
+            <strong class="archive-type-name">Group Archive</strong>
+            <span class="archive-type-description">any group of people, such as a community group or a family</span>
+          </div>
+          <input type="radio" name="archive-type" value="type.archive.family" (change)="setArchiveType($event.target.value)" />
```
jasonaowen, line 30: Should these `<input type="radio">`s have a CSS class of some sort? I would expect them to be hidden (`display: none;`), but they're showing up for me in Firefox:

Screen Shot 2022-02-24 at 12 01 31

They are properly enclosed in `<label>`s, so I can click on the pretty pictures to switch between them, which is great!

The Figma illustration has the containing elements update the background to indicate the selected option. If that's too complicated, maybe we can center them (which is always trivial to do, right?~) for a first draft, and plan to implement the shading later?

src/app/onboarding/components/create-new-archive/create-new-archive.component.scss
```diff
@@ -0,0 +1,130 @@
...
+}
+
+.archive-type-name {
+  margin-bottom: 7px;
+  align-self: center;
+}
+
+.archive-type-description {
+  color: #646464;
+}
```
jasonaowen, line 83: I think we want to make sure to visually separate this from the name!

```suggestion
.archive-type-description {
  color: #646464;
  display: block;
}
```

Otherwise it renders all smushed together:

Screen Shot 2022-02-24 at 12 09 38

src/app/onboarding/components/create-new-archive/create-new-archive.component.scss
```diff
@@ -0,0 +1,130 @@
...
+  }
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  align-items: stretch;
+}
+
+.archive-type-name {
+  margin-bottom: 7px;
+  align-self: center;
```
jasonaowen, line 78: My Firefox dev tools say:

> `align-self` has no effect on this element since it’s not a grid or flex item.
>
> Try adding `display:grid`, `display:flex`, `display:inline-grid`, or `display:inline-flex` to the element’s parent. [Learn more](https://developer.mozilla.org/en-US/docs/Web/CSS/align-self)

jasonaowen (commented):
Getting a few review notes out, but I'm still going!

src/app/onboarding/components/create-new-archive/create-new-archive.component.scss
jasonaowen: Oh, actually, now that I look at the next page which renders correctly, this might just be the same problem as the `align-self` comment below: I think they need to be in a flex container!

src/app/onboarding/components/create-new-archive/create-new-archive.component.html
meisekimiu: Okay I have to look into what is actually going on, because I have style rules for hiding the inputs, updating background, centering things, and dividing everything properly but they're just... not showing up on your end.

For the record, here is what I see:

src/app/onboarding/components/welcome-screen/welcome-screen.component.html
```diff
@@ -1,16 +1,32 @@
...
+      [showDefault]="false"
+      (acceptClick)="acceptPendingArchive(archive)"
+      [showRemove]="false"></pr-archive-small>
+      <br />
+      <br />
+    </div>
+
+    <div class="create-archive">
+      <a href="javascript:;" (click)="goToScreen(OnboardingScreen.newArchive)">Create a new archive</a>
+    </div>
```
jasonaowen, line 30: I don't see any corresponding CSS for `.create-archive`, so this is rendering pretty plainly.

src/app/onboarding/components/create-new-archive/create-new-archive.component.html
jasonaowen: Oh no! Then all these comments are super noisy, sorry!

jasonaowen: Resolving this; more in a main-thread comment to come.

jasonaowen:
It seems that `npm run build:local` is subtly broken, and after an hour of poking at it, @meisekimiu and I haven't been able to figure out how or why. The way it manifests is that `[ngClass]` assignments seem to just not work, so critical styling is lost. It works through the proxy with `npm run dev`, and it works in the dev environment with `npm run build:dev` - as made clear by the fact it works and looks great in dev! Also puzzling: other, existing usages of `[ngClass]` appear to work as expected.

I ran both `npm run build:local` and `npm run build:dev`, and compared the outputs. All the chunks are the same, except of course the `main` chunk:

```diff
< const i={production:!1,apiUrl:"https://local.permanent.org/api",hmr:!1,firebase:{authDomain:"prpledgedev.firebaseapp.com",databaseURL:"https://prpledgedev.firebaseio.com",functionsURL:"https://us-central1-prpledgedev.cloudfunctions.net",projectId:"prpledgedev"},debug:!1,release:n("kiQV").a,environment:"local"}},
> const i={production:!0,apiUrl:"https://dev.permanent.org/api",hmr:!1,firebase:{authDomain:"prpledgedev.firebaseapp.com",databaseURL:"https://prpledgedev.firebaseio.com",functionsURL:"https://us-central1-prpledgedev.cloudfunctions.net",projectId:"prpledgedev"},debug:!1,release:n("kiQV").a,environment:"dev"}},
```

The obvious differences there are `production` mode, plus the expected base URL and environment. But as @meisekimiu pointed out, `npm run dev` also runs with `production` mode off.

I am concerned that the behavior I'm seeing in local will also occur in staging or prod. With deadlines approaching and no leads at the moment, and after discussing with @meisekimiu, we're going to press on. New users won't see the new onboarding workflow until #100 is deployed, which would be next week at the soonest, so we can and will test in production. If it works as expected, we'll follow up on this build issue next week; if it does not, we'll scrub the deploy and go into RootsTech with the old onboarding workflow.

src/app/onboarding/components/welcome-screen/welcome-screen.component.html
jasonaowen: I checked on dev, and this is not just my broken build! :sweat_smile: 

Screen Shot 2022-02-24 at 14 11 11

meisekimiu: That link is plainly styled on the Figma designs. I think because it's a more "minor" action compared to accepting the invitations. In the Figma, there's also an "Accept All" button to direclty contrast with it, which I believe we've agreed is not part of the MVP for this flow.

src/app/onboarding/components/welcome-screen/welcome-screen.component.html
```diff
@@ -1,16 +1,32 @@
...
+    <div class="pending-archive-list" style="margin: 0 auto;">
+      <strong>Pending invitations:</strong>
+      <pr-archive-small *ngFor="let archive of pendingArchives"
+      [largeOnDesktop]="true" [archive]="archive"
+      [accessRole]="archive.accessRole"
+      [showAccept]="true"
+      [showAcceptMobile]="true"
+      [showDefault]="false"
+      (acceptClick)="acceptPendingArchive(archive)"
+      [showRemove]="false"></pr-archive-small>
```
jasonaowen, line 23: Same thing about formatting, if you don't mind:

```suggestion
      <pr-archive-small *ngFor="let archive of pendingArchives"
        [largeOnDesktop]="true"
        [archive]="archive"
        [accessRole]="archive.accessRole"
        [showAccept]="true"
        [showAcceptMobile]="true"
        [showDefault]="false"
        (acceptClick)="acceptPendingArchive(archive)"
        [showRemove]="false"
      ></pr-archive-small>
```

src/app/onboarding/components/welcome-screen/welcome-screen.component.html
jasonaowen: Oh! So it is. Sorry, and nevermind, then!

jasonaowen (approved):
This is incredible work, @meisekimiu! Well done; the onboarding experience looks gorgeous and works great. The code all makes sense to me, and it's clean and well designed. Thank you so much!

src/app/onboarding/components/onboarding/onboarding.component.html
```diff
@@ -36,4 +36,3 @@
 <ng-template #loading>
     <div class="loading">Loading...</div>
 </ng-template>
-<pr-debugger [currentScreen]="screen" (setState)="setState($event)"></pr-debugger>
```
jasonaowen: Goodbye, debugger. We appreciate your contributions!

jasonaowen (approved):

jasonaowen:
> It seems that `npm run build:local` is subtly broken, ...

Opened a ticket for us to follow up on the build issue, which appears to not affect dev, staging, or prod: [PER-8796
Fix web-app local build issue](https://permanent.atlassian.net/browse/PER-8796).

PR #103, opened 2022-03-02
Allow users to verify contact info while signed in

jasonaowen:
The `VerifyComponent` is used both during the sign-up process for multi-factor authentication, and for verifying a user's email address and phone number from their account page.

In #96, we blocked signed-in users from accessing auth routes, in order to prevent them from being faced with a sign in page unnecessarily.

Move the verify routes out from behind the new `AuthGuard`, so that signed-in users can verify their email address and phone number.

### DIFF
--- a/src/app/auth/auth.routes.ts
+++ b/src/app/auth/auth.routes.ts
@@ -15,12 +15,10 @@ import { ShareInviteResolveService } from './resolves/share-invite-resolve.servi
 
 import { AuthGuard } from './guards/auth.guard';
 
-export const childRoutes: Routes = [
+const unauthenticatedRoutes: Routes = [
   { path: 'login', component: LoginComponent, data: { title: 'Log In' } },
   { path: 'signup', component: SignupComponent, data: { title: 'Sign Up' }, resolve: { shareInviteData: ShareInviteResolveService }},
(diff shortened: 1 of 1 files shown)


slifty (approved):
I did *not* test this in local, but the code looks good and I appreciate the use of well named route buckets.

Nice job figuring out the problem and patching it!

jasonaowen:
Kaitlyn found a bug: clicking the log out button during the new onboarding flow left the user in a page `/app/auth`, which showed only the auth banner. I believe this is because the new enclosing route I set up matched: the empty route rendered the `AuthComponent`, and no child routes matched, and there was no trailing redirect.

I moved the `verify` part of the component out of the child routes into the parent route.

Side note: I'm pretty sure the `/app/auth/verify/:email/:code` route no longer works, and hasn't in at least a year, but I decided not to get too deep into that during an emergency bugfix. I look forward to removing this component altogether once we're able to lean on FusionAuth for this functionality.

jasonaowen:
Thank you for the review, @slifty, and for the testing, Kaitlyn! I'm going to merge and deploy this now, but @meisekimiu I would love a post-hoc review in case I missed something, or if there's a better way, or other clean-up for us to do post-RootsTech.

PR #104, opened 2022-03-11
PER-8455/8456: Default share link settings to "on".

meisekimiu:
This PR sets Share Preview and Auto Approve to "on" by default on Share Link settings. Ideally, the default values would also be set on the back end, as that's *actually* what was causing them to be set to off by default!

Resolves both PER-8455 and PER-8456.


meisekimiu:
I did not find where it was, I just noticed that the default "on" behavior was a result of the back end defaulting that.

mithuna:
> Thanks, @meisekimiu ! The "one on, one off" behavior was just because of previous data I had in my local - this works perfectly for newly created share links.
> 
> You mentioned that this is set in the backend. Did you find where or should I go looking?

We set preview and autoapprove [here ](https://github.com/PermanentOrg/back-end/blob/ce8e79d7da1d4e5608f32317f9c150c556f762c8/api/core/shareby_url/bo/shareby_url.bo.php#L163) with the values passed in by UI.

PR #105, opened 2022-03-11
PER-8511: Show "Accept Share" when auto approve is on

meisekimiu:
This PR just changes the text of the "Request Access" button on shares if auto approve is enabled.

Resolves PER-8511.

## Steps to test
- Generate a share link with auto approve enabled
- Access it via another logged in account
- The button should say "Accept Share" instead of "Request Access"
- Test that the button still says "Request Access" if auto approve is disabled.

### DIFF
--- a/src/app/share-preview/components/share-preview/share-preview.component.html
+++ b/src/app/share-preview/components/share-preview/share-preview.component.html
@@ -89,7 +89,13 @@
       </div>
     </div>
     <div class="share-preview-cover-buttons" (click)="stopPropagation($event)" *ngIf="isLoggedIn" >
-      <button class="btn btn-wordpress" (click)="onRequestAccessClick()" [disabled]="waiting || hasRequested" *ngIf="isLinkShare">{{ hasRequested ? 'Access Requested' : requestAccessText }}</button>
+      <button class="btn btn-wordpress" (click)="onRequestAccessClick()" [disabled]="waiting || hasRequested" *ngIf="isLinkShare">
+        <ng-container *ngIf="hasRequested ; else accessText">Access Requested</ng-container>
+        <ng-template #accessText>
+          <ng-container *ngIf="isAutoApprove ; else requestAccess">Accept Share</ng-container>
+          <ng-template #requestAccess>Request Access</ng-template>
+        </ng-template>
+      </button>
       <button class="btn btn-wordpress" (click)="onArchiveThumbClick()" *ngIf="isRelationshipShare">Switch Archive</button>
     </div>
   </div>
@@ -110,7 +116,7 @@
         <button class="btn btn-alternate" *ngIf="isRelationshipShare">Log in</button>
       </strong>
       <strong *ngIf="isLinkShare && isLoggedIn && !hasRequested">
-        <ng-container *ngIf="sharePreviewVO?.autoApproveToggle === 1 ; else requestFullAccessText">
+        <ng-container *ngIf="isAutoApprove ; else requestFullAccessText">
           Accept and view the share in your archive.
         </ng-container>
         <ng-template #requestFullAccessText>
@@ -120,7 +126,13 @@
       <strong *ngIf="isLinkShare && isLoggedIn && hasRequested">Your request is awaiting approval from {{shareAccount.fullName}}.</strong>
       <strong *ngIf="isRelationshipShare && isLoggedIn">Your current archive does not have access to this content. Switch to view or edit.</strong>
       <div *ngIf="isLinkShare && isLoggedIn">
-        <button class="btn btn-alternate" (click)="onRequestAccessClick()" [disabled]="waiting || hasRequested" *ngIf="isLinkShare">{{ hasRequested ? 'Access requested' : requestAccessText }}</button>
+        <button class="btn btn-alternate" (click)="onRequestAccessClick()" [disabled]="waiting || hasRequested" *ngIf="isLinkShare">
+          <ng-container *ngIf="hasRequested ; else accessText">Access Requested</ng-container>
+          <ng-template #accessText>
+            <ng-container *ngIf="isAutoApprove ; else requestAccess">Accept Share</ng-container>
+            <ng-template #requestAccess>Request Access</ng-template>
+          </ng-template>
+        </button>
       </div>
     </span>
   </div>
--- a/src/app/share-preview/components/share-preview/share-preview.component.html
+++ b/src/app/share-preview/components/share-preview/share-preview.component.html
@@ -89,7 +89,7 @@
       </div>
     </div>
     <div class="share-preview-cover-buttons" (click)="stopPropagation($event)" *ngIf="isLoggedIn" >
-      <button class="btn btn-wordpress" (click)="onRequestAccessClick()" [disabled]="waiting || hasRequested" *ngIf="isLinkShare">{{ hasRequested ? 'Access Requested' : 'Request Access' }}</button>
+      <button class="btn btn-wordpress" (click)="onRequestAccessClick()" [disabled]="waiting || hasRequested" *ngIf="isLinkShare">{{ hasRequested ? 'Access Requested' : requestAccessText }}</button>
       <button class="btn btn-wordpress" (click)="onArchiveThumbClick()" *ngIf="isRelationshipShare">Switch Archive</button>
     </div>
   </div>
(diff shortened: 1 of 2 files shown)


cecilia-donnelly (commented):
This works!  Thanks as ever @meisekimiu.  I think it would be neater if the help text also changed:

Screenshot 2022-03-11 at 14-12-32 Sharing Permanent org

That is, if the explainer above the button said "Click here to see the share" or "Click here to view your new file" or was even blank.  What do you think, @k8lyn6 ? 

(Waiting on that question to officially approve.)

k8lyn6:
@meisekimiu @cecilia-donnelly so sorry for the delay! If I'm understanding the workflow to get here (and what the button does) correctly, I think we should write "Accept and view the share in your archive." at the top.

cecilia-donnelly (changes requested):
Officially requesting changes to match @k8lyn6 's note on what the explanatory text should read when auto approve is on.

meisekimiu:
Updated the text when auto approve is on!

src/app/share-preview/components/share-preview/share-preview.component.html
```diff
@@ -109,11 +109,18 @@
...
+        </ng-container>
+        <ng-template #requestFullAccessText>
+          Request full access from {{shareAccount.fullName}}.
+        </ng-template>
+      </strong>
       <strong *ngIf="isLinkShare && isLoggedIn && hasRequested">Your request is awaiting approval from {{shareAccount.fullName}}.</strong>
       <strong *ngIf="isRelationshipShare && isLoggedIn">Your current archive does not have access to this content. Switch to view or edit.</strong>
       <div *ngIf="isLinkShare && isLoggedIn">
-        <button class="btn btn-alternate" (click)="onRequestAccessClick()" [disabled]="waiting || hasRequested" *ngIf="isLinkShare">{{ hasRequested ? 'Access requested' : 'Request access' }}</button>
+        <button class="btn btn-alternate" (click)="onRequestAccessClick()" [disabled]="waiting || hasRequested" *ngIf="isLinkShare">{{ hasRequested ? 'Access requested' : requestAccessText }}</button>
```
jasonaowen, line 123: I see at least four different approaches to varying text in this file:

- HTML elements with `*ngIf` expressions to show/hide
- HTML elements with `*ngIf="(expr) ; else templateName` and a partner `<ng-template>`
- strings selected via the conditional operator, eg `{{ (expr) ? 'string1' : 'string2' }}`
- strings injected via code, eg `chooseArchiveText` and `requestAccessText`

That seems like... kind of a lot. Do we need all of those different approaches to live side-by-side?

Personally, I quite like the first: `*ngIf` attributes with simple boolean expressions to either show or hide the element: it's simple, straightforward, and easy to reason about. In general, my default stance is that text belongs in the component HTML, rather than in code, and the component class should have the appropriately-named booleans to select the desired behavior.

In this specific case, of wanting a single button with various text content, can we use an [`<ng-container>`](https://angular.io/api/core/ng-container) to hang the `*ngIf` conditions on?

Or maybe I'm missing something, in which case I'd love to learn! Is there some drawback to the approach I am suggesting? Is the complexity here merited because the logic of which string to show is equally complex?

jasonaowen (approved):
I don't know how urgent this is, given the turnaround time on this PR so far, but I wanted to get all my reviews done so that this can have the option of going out. It looks like the desired text is being shown, and if not we can revisit. Thanks for the screenshot, @meisekimiu!

That said, I have a suggestion / question below!

meisekimiu:
I updated this PR to take the text out of the component class and put it in the HTML template instead. I'm using the `*ngIf ; else` pattern in the template to choose between the three variations of text for the button.

src/app/share-preview/components/share-preview/share-preview.component.ts
```diff
@@ -194,7 +194,7 @@ export class SharePreviewComponent implements OnInit, OnDestroy {
       this.hasRequested = !!this.sharePreviewVO.ShareVO;
       this.hasAccess = this.hasRequested && this.sharePreviewVO.ShareVO.status.includes('ok');
       if (this.sharePreviewVO?.autoApproveToggle === 1) {
-        this.requestAccessText = 'Accept Share';
+        this.isAutoApprove = true;
```
jasonaowen, line 197: It looks like `this.sharePreviewVO` is initialized upon construction; can we also initialize `this.isAutoApprove` at the same time? I confess I do not fully know how object initialization works, but it would be great if we could just write

```ts
  public isAutoApprove: boolean = (this.sharePreviewVO?.autoApproveToggle === 1);
```

src/app/share-preview/components/share-preview/share-preview.component.html
```diff
@@ -89,7 +89,13 @@
       </div>
     </div>
     <div class="share-preview-cover-buttons" (click)="stopPropagation($event)" *ngIf="isLoggedIn" >
-      <button class="btn btn-wordpress" (click)="onRequestAccessClick()" [disabled]="waiting || hasRequested" *ngIf="isLinkShare">{{ hasRequested ? 'Access Requested' : requestAccessText }}</button>
+      <button class="btn btn-wordpress" (click)="onRequestAccessClick()" [disabled]="waiting || hasRequested" *ngIf="isLinkShare">
+        <ng-container *ngIf="hasRequested ; else accessText">Access Requested</ng-container>
+        <ng-template #accessText>
+          <ng-container *ngIf="isAutoApprove ; else requestAccess">Accept Share</ng-container>
+          <ng-template #requestAccess>Request Access</ng-template>
+        </ng-template>
```
jasonaowen, line 97: This looks better, thank you!

If you're willing, I'd love to hear what you like about `ng-template`s! I think maybe I've over-internalized how Handlebars does things, because I would have defaulted to something more like this:

```
        <ng-container *ngIf="hasRequested">Access Requested</ng-container>
        <ng-container *ngIf="!hasRequested && isAutoApprove">Accept Share</ng-container>
        <ng-container *ngIf="!hasRequested && !isAutoApprove">Request Access</ng-container>>
```

I suppose maybe this construction leans on the developer to make sure they're correctly handling the if-else sequence, whereas the template approach ensures that only one of these will render?

src/app/share-preview/components/share-preview/share-preview.component.html
```diff
@@ -120,7 +126,13 @@
       <strong *ngIf="isLinkShare && isLoggedIn && hasRequested">Your request is awaiting approval from {{shareAccount.fullName}}.</strong>
       <strong *ngIf="isRelationshipShare && isLoggedIn">Your current archive does not have access to this content. Switch to view or edit.</strong>
       <div *ngIf="isLinkShare && isLoggedIn">
-        <button class="btn btn-alternate" (click)="onRequestAccessClick()" [disabled]="waiting || hasRequested" *ngIf="isLinkShare">{{ hasRequested ? 'Access requested' : requestAccessText }}</button>
+        <button class="btn btn-alternate" (click)="onRequestAccessClick()" [disabled]="waiting || hasRequested" *ngIf="isLinkShare">
+          <ng-container *ngIf="hasRequested ; else accessText">Access Requested</ng-container>
+          <ng-template #accessText>
+            <ng-container *ngIf="isAutoApprove ; else requestAccess">Accept Share</ng-container>
+            <ng-template #requestAccess>Request Access</ng-template>
```
jasonaowen, line 133: Oh, and I actually hadn't noticed before that we have this same code twice! Two times is right on the border, so I'll ask and defer to your judgement: would it make sense to extract a helper component or something to avoid this duplication, or would that be much more complexity than is saved?

jasonaowen (approved):
This looks much cleaner to me, thank you, @meisekimiu!

src/app/share-preview/components/share-preview/share-preview.component.html
meisekimiu: Yes, I'd prefer in this structure to have a specific `else` clause instead of having to specify with mutliple `if`s.

(This would be cleaner in something like Vue which has appropriate `v-if`, `v-else`, and `v-else-if` directives.)

src/app/share-preview/components/share-preview/share-preview.component.html
meisekimiu: If this were React I'd definitely extract it out into a helper component because that would take like 5 seconds.

A complaint I have about Angular, or at least the way our Angular codebase is structured, is that creating a new component feels like an incredibly heavy operation (which is why they have the Angular CLI to help out with the boilerplate). So I personally didn't want to do run `ng generate` just for this code here.

src/app/share-preview/components/share-preview/share-preview.component.html
jasonaowen: Thank you! That makes sense to me, and I appreciate the explanation.

src/app/share-preview/components/share-preview/share-preview.component.html
jasonaowen: Oh yep, that sounds like much more work than is merited here. Unfortunate, but thanks for the explanation!

PR #106, opened 2022-03-14
PER-8457: Change "Manage Link" to "Link Settings"

meisekimiu:
This PR changes button text for managing share links on mobile. It is just a text change to make it match the mobile app.

Resolves PER-8457.


cecilia-donnelly (approved):
I didn't actually test this but as you say, it's just a text change.

PR #107, opened 2022-03-17
Remove record attachments

jasonaowen:
This web-app has never supported record attachments, and in fact might not have ever been live at the same time as the partially-implemented back-end feature. Remove the type definitions around record attachments, error messages related to record attachments, and update the test responses to remove the deleted record attachment keys.

[PER-8412 Remove record attachments](https://permanent.atlassian.net/browse/PER-8412)


jasonaowen:
cc @slifty: note that removing keys from API responses - which I believe you are considering - should be matched with updating the front-end type definitions and test data. Probably you already know that, but just in case. :)

meisekimiu (approved):
I'm always happy deleting unused code. Thanks @jasonaowen !

PR #108, opened 2022-03-17
Remove reset password component

jasonaowen:
The functionality for resetting a forgotten password was moved from being in-app (front- and back-end) to being hosted by FusionAuth.

It used to work by sending the user an email with a special link; clicking the link would take the user to the reset password component, which knew how to make the necessary API calls. The API no longer works, and no such emails are being sent any more: instead, FusionAuth sends out the emails, and the link goes to a FusionAuth-hosted password reset page.

Remove the unused password reset component.

[PER-8808 Remove pre-FusionAuth login code](https://permanent.atlassian.net/browse/PER-8808)


meisekimiu (approved):
Looks good! :+1:

PR #110, opened 2022-04-01
Removing unused minor constants

mithuna:
As part of removing the unused minor functionality (backend [PR#60](https://github.com/PermanentOrg/back-end/pull/60)),
removing constants from the web-app.

Ref: https://permanent.atlassian.net/browse/PER-8627


jasonaowen (approved):
Thanks, @mithuna!

PR #111, opened 2022-04-01
Set new onboarding chance to 100%

meisekimiu:
This sets the random chance for new onboarding to 100%, meaning the only way the old onboarding flow is activated is through share invites.

💯


cecilia-donnelly (approved):
Thanks @meisekimiu !

PR #113, opened 2022-04-07
Remove `--rewrite` flag when uploading sourcemaps

meisekimiu:
The upload sourcemaps job of build process is currently broken! Looks like `sentry-cli` updated and we don't need to add the `--rewrite` flag. The `--verbose` flag is also no longer a valid parameter, so I've replaced it with `--log-level info`.

### DIFF
--- a/.github/workflows/build-dev.yml
+++ b/.github/workflows/build-dev.yml
@@ -42,7 +42,7 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: Setup Sentry CLI
-        uses: mathrix-education/setup-sentry-cli@1.0.0
+        uses: mathieu-bour/setup-sentry-cli@1.2.0
         with:
           token: ${{ secrets.SENTRY_TOKEN }}
           organization: permanentorg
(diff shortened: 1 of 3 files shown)


meisekimiu:
I ran a dev build on this branch and the [changes fixed the upload sourcemaps job!](https://github.com/PermanentOrg/web-app/runs/5871694614?check_suite_focus=true)

jasonaowen (approved):
Thanks for digging into this, @meisekimiu! These changes look good.

I think I figured out what happened, specifically:

1. We are using the setup-sentry-cli action to install the Sentry CLI, but are not specifying a version [as they recommend](https://github.com/mathieu-bour/setup-sentry-cli#version), to avoid specifically this kind of problem
    - also the action has been renamed and there's a new version out, which we should upgrade to: https://github.com/mathieu-bour/setup-sentry-cli/blob/main/CHANGELOG.md
2. Three days ago, Sentry published a beta version of the CLI: [2.0.0-beta.0](https://github.com/getsentry/sentry-cli/releases/tag/2.0.0-beta.0). That release has a bunch of breaking changes - as I would expect from the major version number change - and, sure enough, our build broke.

The alternative to these changes is to pin the version number to a stable release. That'll protect us from any future such breaking changes, but also requires periodic proactive maintenance to stay up-to-date.

@meisekimiu since you're the one who is most likely to notice and have to deal with any additional breaking changes in future releases, it's your call on which approach to take!

meisekimiu:
For now I'm fine with using the `latest` version of the Sentry CLI. Though I might suggest in the future we should pin the CLI to 2.x to avoid later breaking changes again. I will upgrade the action we're using though!

PR #115, opened 2022-04-08
PER-8781: Show permissions error for adding members

meisekimiu:
This PR shows a permission error when you try to add members to an archive and don't have the permission to do so (instead of showing the form and only showing the error after you submit it). You should still see the normal form if you are a manager or owner.

Resolves PER-8781.

### DIFF
--- a/src/app/core/components/members-dialog/members-dialog.component.ts
+++ b/src/app/core/components/members-dialog/members-dialog.component.ts
@@ -138,17 +138,18 @@ export class MembersDialogComponent implements OnInit, IsTabbedDialog {
     if (
       archive.accessRole !== 'access.role.manager' && archive.accessRole !== 'access.role.owner'
     ) {
-      this.promptService.confirm(
-        'Learn More',
-        'Add Member',
-        null,
-        null,
-        `You do not have permission to add members to this archive.`
-      ).then(() => {
+      try {
+        await this.promptService.confirm(
+          'Learn More',
+          'Add Member',
+          null,
+          null,
+          `You do not have permission to add members to this archive.`
+        );
         window.open('https://desk.zoho.com/portal/permanent/en/kb/articles/roles-for-collaboration-and-sharing');
-      }).catch(() => {
-        // Catch Promise rejection, but do nothing on "Cancel" button press
-      });
+      } catch {
+        // Catch PromptService rejection, but do nothing on "Cancel" button press
+      }
       return;
     }
     const deferred = new Deferred();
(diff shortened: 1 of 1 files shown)


src/app/core/components/members-dialog/members-dialog.component.ts
```diff
@@ -134,6 +134,23 @@ export class MembersDialogComponent implements OnInit, IsTabbedDialog {
...
+    if (
+      archive.accessRole !== 'access.role.manager' && archive.accessRole !== 'access.role.owner'
+    ) {
+      this.promptService.confirm(
+        'Learn More',
+        'Add Member',
+        null,
+        null,
+        `You do not have permission to add members to this archive.`
+      ).then(() => {
```
jasonaowen, line 147: This method is already marked `async`, and we have other `await`ed code. Can we structure this as a `try`-`catch` block instead of as promises?

src/app/core/components/members-dialog/members-dialog.component.ts
```diff
@@ -134,6 +134,23 @@ export class MembersDialogComponent implements OnInit, IsTabbedDialog {
   }
 
   async onAddMemberClick() {
+    const archive = this.accountService.getArchive();
+    if (
+      archive.accessRole !== 'access.role.manager' && archive.accessRole !== 'access.role.owner'
```
jasonaowen, line 139: Since this is also coming up on the back-end (see https://github.com/PermanentOrg/back-end/pull/80#discussion_r846524903), this is top of mind for me: does the web-app have a mapping of roles to permissions? Ideally this check would not be directly examining the role, but instead be looking up the list of permission the current user's role has and checking to see if the required permission is available.

jasonaowen (commented):
I'll let @cecilia-donnelly test this as we discussed in today's standup, but wanted to ask a question and make a suggestion!

src/app/core/components/members-dialog/members-dialog.component.ts
meisekimiu: It doesn't! That definitely crossed my mind while writing this code, though, and I think it'd be very good to add that, though there are other explicit permissions checks throughout the application that would also need to be changed.

cecilia-donnelly (approved):
I tested as promised and this looks correct to me. Thanks @meisekimiu !

meisekimiu:
I've addressed @jasonaowen's comment about using `async` on the `PromptService` method! This method is completely equivalent to the old code, but might need some checking really quick before it's merged in!

src/app/core/components/members-dialog/members-dialog.component.ts
jasonaowen:
```sh
$ git grep 'access.role' src/app | wc -l
99
```

Yeah, good point, that's out of scope for this PR! Probably worth spending some time thinking about how we'd want this to work, including from the angle of SDK. (cc @slifty)

jasonaowen (approved):
Yep, that looks like a good promise->async transformation to me! Thanks, @meisekimiu!

PR #116, opened 2022-04-11
PER-8497: Fix overlap between widget and upload progress notification

meisekimiu:
Moving it over so it doesn't overlap the Zoho widget. The 92px size was calculated from the Zoho widget's specified CSS width+padding dimensions. This fix is only on desktop view. I'm not even sure how we'd do a fix on mobile view?

Resolves PER-8497.

### DIFF
--- a/src/app/core/components/upload-progress/upload-progress.component.scss
+++ b/src/app/core/components/upload-progress/upload-progress.component.scss
@@ -71,7 +71,7 @@ $easing: $tweaked-ease;
   visibility: hidden;
   background: white;
   position: relative;
-  
+
   &.visible {
     transform: translateY(-100%);
     transition: transform $transition-length $easing, opacity $transition-length $easing, visibility 0s linear  0s;
@@ -106,6 +106,7 @@ $easing: $tweaked-ease;
     box-shadow: $box-shadow;
     user-select: none;
     opacity: 0;
+    margin-right: 92px;
     transform: translateY(#{$screen-padding});
 
     &.visible {


jasonaowen:
> I'm not even sure how we'd do a fix on mobile view?

We discussed this in today's standup, and @meisekimiu is planning on uploading a screenshot to this PR to start a discussion with @k8lyn6 and whoever else is interested about what our options are.

cecilia-donnelly (approved):
This certainly moves the upload progress bar over on desktop.  I looked at it in a couple different views using the responsive design tool in Firefox and it doesn't look bad there, either, for the record.

@meisekimiu I do think it would be good to attach some screenshots here for @k8lyn6 .

meisekimiu:
Here's a screenshot of the adjusted upload progress bar, @k8lyn6 !



Here's a fullscreen screenshot. Forgive the garbled text of the upload modal here, I'm using some extra (non-commited) code here to force the toolbar to pop up when it shouldn't be just so I could work on this ticket and take screenshots. So... just pretend the progress bar and file count are showing correctly here!

jasonaowen:
Thanks for uploading the desktop screenshot, @meisekimiu! I think it would also be helpful to have a screenshot of the mobile view while uploading, so:

Screen Shot 2022-04-14 at 17 25 21

k8lyn6:
@meisekimiu I think the desktop looks perfect! @jasonaowen for the mobile view, what does it look like when the text says "X of X files uploading" in the right corner? I worry that the widget still somewhat blocks that information there.

cecilia-donnelly:
Let's merge this since it's a good fix for desktop and isn't making anything worse on mobile.

PR #117, opened 2022-04-13
Remove batch constants and batchNbr

jasonaowen:
Batching is part of the old upload workflow, where multiple files were collected together for uploading. The new upload workflow does not use batches.

As a result, the API has been returning null values for `batchNbr` since we switched everyone over to the new upload flow. Soon, the API will no longer be sending `batchNbr`.

Delete the unused constants relating to batches, remove `batchNbr` from the type definition for records, and update our test responses.

See also companion PR https://github.com/PermanentOrg/back-end/pull/84, removing batches from the back-end.


meisekimiu (approved):
Thanks for clearing out this code! :+1:

PR #118, opened 2022-04-26
Upgrade to Angular 11

meisekimiu:
This PR is a smaller upgrade to Angular 11 instead of jumping all the way to Angular 13. Currently, the web-app is running a hybrid of Angular 10 and an Angular 11 prerelease. This upgrades the web-app to the latest stable (not-prerelease!) version of Angular 11, including peer dependency upgrades, automatic code migrations, and post-upgrade adjustments to those code migrations and our Angular configuration.

@jasonaowen and I have tested all of these already, but here's a quick testing checklist to verify that the things we've upgraded aren't broken:
- The countup library we're using shows a quick animation of a user's profile progress ticking up when they first navigate to the profile page. Verify that it still works since we've upgraded the library.
- Verify Google Maps still works after two major version upgrades.
- Verify that assets aren't broken on the local proxy server (the one that is run by `npm run dev`/`ng serve`, not the built version served by the Virtualbox environment)
  - A good test is to see if the Permanent logo renders properly on the login/signup/etc. pages.
- Links between pages all work as expected.
  - Some link resolution behavior has changed in the Angular upgrade and we're *pretty* sure nothing is broken here, but double checking is always a good idea.
  - A good test link is the "Already have an account?" link on the signup page.

Part of PER-8819: Upgrade Angular.

Also will fix the following dependabot PRs:
- Resolves #101 (`url-parse` has been bumped to 1.5.10)
- Resolves #102 (`karma` has been bumped to latest version)


jasonaowen (approved):
There were a few more dependencies that need to be upgraded to 11:
- `@angular-eslint/builder`
- `@fortawesome/angular-fontawesome`
- `shallow-render`

After checking in with @meisekimiu, I updated this PR to include those in the Angular upgrade commit.

In addition to pairing, I was able to test this locally with `npm run dev` and running through the list of functionality in the description. I was also able to run the linter with `npm run lint`, and wow is it mad, but it does run! This also fixes the local CSS issue: `npm run build:local` and going through the new onboarding shows the correct styling.

jasonaowen:
Thanks for letting me pair with you on this, @meisekimiu! I feel like I understand these changes much better, and learned some things about Angular.

jasonaowen:
I see a bunch of errors in the [test log](https://github.com/PermanentOrg/web-app/runs/6185554326?check_suite_focus=true#step:6:143), but the tests continue to pass:

> ERROR: 'NG0303: Can't bind to 'ngbTooltip' since it isn't a known property of 'div'.'

I think those are related to [ng-bootstrap tooltips](https://ng-bootstrap.github.io/#/components/tooltip), but tooltips continue to work, so I'm not entirely sure what's up with that. Any ideas, @meisekimiu?

meisekimiu:
I've fixed these `ngbTooltip` errors! I notice there are also a lot of `fa-icon` errors as well, which are caused by the same problem. Unlike these tooltip errors though, I believe these ~~warnings~~ errors have been around for a while. I now know how to actually fix them, but I'm not sure we want to do that in this PR since it isn't caused by the upgrade to Angular 11.

jasonaowen (approved):
> The cause of these errors is that the proper ng-bootstrap module isn't being imported properly in the testing environment.

Great find! That totally makes sense; thanks for looking into that and fixing it.

> I notice there are also a lot of fa-icon errors as well, which are caused by the same problem. Unlike these tooltip errors though, I believe these warnings errors have been around for a while. I now know how to actually fix them, but I'm not sure we want to do that in this PR since it isn't caused by the upgrade to Angular 11.

Yeah, that makes sense to me - +1 to addressing that in a separate PR!

I also see several "has no expectations" warnings, and confirmed that those existed before, so fixing them is out of scope of this PR, too.

PR #120, opened 2022-04-29
Remove unused search

jasonaowen:
Remove some dead code: a method that is never called, an unused component that was reachable but not linked to, and the helper method that backed both.


src/app/public/public.routes.ts
```diff
@@ -36,10 +35,6 @@ export const routes: RoutesWithData = [
         path: 'error',
         component: ItemNotFoundComponent
       },
-      {
-        path: 'search',
```
jasonaowen, line 40: Note that I'm just letting the default error handler take control when someone visits `/p/search`. I'm making the assumption that if we were directly linking to this component we would have, y'know, finished it, but let me know if you think we should redirect to `/p` instead of showing an error!

meisekimiu (approved):
Thanks for doing this! I think showing the default error page is fine, since this was never supposed to be accessed in the first place and we shouldn't really have to acknowledge that it ever existed in the code.

jasonaowen:
Awesome - thanks @meisekimiu!

PR #121, opened 2022-05-02
Remove code for unused `/search/record` endpoint

meisekimiu:
The `recordByNameObservable` function which calls the /api/search/record endpoint is not used anywhere in the app currently. Let's go ahead ahead and remove it!


jasonaowen (approved):
Thanks, @meisekimiu!

Looks like this function was added in commit b801ae9c02fceb48456daf76e6110d1b6c47ad7c, and then the last call to it was removed in 2da2d4540d6b74894339e77a5ef9338f4f91715f - less than 2 hours later.

PR #124, opened 2022-05-25
PER-8843: Remove donor level

cecilia-donnelly:
In a sibling change, I removed the concept of account or donor level
from the backend and database.  Remove it here too for completeness.

PER-8843: Add support for Apple and Google Pay on donations-firebase

See https://github.com/PermanentOrg/back-end/pull/178 for the back-end side of this.

@meisekimiu , this is quite minimal but I didn't see any other places where we use this "account level" concept on the front end.  Apologies if I'm missing something!

### DIFF
--- a/constants/master_en.json
+++ b/constants/master_en.json
@@ -10,18 +10,6 @@
       "editor": "Editor"
     }
   },
-  "account": {
-    "level": {
-      "benefactor": "Benefactor",
-      "founding_patron": "Founding Patron",
-      "innovators_circle": "Innovators Circle",
-      "none": "None",
-      "partner": "Partner",
-      "presidents_circle": "Presidents Circle",
-      "principals_circle": "Principals Circle",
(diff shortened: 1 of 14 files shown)


jasonaowen:
Please also remove the `level` attribute from the test responses! For example, https://github.com/PermanentOrg/web-app/blob/caf43d51632b6a8d49d22a0e67319e8446292e49/src/test/responses/auth.login.success.json#L18

jasonaowen:
Also the string constants, please! https://github.com/PermanentOrg/web-app/blob/caf43d51632b6a8d49d22a0e67319e8446292e49/constants/master_en.json#L13-L24

jasonaowen (approved):
This looks good to me - thanks, @cecilia-donnelly!

meisekimiu (approved):
Just going to add my review here too: looks good!

Issue #91, opened 2022-02-09
Need Linting / Formatting Tool


meisekimiu:
While we're eventually going to work on adding `eslint` to the web-app, I want to put a ticket here in the repo because certain spacing styles are inconsistent with the app and my IDE is freaking out over how exactly to indent things (Specifically within the `AuthModule` for whatever reason). I would like to run some kind of formatter to end these troubles and so I don't have to worry about how many spaces I need to indent everything by, knowing that the formatter can fix everything later.

dyerjos:
I recommend using eslint with [prettier](https://prettier.io/). Something like [this](https://gist.github.com/Imshyamal/95833b32124c68480737f2a28f0a37d4)

PR #125, opened 2022-05-25
Improvements to Left Side Menu

meisekimiu:
This PR updates the left side menu to be more consistent with the mobile app's design. One thing that wasn't really specified in the zeplin designs was what this should look like in mobile view. The layout of the mobile view is kept intact, just including the new icon and color changes.

Resolves PER-8667.


meisekimiu:
Unit tests are randomly breaking? It works on my machine:

<span title="have to re-roll this PR's stats">Going to try rerunning until they work I guess.</span>

src/app/core/components/left-menu/left-menu.component.html
```diff
@@ -44,53 +51,61 @@
...
+      <div class="menu-item apps-menu hide-mobile" [class.active]="checkMenuItemActive('/apps')">
+        <a class="apps-link-main" [routerLink]="['/apps']">
+          <img class="svg-icon" src="assets/svg/left-menu/app-files-icon.svg" />
+          <span>App Files</span>
+        </a>
+        <div class="apps-toggle" [class.reverse]="showAppsSubfolders" (click)="showAppsSubfolders = !showAppsSubfolders">
+          <i class="material-icons">expand_more</i>
+        </div>
+      </div>
+      <a class="menu-item hide-desktop" [routerLink]="['/apps']" [class.active]="checkMenuItemActive('/apps')">
```
jasonaowen, line 78: There are several long lines introduced in this file; would you mind spilling them to multiple lines? (Probably we should look into using Prettier in this repo!)

src/app/core/components/left-menu/left-menu.component.scss
```diff
@@ -33,6 +33,60 @@ $archive-options-shadow-offset: 0.5rem;
...
+    .apps-toggle {
+      padding-right: 12px;
+      margin-right: 20px;
+      height: 3 * $grid-unit;
+      display: flex;
+      align-items: center;
+
+      i {
+        font-size: $mat-icon-lg;
+        transition: transform 250ms $tweaked-ease;
```
jasonaowen, line 79: This is very fun! Amazing how rewarding a tiny bit of animation can feel. :slightly_smiling_face:

src/app/core/components/left-menu/left-menu.component.html
```diff
@@ -44,53 +51,61 @@
...
-      <a class="menu-item" [routerLink]="['/apps']" [class.active]="checkMenuItemActive('/apps')">
-        <i class="ion-md-swap"></i>
-        Apps
+      <hr class="hide-mobile" />
+      <div class="menu-item apps-menu hide-mobile" [class.active]="checkMenuItemActive('/apps')">
+        <a class="apps-link-main" [routerLink]="['/apps']">
+          <img class="svg-icon" src="assets/svg/left-menu/app-files-icon.svg" />
+          <span>App Files</span>
+        </a>
+        <div class="apps-toggle" [class.reverse]="showAppsSubfolders" (click)="showAppsSubfolders = !showAppsSubfolders">
```
jasonaowen, line 74: Would the [`<details>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) be a simpler alternative to using JavaScript to expand/collapse functionality?

src/app/core/components/left-menu/left-menu.component.html
```diff
@@ -44,53 +51,61 @@
...
+          <i class="material-icons">expand_more</i>
```
jasonaowen, line 75: Should we show the expand icon if there are no app folders? Feels a little weird to click and just have the icon spin around.

src/app/core/components/left-menu/left-menu.component.html
```diff
@@ -12,25 +12,32 @@
       <div class="menu-header-desktop" [class.show-shadow]="showArchiveOptions">
         <div class="archive">
           <div class="archive-thumb" prBgImage [bgSrc]="archive.thumbURL500" [cover]="true"></div>
-          <div class="archive-name">The {{archive.fullName}} Archive</div>
-          <div class="archive-access">Access role: {{archive.accessRole | prConstants}}</div>
-          <div class="archive-options-toggle" (click)="showArchiveOptions = !showArchiveOptions"
+          <div class="archive-name-access">
+            <div class="archive-options-toggle" (click)="showArchiveOptions = !showArchiveOptions"
```
jasonaowen, line 16: This isn't new behavior, so feel free to tell me this is out of scope, but I find it frustrating for the archives options to shut after clicking on one - particularly when the menu collapse animation completes before the dialog box (or whatever) begins animating in. Can we make it stay open? Should that be part of this?

jasonaowen (commented):
This is looking good! Quick screenshot for posterity:

Screen Shot 2022-05-26 at 14 21 40

jasonaowen:
While reviewing this PR, I learned a bit more about how the old icons work, and opened #127 and #128!

src/app/core/components/left-menu/left-menu.component.html
meisekimiu: Regarding Prettier, see #91!

src/app/core/components/left-menu/left-menu.component.html
meisekimiu: I think at least based on the fact that we're using very specifically placed expander buttons, using JavaScript is better here. Especially on the Apps link.

src/app/core/components/left-menu/left-menu.component.html
meisekimiu: This is a really good suggestion and I think that should be part of the ticket because it should be relatively simple. The ticket *is* called "Improvements to Left Side Menu", and this is an improvement!

src/app/core/components/left-menu/left-menu.component.html
jasonaowen: Ah, that makes sense; I wasn't sure how much the details disclosure widget could be styled / positioned.

src/app/core/components/left-menu/left-menu.component.html
jasonaowen: Nice, thank you!

src/app/core/components/left-menu/left-menu.component.html
jasonaowen: Awesome, this feels great!

jasonaowen (approved):
This is fantastic, thanks @meisekimiu! The formatting looks good, although yeah not sure if Prettier will disagree - +1 to setting all that up ASAP.

I was able to run this locally in both desktop and mobile views, and it looks and feels great!

PR #129, opened 2022-05-27
Install Prettier

### LINKED ISSUE #91: Need Linting / Formatting Tool

meisekimiu:
I'm tired of having to manually format code. As discussed in #125, we should not even really have to think of this. This PR installs prettier to do formatting for us. This PR does not actually add it to CI, install pre-commit hooks, or run it against any files (though I did have to test it out on some of the files I've recently working on just to make sure it works properly).

My thought is that we probably shouldn't just run it on our whole project all at once. Maybe I'm wrong though? Instead I'm thinking we just run it on any files we have changes on in the future.

Resolves #91

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -26,6 +26,9 @@ jobs:
       - name: Install dependencies
         run: npm install
 
+      - name: Run Prettier
+        run: npx prettier --check --require-pragma .
+
       - name: Run unit tests
         env:
           GOOGLE_API_KEY: ${{ secrets.GOOGLE_API_KEY_DEV }}
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,2 +1,6 @@
-# package.json is formatted by package managers, so we ignore it here
-package.json
+*.json
+*.webmanifest
+*.yml
+node_modules
+test-reports
+dist
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,3 @@
+{
+  "singleQuote": true
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -16842,6 +16842,12 @@
       "integrity": "sha512-vkcDPrRZo1QZLbn5RLGPpg/WmIQ65qoWWhcGKf/b5eplkkarX0m9z8ppCat4mlOqUsWpyNuYgO3VRyrYHSzX5g==",
       "dev": true
     },
+    "prettier": {
+      "version": "2.6.2",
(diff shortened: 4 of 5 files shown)


jasonaowen:
The approach I'm taking with back-end is to use the [pragma functionality](https://prettier.io/docs/en/options.html#require-pragma) to select which files to run on! Basically, Prettier can insert a comment when we tell it to, and then in CI we can tell Prettier to only check files with that comment. I'm not sure if/how that works with HTML files, though, which we definitely want to include in the scope of Prettier; that might require some experimentation.

We'll probably also want a [`.prettierignore` file](https://prettier.io/docs/en/ignore.html), because otherwise it'll likely want to reformat all our sample response JSON files.

meisekimiu:
Fun fact: We already [have a `prettierignore` file!](https://github.com/PermanentOrg/web-app/blob/main/.prettierignore) Even though we only now just added prettier... :thinking: 

Anyway, I've added it to CI with the pragma functionality!

meisekimiu:
I added the `@format` to some files in a now-removed commit to test the CI. You can see the [failing run from formatting errors.](https://github.com/PermanentOrg/web-app/runs/6631693714) Also, you apparently just add the pragma like this in HTML templates:
```html
<!-- @format -->
```

jasonaowen:
> Fun fact: We already [have a `prettierignore` file!](https://github.com/PermanentOrg/web-app/blob/main/.prettierignore) Even though we only now just added prettier... :thinking:

Haha, looks like the previous developer at least tried it out! I'm glad we're adding it in a more sustainable fashion.

jasonaowen:
> Also, you apparently just add the pragma like this in HTML templates: ...

I think you probably know this, but just to be clear, Prettier can add the pragma itself! eg:

```sh
npm run prettier -- --write --insert-pragma \
  src/app/announcement/components/announcement/announcement.component.html
```

Which brings up a question: should we document this in the README? You can read - or take! - what I wrote in the [back-end README](https://github.com/PermanentOrg/back-end/blob/main/README.md#prettier).

jasonaowen (approved):
Yay for prettier code! I look forward to formatting all the things. Thanks, @meisekimiu!

PR #130, opened 2022-05-31
Delete donate component

### LINKED ISSUE #126: Delete unused donation component
The [`DonateComponent`](https://github.com/PermanentOrg/web-app/blob/6457b88a3f759235a042c4420309f273963e12de/src/app/core/components/donate/donate.component.ts) is not used anywhere. There is a route that displays it, but nothing links to that URL.

It's using an API endpoint that may no longer work, and which we intend to delete as unused. Delete the unused component and the repo code that only it calls.

meisekimiu:
This PR deletes the `DonateComponent`, the now-unused `BillingRepo` methods it used to call, as well as some other dependents that are no longer used including: The `BillingCardVO` model and its associated test data and response helper methods and prompt types that are now unused.

(I also ran Prettier on the changed files that still exist and added the pragma)

Resolves #126.


src/app/shared/components/prompt/prompt-fields.ts
```diff
@@ -150,133 +150,6 @@ export function INVITATION_FIELDS(initialEmail?: string): PromptField[] {
   return fields;
 }
 
-export const CREDIT_CARD_FIELDS: PromptField[] = [
```
jasonaowen, line 153: Ooh, fantastic catch, I didn't notice those when I spent a few minutes on this last Friday!

src/app/shared/components/prompt/prompt-fields.ts
```diff
@@ -142,164 +160,41 @@ export function INVITATION_FIELDS(initialEmail?: string): PromptField[] {
...
-    fieldName: 'cardExpMonth',
-    placeholder: 'Expiration month',
-    type: 'select',
-    validators: [Validators.required],
-    config: {
-      autocomplete: 'cc-exp-month',
-      autocorrect: 'off',
-      autocapitalize: 'off'
-    },
-    selectOptions: expMonths.map((month) => {
```
jasonaowen, line 186: Looks like this was the last reference to `expMonths`, so let's delete that array, too!

src/app/shared/components/prompt/prompt-fields.ts
```diff
@@ -142,164 +160,41 @@ export function INVITATION_FIELDS(initialEmail?: string): PromptField[] {
...
-    fieldName: 'cardExpYear',
-    placeholder: 'Expiration year',
-    type: 'select',
-    validators: [Validators.required],
-    config: {
-      autocomplete: 'cc-exp-yeah',
-      autocorrect: 'off',
-      autocapitalize: 'off'
-    },
-    selectOptions: expYears.map((year) => {
```
jasonaowen, line 203: As well

src/test/responses/billing.getBillingCards.multiple.success.json
```diff
@@ -1,101 +0,0 @@
-{
```
jasonaowen, line 1: Putting this here for lack of a better place: can we also delete `src/test/responses/billing.processPayment.storageOnly.success.json`?

src/app/core/components/donate/donate.component.ts
```diff
@@ -1,264 +0,0 @@
...
-export class DonateComponent {
-  public donationStage: DonationStage = DonationStage.Storage;
-  public donationForm: FormGroup;
-  public cards: BillingCardVO[];
-  public currentAddress: any;
-  public waiting: boolean;
-
-  public storageOptions = [1, 3, 5, 10, 25];
-  public storageAmount = DEFAULT_STORAGE_AMOUNT;
-  public pricePerGb: number = APP_CONFIG.pricePerGb;
```
jasonaowen, line 43: https://github.com/PermanentOrg/web-app/blob/1d86f65af0d7e39801324a9c296f5e804899d68e/src/app/app.config.ts#L3

Looks like there's one more reference to `APP_CONFIG.pricePerGb` in `PhaseProgressComponent`: https://github.com/PermanentOrg/web-app/blob/1d86f65af0d7e39801324a9c296f5e804899d68e/src/app/pledge/components/phase-progress/phase-progress.component.ts#L74

Which is maybe a thread to tug a bit harder on? But maybe in a separate PR?

jasonaowen (approved):
This looks great! Everything I had found last week and more, and - most importantly - deletes the API calls I wanted.

A few small additional ideas of additional things to delete below.

What's up with the test failures?

meisekimiu:
> What's up with the test failures?

This is one of those random manage-tags test failures. I actually was looking into fixing that right after I put this PR up since it's rather elusive (read: random) and I think at this point we should try getting rid of this bug while we can identify it... but I wasn't able to figure out how to really fix it and didn't want to delay this too much so I just reran the test job on this PR.

src/test/responses/billing.getBillingCards.multiple.success.json
meisekimiu: Oh right, I was so caught up on deleting the BillingCardVO or whatever it was called that I totally forgot there were tests for other deleted endpoints!

jasonaowen:
That makes sense! Glad to hear it's existing / unrelated, and +1 to fixing separately and soon. Thanks!

jasonaowen (approved):
Awesome, thanks for following up on those! Net negative one thousand(ish) lines, love it!

PR #133, opened 2022-06-09
Remove references to directives

liam-lloyd:
As https://github.com/PermanentOrg/back-end/pull/199 finishes eliminating the old directives code from the backend, this PR follows upon that work and removes directive-related constants and data members from the frontend.

https://permanent.atlassian.net/browse/PER-8909


jasonaowen (approved):
Looks like you got it all! CI runs and the project builds, so +1 from me. Thanks, @liam-lloyd!

PR #135, opened 2022-07-01
Run production build in CI

meisekimiu:
Angular uses different compilers for development and production modes. The development proxy server, as well as the test runner, both use the "development" compiler which allows things that the production compiler does not. There is currently no way to configure tests to run in the production compiler, so let's run a production build as part of CI so we don't merge something that won't build.


meisekimiu:
Turns out that upgrading angular seems to have fixed this issue? Neat!

PR #142, opened 2022-08-17
Fix Public Gallery environment detection issue

meisekimiu:
A tricky thing we didn't catch in the original PR was that the `archives` property on the GalleryComponent class was being intialized before the `environment` property. As a result, when loading in Featured Archives, the environment property was still undefined and could not detect when the application was in production mode, making it unable to load the production Featured Archives data. This fixes this issue by explicitly defining the `environment` property in `getFeaturedArchives` so this initialization issue doesn't happen.

I also fixed some CSS that doesn't cause any issues with archives but does cause issues in the rare case there are no featured archives (when the fixed bug is active or on the staging environment for example).

Picture of both bugs:

Notice:
- the lack of archives even though this is production and it should be loading from our array of static data
- the weird scrollbars on the side of the screen

This fixes both of those.

### DIFF
--- a/src/app/gallery/components/gallery/gallery.component.scss
+++ b/src/app/gallery/components/gallery/gallery.component.scss
@@ -8,7 +8,6 @@
 }
 
 .public-gallery {
-  overflow-x: hidden;
   padding: 0 2em;
   color: #131b4a;
   max-width: 1080px;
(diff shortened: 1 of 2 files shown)


liam-lloyd (approved):
If I'm understanding correctly the bug never appeared locally anyway, so we can't test this fix locally. The code looks good, so approved.

meisekimiu:
Thanks! There's a unit test failure due to that pesky random Manage Tags test failure, but they pass locally without issue:



Going to merge now!

PR #144, opened 2022-08-18
Delete Unused Components

meisekimiu:
This PR deletes several unused components from the codebase. These components were mostly found by:
- Generating a list of all components
- Subtracting out all components that are attached to a route
  - Also verifying that each routed component is currently used! This only resulted in deleting `TermsComponent`
- Searching the rest of the project for usages of the rest of these components

Unfortunately, none of these components were the unique users of any specific endpoints. This means that while this search is good for cleaning up the web-app codebase, this didn't actually help identify any unused backend endpoints in particular.

I've also not committed my "detect unused components" script. Should I?


meisekimiu:
Manage tags tests are failing and they keep failing after I rerun them every time. I think it could be possible that deleting some components (and their tests as well) is contributing to the random/chaotic failure of the manage tags test. Anyway, I've opened up #145 to fix the broken test.

meisekimiu:
(Rebased this on main now that we merged #145, so tests pass now! :tada: )

liam-lloyd (approved):

PR #145, opened 2022-08-18
Fix Manage Tags Tests

meisekimiu:
We were throwing an Error when the delete endpoint was inaccessible, but this was causing issues in our tests. Let's skip throwing the error for now, as the cause of this error is either a loss of internet connection or an issue with the backend endpoint (in which case, we'll have backend logs to see the issue) and therefore don't really need to throw (and therefore report) this error.

The only reason why we were initially throwing the error here was so Sentry can catch it. I think it's easier just to delete this line since I think it will only ever report one of the previously mentioned scenarios (we don't need to be alerted whenever someone has a blip in their internet connection), especially if it fixes our tests.

### DIFF
--- a/src/app/core/components/manage-tags/manage-tags.component.spec.ts
+++ b/src/app/core/components/manage-tags/manage-tags.component.spec.ts
@@ -117,12 +117,15 @@ describe('ManageTagsComponent #manage-tags', () => {
   });
 
   it('should not delete a tag if an error happens', async () => {
-    const { find, fixture, outputs } = await defaultRender();
+    const { element } = await defaultRender();
     throwError = true;
-    find('.delete')[0].nativeElement.click();
-    await fixture.whenStable();
-    await fixture.detectChanges();
-    expect(find('.tag').length).toBe(2);
+    try {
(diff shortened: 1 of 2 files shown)


liam-lloyd:
I think there may be some cases where we would still want to see this error in Sentry. We can't do anything about connection interruptions, but if there's a problem on the backend (or a problem on the frontend causing us to send invalid requests) logging here seems potentially useful for tracing an error through the stack. Also, if I'm reading the code right we're mocking calls to this endpoint in the tests that are failing, and it seems very strange that mocked calls are intermittently failing.

meisekimiu:
Pushed up a new fix. This new version makes sure the thrown error is bubbled up the various promises correctly so we can properly catch the error in our tests. I think this was the issue beforehand; due to the wrapped promises and async nature of the error, sometimes the thrown error would end up leaking out of the test that ran it and wouldn't be able to be handled correctly. With the new way the promises are set up, the test can wait until the error is actually thrown and catch it properly.

liam-lloyd (approved):

PR #147, opened 2022-08-24
Add Apple Smart App Banner Meta Tag

meisekimiu:
Add a meta tag to the base HTML pages to allow iOS to show a smart app banner for the Permanent iOS app.

Resolves PER-8453: Notify members that they can download app (iOS)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -21,6 +21,7 @@
   <meta name="msapplication-TileColor" content="#2b5797">
   <meta name="msapplication-config" content="assets/icon/browserconfig.xml">
   <meta name="theme-color" content="#ffffff">
+  <meta name="apple-itunes-app" content="app-id=1571883070">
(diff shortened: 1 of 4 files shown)


liam-lloyd (approved):
I can't test this due to lack of an iOS device, but from visual inspection it looks good to me.

meisekimiu:
> Yey! My only thought is that it would be nice to store the app id in one place instead of 4. (It probably won't ever change, but just in case.)

I've looked into this and it seems like this is the only way to do it without investing a disproportionate amount of time on it. It seems like the "having multiple different `index.html` files" approach is the standard Angular way of doing things, so embedding variables in the static HTML files is not really a thing. We could dynamically add the `<meta>` tag, though I'd prefer that it's in the static HTML to make sure that browsers and devices can read it on page load, even if there are JS errors.

(That being said, it would be nice to maybe revisit this issue one day, since I'd love to get rid of the giant New Relic JS import in each `index.html`)

cecilia-donnelly:
Interesting, thank you!

PR #151, opened 2022-09-15
Add create account modal to share preview screen

liam-lloyd:
Notes:
- I set the existing banner up to not render in cases where the user is not logged in, because making that banner coexist with the modal was looking difficult and we never want it to do that anyway
- Most of the changes to `share-preview.component.ts` are because I semi-accidentally ran Prettier on it
- I also deleted some commented out CSS that I came across

### DIFF
--- a/src/app/share-preview/components/create-account-dialog/create-account-dialog.component.ts
+++ b/src/app/share-preview/components/create-account-dialog/create-account-dialog.component.ts
@@ -1,0 +1,29 @@
+/* @format */ 
+import { Component, OnInit, Input, Inject } from '@angular/core';
+import { DialogRef, DIALOG_DATA } from '@root/app/dialog/dialog.module';
+import { DeviceService } from '@shared/services/device/device.service';
+
+@Component({
+  selector: 'pr-create-account-dialog',
+  templateUrl: './create-account-dialog.component.html',
+  styleUrls: ['./create-account-dialog.component.scss'],
+})
+export class CreateAccountDialogComponent implements OnInit {
+  sharerName: string = this.data.sharerName;
+
+  constructor(
+    private dialogRef: DialogRef,
+    private device: DeviceService,
+    @Inject(DIALOG_DATA) public data: any
+  ) {}
+
+  ngOnInit(): void {}
+
+  isMobile(): boolean {
+    return this.device.isMobileWidth();
+  }
+
+  close(): void {
+    this.dialogRef.close();
+  }
+}
--- a/src/app/share-preview/components/share-preview/share-preview.component.html
+++ b/src/app/share-preview/components/share-preview/share-preview.component.html
@@ -30,9 +30,9 @@
   <pr-breadcrumbs></pr-breadcrumbs>
 </div>
 <div class="share-preview-content" [ngClass]="{'navigating': isNavigating}">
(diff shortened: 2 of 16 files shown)


src/app/share-preview/components/create-account-dialog/create-account-dialog.component.ts
```diff
@@ -0,0 +1,29 @@
...
+import { DialogRef, DIALOG_DATA } from '@root/app/dialog/dialog.module';
+import { DeviceService } from '@shared/services/device/device.service';
+
+@Component({
+  selector: 'pr-create-account-dialog',
+  templateUrl: './create-account-dialog.component.html',
+  styleUrls: ['./create-account-dialog.component.scss']
+})
+export class CreateAccountDialogComponent implements OnInit {
+
```
meisekimiu, line 11:
```suggestion

```
Minor formatting fix for consistency.

src/app/share-preview/components/share-preview/share-preview.component.html
```diff
@@ -30,9 +30,9 @@
   <pr-breadcrumbs></pr-breadcrumbs>
 </div>
 <div class="share-preview-content" [ngClass]="{'navigating': isNavigating}">
-  <router-outlet></router-outlet>
+  <router-outlet (activate)="subscribeToItemClicks($event)" (deactivate)="unsubscribeFromItemClicks()"></router-outlet>
```
meisekimiu, line 33: TIL about `(activate)` and `(deactivate)` events on `router-outlet`! That's really useful!

src/app/share-preview/components/share-preview/share-preview.component.ts
```diff
@@ -19,22 +28,23 @@ import { EVENTS } from '@shared/services/google-analytics/events';
 import { READ_ONLY_FIELD } from '@shared/components/prompt/prompt-fields';
 import { PromptService } from '@shared/services/prompt/prompt.service';
 import { Deferred } from '@root/vendor/deferred';
+import { Dialog } from '@root/app/dialog/dialog.module';
+import { FileListComponent } from '../../../file-browser/components/file-list/file-list.component';
```
meisekimiu, line 32:
```suggestion
import { FileListComponent } from '@fileBrowser/components/file-list/file-list.component';
```
You can import it like this as well! ...Although my IDE is saying this import isn't even needed, so actually maybe we can just delete this line.

meisekimiu (changes requested):
This looks really good! Thanks for working through the issues with the Dialog and FileListItem components! I've left some suggestions for some small formatting changes.

There is just one slightly bigger issue I've noticed: You can get two copies of the modal to show up at the same time if you click a thumbnail within the 1 second window after the page loads. I think it's completely plausible for a user to just instinctively click the big colorful thumbnail as soon as possible (that's how I found this issue) and trigger this bug.

I've already tested a bit, and this should be an easy fix: you should be able to just store a boolean value on the SharePreviewComponent class that is set when a dialog is opened, and reset when the dialog is closed (`Dialog.open` is a promise so you can already react to when the dialog is closed from within the component!).

meisekimiu (approved):
Looks good! Thanks for the change! :+1:

PR #153, opened 2022-09-21
Rename /myfiles to /private

meisekimiu:
We have been slowly renaming the "My Files" folder to "Private Files" in our application. This PR changes the url for the Private Root from /app/myfiles/** to /app/private/**. The core route configuration was changed so /app/private is now the URL for the private root, and any internal links/redirects within the app were also changed to point to this new location.

To make things a bit nicer for people who may have bookmarked URLs within their archive or may just be hitting a link we missed somewhere else, a new route guard was made that will redirect any subdirectories from /app/myfiles/** to /app/private/**. This is used instead of a more standard redirect to keep the subdirectories in the path intact instead of just redirecting to /app/private.

**Steps 2 Test:**
1. Log into the app and check if the URL says /app/private
2. Open various dialogs and other pages and make sure they redirect to /app/private properly
3. Manually navigate to /app/myfiles and see that it redirects properly to /app/private
4. Move into a folder that's inside of your Private Root, and then change the "private" in the url to "myfiles". Verify that it redirects properly and you're still inside of the folder
   * So for example, I navigated into a subfolder and changed `/app/private/0001-000m/22` to `/app/myfiles/0001-000m/22`. The URL should change back to the private version without kicking you out of the folder.

Resolves PER-8999.

### DIFF
--- a/src/app/core/guards/myfiles.guard.ts
+++ b/src/app/core/guards/myfiles.guard.ts
@@ -1,0 +1,31 @@
+/* @format */
+import { Injectable } from '@angular/core';
+import { Router } from '@angular/router';
+import {
+  ActivatedRouteSnapshot,
+  CanActivate,
+  RouterStateSnapshot,
+  UrlTree,
+} from '@angular/router';
+import { Observable } from 'rxjs';
+
+/**
+ * Redirects any nested routes from /app/myfiles/** to /app/private/**
+ */
+@Injectable({
+  providedIn: 'root',
+})
+export class MyfilesGuard implements CanActivate {
+  constructor(private router: Router) {}
+
+  canActivate(
+    route: ActivatedRouteSnapshot,
+    state: RouterStateSnapshot
+  ):
+    | Observable<boolean | UrlTree>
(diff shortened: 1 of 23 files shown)


src/app/core/guards/myfiles.guard.ts
```diff
@@ -0,0 +1,31 @@
...
+} from '@angular/router';
+import { Observable } from 'rxjs';
+
+/**
+ * Redirects any nested routes from /app/myfiles/** to /app/private/**
+ */
+@Injectable({
+  providedIn: 'root',
+})
+export class MyfilesGuard implements CanActivate {
```
liam-lloyd, line 18: Why isn't this `MyFilesGuard`?

liam-lloyd (approved):

src/app/core/guards/myfiles.guard.ts
meisekimiu: This was just because of the filename I picked with the Angular CLI. I used "myfiles" to match what the URL looks like.

PR #155, opened 2022-10-06
Add New Share Banner

meisekimiu:
This implements a new share banner that appears on the share preview page. This includes implementing the design as well as implementing the functionality for triggering and dismissing the banner.

PER-8966: Create new "create account" banner that is closeable
PER-8968: Show new "create account" banner on desired actions

### DIFF
--- a/src/app/share-preview/components/share-preview/share-preview.component.ts
+++ b/src/app/share-preview/components/share-preview/share-preview.component.ts
@@ -22,7 +22,7 @@ import {
 } from '@shared/utilities/forms';
 import { AccountResponse, AuthResponse } from '@shared/services/api/index.repo';
 import { DeviceService } from '@shared/services/device/device.service';
-import { Subscription } from 'rxjs';
+import { Subscription, Subject } from 'rxjs';
 import { filter } from 'rxjs/operators';
 import { GoogleAnalyticsService } from '@shared/services/google-analytics/google-analytics.service';
 import { EVENTS } from '@shared/services/google-analytics/events';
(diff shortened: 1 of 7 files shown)


src/app/share-preview/components/share-preview/share-preview.component.ts
```diff
@@ -22,7 +22,7 @@ import {
 } from '@shared/utilities/forms';
 import { AccountResponse, AuthResponse } from '@shared/services/api/index.repo';
 import { DeviceService } from '@shared/services/device/device.service';
-import { Subscription } from 'rxjs';
+import { Subscription, Subject } from 'rxjs';
```
cecilia-donnelly, line 25: What does this new import do?

cecilia-donnelly: (just curious as ever)

cecilia-donnelly (approved):
This looks great! I confess I rushed through this review a bit but it makes sense to me. Thank you, @meisekimiu !

I know you were thinking about removing the old banner in the same PR but it looks like that's not here. Totally fine, just calling it out so we don't forget.

src/app/share-preview/components/share-preview/share-preview.component.ts
meisekimiu: `Subject` is an object that can be subscribed to in RxJS, and this is used so the share banner component can subscribe to this subject and receive events from its parent. This is apparently the "proper" way to do this kind of parent-child event communication in Angular!

PR #158, opened 2022-10-26
Show reCaptcha before verifying phone numbers

meisekimiu:
Show the re-captcha component before sending any verification text messages to the user.

To test, you will need to add a reCAPTCHA Site Key to your `.env` file under the `RECAPTCHA_API_KEY` key. You should be able to get a reCAPTCHA site key from [here,](https://www.google.com/u/1/recaptcha/admin) and if you are on the Permanent Engineering Team, you should already see a Permanent Web-App key to use with your environment. We will need to make sure to add the Permanent Web App key to our secrets before merging!

If you want to test the behavior without having to send text messages, you can set the `showCaptchaForEmail` property on the `VerifyComponent` to true, which will enable the same reCAPTCHA functionality for e-mails as well, which should be a lot easier to test with (since you can just add +'s to your e-mail address if it's already verified... that doesn't work as well with phone numbers!).

Steps to test:
1. Make sure you have a reCAPTCHA key in your `.env` file
2. Add a phone number to your account (or update your e-mail if you are using the `showCaptchaForEmail` functionality)
3. Click the verify phone number/e-mail button.
4. Verify that a reCAPTCHA is shown, and that the verification code is not actually sent until the CAPTCHA has been verified.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4316,6 +4316,11 @@
         "@types/node": "*"
       }
     },
+    "@types/grecaptcha": {
+      "version": "3.0.4",
+      "resolved": "https://registry.npmjs.org/@types/grecaptcha/-/grecaptcha-3.0.4.tgz",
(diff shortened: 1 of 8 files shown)


meisekimiu:
Yep that would be great!

PR #159, opened 2022-11-01
Update tag editting to handle custom metadata separately

liam-lloyd:
Currently, custom metadata is shown just like any other tag in the metadata sidebar of a file or folder. This commit creates a separate custom metadata section in that sidebar and updates how individual pieces of custom metadata are displayed to make clearer that they are key:value pairs.

Resolves PER-9020: Front-end: Implement design for displaying custom metadata

### DIFF
--- a/src/app/file-browser/components/edit-tags/edit-tags.component.html
+++ b/src/app/file-browser/components/edit-tags/edit-tags.component.html
@@ -2,8 +2,11 @@
   <pr-tags [tags]="itemTags" (click)="startEditing()" [readOnly]="false" [class.clickable]="canEdit && !isEditing"
     [canEdit]="canEdit" [animate]="isEditing" [isEditing]="isEditing"></pr-tags>
   <div class="input-group" *ngIf="isEditing" @ngIfScaleAnimation [class.has-tags]="allTags?.length">
-    <input type="text" class="new-tag form-control" placeholder="Add new tag"
-      [(ngModel)]="newTagName" (ngModelChange)="onTagType(newTagName)" (keydown.enter)="onInputEnter(newTagName)">
+    <input type="text" class="new-tag form-control" [placeholder]="placeholderText" [(ngModel)]="newTagName"
+    (ngModelChange)="onTagType(newTagName)" (keydown.enter)="onInputEnter(newTagName)">
+    <div *ngIf="newTagInputError" class="input-vertical-error">
+      {{this.tagType === 'keyword' ? "Tag cannot contain ':'" : "Format must be key:value"}}
+    </div>
     <div class="input-group-append">
       <button class="btn btn-primary" (click)="newTagName ? onInputEnter(newTagName) : endEditing()">{{ newTagName ? 'Add' : 'Done' }}</button>
     </div>
@@ -19,6 +22,10 @@
       </div>
     </div>
   </div>
+  <p class="manage-tags-message" *ngIf="isEditing" @ngIfScaleAnimation>
+    <span class="manage-tags-link" (click)="onManageTagsClick()">
+      Manage {{this.tagType === 'keyword' ? "tags" : 'custom metadata'}}</span> in archive settings.
+  </p>
 </ng-container>
 <ng-container *ngIf="isDialog">
   <div class="dialog-content">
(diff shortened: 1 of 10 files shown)


src/app/file-browser/components/edit-tags/edit-tags.component.html
```diff
@@ -2,8 +2,11 @@
   <pr-tags [tags]="itemTags" (click)="startEditing()" [readOnly]="false" [class.clickable]="canEdit && !isEditing"
     [canEdit]="canEdit" [animate]="isEditing" [isEditing]="isEditing"></pr-tags>
   <div class="input-group" *ngIf="isEditing" @ngIfScaleAnimation [class.has-tags]="allTags?.length">
-    <input type="text" class="new-tag form-control" placeholder="Add new tag"
-      [(ngModel)]="newTagName" (ngModelChange)="onTagType(newTagName)" (keydown.enter)="onInputEnter(newTagName)">
+    <input type="text" class="new-tag form-control" [placeholder]="placeholderText" [(ngModel)]="newTagName"
+    (ngModelChange)="onTagType(newTagName)" (keydown.enter)="onInputEnter(newTagName)">
+    <div *ngIf="newTagInputError" class="input-vertical-error">
+      {{this.tagType === 'keyword' ? "Tag cannot contain ':'" : "Format must be key:value"}}
+    </div>
```
liam-lloyd, line 9: This error wasn't in the design, so I just followed the convention used for validation errors in the signup form

src/app/file-browser/components/edit-tags/edit-tags.component.html
```diff
@@ -19,6 +22,10 @@
       </div>
     </div>
   </div>
+  <p>
+    <span class="manage-tags-link" (click)="onManageTagsClick()">
+      Manage {{this.tagType === 'keyword' ? "tags" : 'custom metadata'}}</span> in archive settings.
+  </p>
```
liam-lloyd, line 28: Currently this link goes to the existing manage tags modal, which for the moment is the place to manage both tags and custom metadata. I expect it will make sense to start differentiating where it goes as part of https://permanent.atlassian.net/browse/PER-9023

src/app/file-browser/components/edit-tags/edit-tags.component.ts
```diff
@@ -1,27 +1,47 @@
...
 import { ApiService } from '@shared/services/api/api.service';
 import { TagResponse } from '@shared/services/api/tag.repo';
 import { BaseResponse } from '@shared/services/api/base';
 import { MessageService } from '@shared/services/message/message.service';
 import { ngIfScaleAnimation } from '@shared/animations';
-import { DIALOG_DATA, DialogRef } from '@root/app/dialog/dialog.module';
+import { DIALOG_DATA, DialogRef, Dialog } from '@root/app/dialog/dialog.module';
 import { SearchService } from '@search/services/search.service';
 
+export type TagType = 'keyword' | 'customMetadata';
```
liam-lloyd, line 31: `keyword` is in anticipation of what's currently called "Tags" being renamed to "Keywords" in the UI. I think at that point "tag" can become a more generic term used in code to refer to the union of keywords and custom metadata (until these concepts eventually separate entirely and that union ceases to be relevant).

src/app/file-browser/components/edit-tags/edit-tags.component.html
meisekimiu: Hm... my interpretation of the design is that this link should only appear when the tag editor is actually open (otherwise it takes up a lot of space, especially with two links being present). But I think this is ambiguous; I can't actually find a design where the tag editors are closed. We should probably run this by the experience team to see what they think.

Also, a slight nitpick but I there should be a tiny bit of margin between this link and what is above it. It looks slightly off compared to the design since it seems a bit to close to the tag editor. (Though again, the exp team can approve/disapprove how it looks here!)

src/app/file-browser/components/edit-tags/edit-tags.component.ts
```diff
@@ -33,11 +53,15 @@ export class EditTagsComponent implements OnInit, DoCheck, OnDestroy, HasSubscri
 
   public isEditing = false;
 
+  public newTagInputError = false;
```
meisekimiu, line 56: Can we reset this value when you open/close the tag editor? I feel like once the error appears, I should be able to make it go away by just closing the editor, but when I re-open it, the error is still there and it feels like something's wrong.

On a related note, when this error is being displayed, you can switch to another file in your archive, open the tag editor on it, and the error will still be present which I think definitely shouldn't be the case.

meisekimiu (changes requested):
This looks good! Thanks for working on this! It works great in testing, and tags even show up properly in other places such as public archives:



Can you show some screenshots to the experience team? Including the tag editors opened, closed, and with error messages.

The code looks good, minus the small bug I mentioned with the `newTagInputError` value, which I think should be fixed.

meisekimiu (approved):
Thanks for the fixes, this looks good!

PR #166, opened 2022-11-19
Wrap tags and custom metadata within colored background "pills"

liam-lloyd:
Currently, when tags or custom metadata are very long they wrap, causing their colored background padding to overlap in an odd-looking way. This commit resolves the issue by using `<p>` elements instead of `<span>`s, so the text wraps within a taller block of background padding if it is long enough to wrap.

### DIFF
--- a/src/app/file-browser/components/file-viewer/file-viewer.component.html
+++ b/src/app/file-browser/components/file-viewer/file-viewer.component.html
@@ -1,28 +1,59 @@
-<div class="file-viewer" [ngClass]="{'minimal': useMinimalView, 'editing-date': editingDate, 'can-edit': canEdit}">
+<!-- @format -->
+<div
+  class="file-viewer"
+  [ngClass]="{
+    minimal: useMinimalView,
+    'editing-date': editingDate,
+    'can-edit': canEdit
+  }"
+>
   <div class="file-viewer-close">
-    <a (click)="close()">
-      <i class="ion-md-arrow-dropleft"></i> Back
-    </a>
+    <a (click)="close()"> <i class="ion-md-arrow-dropleft"></i> Back </a>
   </div>
   <div class="thumb-target">
-    <div class="thumb-wrapper" *ngFor="let record of records; let i = index;" [hidden]="!isQueued(i)"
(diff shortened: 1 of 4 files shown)


cecilia-donnelly (commented):
Hmm this still doesn't wrap for me on Firefox unless I'm missing something about where this supposed to show up.

Here it is on the public version of the item:
Screenshot 2022-11-23 at 15-55-57 Public Permanent org

cecilia-donnelly:
And here is how it looks on main in Firefox:

Screenshot 2022-11-23 at 16-01-30 Public Permanent org

cecilia-donnelly:
Okay, I see! It looks like this works better on the "edit" panel with tags that have spaces:

Screenshot 2022-11-23 at 16-05-33 Public Permanent org

It still looks a bit weird on the public version - I'm not sure why the "quickkey" custom metadata is being pushed off the side:

Screenshot 2022-11-23 at 16-04-34 Public Permanent org

But it looks similar in Chrome, so at least this doesn't appear to be a browser-specific issue!

liam-lloyd:
@cecilia-donnelly looks like these oddities come from the fact that many of the children of this sidebar believe they have more width available than they actually do. I'll figure out how to fix that. And should we also wrap long strings without separators? Or is that a question for Kaitlyn?

liam-lloyd:
@cecilia-donnelly I've fixed the issue where the text goes off the edge of the window.

@cecilia-donnelly @k8lyn6 I've also added what seems to me the most reasonable handling of long tags with no natural breakpoints (though I'm happy to do something else if you have a different idea Kaitlyn, I just figured I'd put this in for the moment because it turned out to be fairly simple). This is to break them and display a hyphen the way books do on broken words. The setting I've chosen for this _should_ avoid breaking words where possible, but in practice this appears to be browser dependent; Chrome breaks words unnecessarily sometimes, Firefox does not (at least, not that I've seen so far).

Some long tags (where by "tags" I mean both keywords and custom metadata) in public gallery view
long_tags_public

Long tags in the archive manager on Chrome (note unnecessarily broken word)
long_tags_archive_manager

Long tags in the archive manager on Firefox
long_tags_archive_manager_firefox

k8lyn6:
@liam-lloyd I think that looks good!

PR #168, opened 2022-11-22
Restrict tag service cache to tags from the current archive

liam-lloyd:
The TagsService is intended to keep track of all tags that exist in the current archive of a logged-in user. It updates its cache of tags when a new item is viewed in case that item has a recently added tag. However, this update also happens when viewing items in a public archive, and tags from such items are added to TagsService cache of tags. This will cause those tags to appear as options when a user subsequently returns to their own archive and tries to add a tag to an item, even though those tags don't actually exist in that archive. This commit fixes that bug by updating TagsService to check the archiveId of tags before adding them to its cache, and not add them if they don't match the current archive.

Resolves: https://permanent.atlassian.net/browse/PER-9082

### DIFF
--- a/src/app/core/services/tags/tags.service.spec.ts
+++ b/src/app/core/services/tags/tags.service.spec.ts
@@ -1,16 +1,40 @@
-// import { TestBed } from '@angular/core/testing';
+/* @format */
+import { TestBed } from '@angular/core/testing';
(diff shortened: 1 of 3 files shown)


meisekimiu (approved):
Thank you for writing new tests!

PR #169, opened 2022-11-22
Center banner images on public gallery page

liam-lloyd:
Resolves: https://permanent.atlassian.net/browse/PER-8985


cecilia-donnelly (approved):
Hurray!

PR #171, opened 2022-12-08
Create FusionAuth and Permanent accounts as separate steps on signup

liam-lloyd:
Currently, this project creates Permanent accounts by calling /account/post with a V1 request. This causes /account/post to create both a FusionAuth account and a Permanent account. Handling these in a combined endpoint can cause data integrity problems, however, so the backend now provides an /auth/createCredentials endpoint to create a FusionAuth account, and creates only a Permanent account on receiving a V2 request to /account/post. This commit updates signup functionality to use these new features to decouple FusionAuth and Permanent account creation.

Resolves: https://permanent.atlassian.net/browse/PER-8812
Do not merge without https://github.com/PermanentOrg/back-end/pull/353

### DIFF
--- a/src/app/shared/services/account/account.service.spec.ts
+++ b/src/app/shared/services/account/account.service.spec.ts
@@ -1,187 +1,112 @@
-// import { TestBed, inject } from '@angular/core/testing';
-// import { HttpClientTestingModule, HttpTestingController } from '@angular/common/http/testing';
-// import { CookieService } from 'ngx-cookie-service';
+/* @format */
+import { TestBed, inject } from '@angular/core/testing';
+import { Router } from '@angular/router';
+import { Shallow } from 'shallow-render';
+import { Observable } from 'rxjs';
 
-// import { last } from 'rxjs/operators';
-// import { concat } from 'rxjs';
+import { AccountService } from '@shared/services/account/account.service';
+import { ApiService } from '@shared/services/api/api.service';
+import { AuthResponse } from '@shared/services/api/index.repo';
+import { CreateCredentialsResponse } from '@shared/services/api/auth.repo';
+import { AppModule } from '../../../app.module';
+import { AccountVO, ArchiveVO } from '@root/app/models';
 
-// import { AccountService } from '@shared/services/account/account.service';
-// import { TEST_DATA } from '@core/core.module.spec';
-// import { AuthResponse } from '@shared/services/api/auth.repo';
-// import { AccountVO, ArchiveVO } from '@root/app/models';
-// import { environment } from '@root/environments/environment';
-// import { AccountResponse } from '@shared/services/api/index.repo';
-// import { StorageService } from '@shared/services/storage/storage.service';
-// import { RouterTestingModule } from '@angular/router/testing';
+describe('AccountService', () => {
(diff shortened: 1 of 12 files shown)


src/app/shared/services/api/account.repo.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    repo = new AccountRepo(TestBed.get(HttpService));
+    httpMock = TestBed.get(HttpTestingController);
+  });
+
+  afterEach(() => {
+    httpMock.verify();
+  });
+
+  it('signUp should call /account/post', () => {
```
liam-lloyd, line 31: I wanted to use `shallow-render` for this test, but it appeared that some oddity about our organization of NgModules prevented it from finding my mocked HttpClient (or the HttpTestingController, which I also tried), so I did it in the style of existing API repo tests.

src/app/shared/services/account/account.service.spec.ts
```diff
@@ -1,187 +1,112 @@
...
-//         fail();
-//       })
-//       .catch((response: AccountResponse) => {
-//         expect(response.isSuccessful).toBeFalsy();
-//       });
-
-//     const req = httpMock.expectOne(`${environment.apiUrl}/account/post`);
-//     req.flush(expected);
-//   });
-// });
```
cecilia-donnelly, line 187: Thank you!

src/app/shared/services/account/account.service.ts
```diff
@@ -201,9 +221,7 @@ export class AccountService {
 
   public async hasOwnArchives() {
     const archives = await this.refreshArchives();
-    const ownArchives = archives.filter(
-      (archive) => !archive.isPending()
-    );
+    const ownArchives = archives.filter((archive) => !archive.isPending());
```
cecilia-donnelly, line 224: I know you only changed the formatting here and I haven't followed the whole code path, but it seems like we should not return "hasOwnArchives" true if the user only has deleted archives.

cecilia-donnelly: (i.e., maybe we should check for archives in a good state instead of specifically excluding pending archives?)

src/app/shared/services/http/http.service.ts
```diff
@@ -36,37 +38,56 @@ export class HttpService {
...
+          'Content-Type': 'application/json',
+        },
+      })
+      .pipe(
+        map((response: any) => {
+          if (response?.csrf) {
+            this.storage.session.set(CSRF_KEY, JSON.stringify(response.csrf));
+          }
+          if (responseClass) {
+            return new responseClass(response);
```
cecilia-donnelly, line 86: Nice.

cecilia-donnelly (approved):
This is great, thank you Liam!

As I mentioned in https://github.com/PermanentOrg/web-app/pull/174#pullrequestreview-1225223669, it would be helpful to separate out formatting changes from substantive changes. I appreciate your formatting efforts, though! Looks nice.

src/app/shared/services/account/account.service.ts
cecilia-donnelly: (But obviously that shouldn't be part of this PR!)

PR #174, opened 2022-12-16
Separate keywords from custom metadata in public archive file viewer

liam-lloyd:
Now that keywords and custom metadata are handled separately in the archive manager, we want to do the same in the public archive file viewer.


cecilia-donnelly (approved):
@meisekimiu might have more specific feedback, but this makes sense to me! I could also see it work locally. Thanks for including tests :) 

One thing that would make this kind of work clearer to me and future readers is to separate out the formatting changes into their own commit - the actual change here is fairly small as I see it, but it was a little difficult to find the meaningful diff among all the changes.

PR #181, opened 2023-02-01
Add giveaway banner in app to match marketing site

cecilia-donnelly:
Tagging @Fenn-CS as a reviewer since @meisekimiu is out today and ideally this would get deployed tomorrow morning.

Natalie, I'm not totally sure I got the timezone right on the end date. I followed what you used for our previous banner. Thanks so much for setting that up! @k8lyn6 asked if I could put this up so that people who navigate directly into the app still see the announcement.

### DIFF
--- a/src/app/announcement/data/events.ts
+++ b/src/app/announcement/data/events.ts
@@ -2,9 +2,9 @@ import { AnnouncementEvent } from '../models/announcement-event';
 
 export const ANNOUNCEMENT_EVENTS: AnnouncementEvent[] = [
   {
-    start: new Date('2022-10-18T20:00:00-05:00').getTime(),
-    end: new Date('2022-10-25T20:10:00-05:00').getTime(),
+    start: new Date('2023-01-31T20:00:00-05:00').getTime(),
+    end: new Date('2023-02-15T23:59:00-05:00').getTime(),
     message:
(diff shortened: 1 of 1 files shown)


Fenn-CS (approved):
Reviewed by looking at the existing/changed "announcement event" and everything looks good.

Side note: It's already February 1st for me and would 2nd in a few hours, this time is probably in the past?

src/app/announcement/data/events.ts
```diff
@@ -2,9 +2,9 @@ import { AnnouncementEvent } from '../models/announcement-event';
 
 export const ANNOUNCEMENT_EVENTS: AnnouncementEvent[] = [
   {
-    start: new Date('2022-10-18T20:00:00-05:00').getTime(),
-    end: new Date('2022-10-25T20:10:00-05:00').getTime(),
+    start: new Date('2023-01-31T20:00:00-05:00').getTime(),
```
Fenn-CS, line 5: Side note: It's already February 1st for me and would 2nd in a few hours, this time is probably in the past?

cecilia-donnelly: Yes, I figured I'd do a time in the past so it would appear as soon as I deploy in the morning. Thanks for checking! :)

PR #184, opened 2023-02-07
Replace `@permanentorg/vis-timeline` with `vis-timeline`

meisekimiu:
This PR uninstalls [our fork of vis-timeline](https://github.com/PermanentOrg/vis-timeline) and replaces it with the standard `vis-timeline` package. Somehow, the old `@permanentorg/vis-timeline` package was also supplying various other dependencies (how did that work???), so this PR also explicitly installs those as dependencies.

The main code addition is mainly a hacky workaround to recreate the styling that our fork added to the vis-timeline. There were 2 main functionalities we added to the library: A customizable `pixelMargin` value that adds margin to the sides of the screen when the library picks an initial zoom level, and a `center` orientation for positioning the lines in the timeline.

The `pixelMargin` addition is fixed by adjusting each timeline item to have a higher width so that vis-timeline ends up picking the same zoom level. The line centering is fixed by running the resize event on the timeline (I'm not sure why this works but it fixes it).


src/app/views/components/timeline-view/timeline-view.component.ts
```diff
@@ -498,23 +552,40 @@ export class TimelineViewComponent implements OnInit, AfterViewInit, OnDestroy {
...
       const groups = this.timelineGroups.get(breadcrumb.timespan);
       const group = find(groups, (g: TimelineGroup | TimelineItem) => {
-        return g.dataType === 'group' && g.content === breadcrumb.text ;
+        return g.dataType === 'group' && g.content === breadcrumb.text;
       });
       this.onGroupClick(group as TimelineGroup);
     }
   }
 
+  protected addPixelMargin(): void {
```
liam-lloyd, line 580: Just out of curiosity, what is it that forces us to include the `@ts-ignore`s here?

src/app/views/components/timeline-view/timeline-view.component.ts
```diff
@@ -498,23 +552,40 @@ export class TimelineViewComponent implements OnInit, AfterViewInit, OnDestroy {
...
+    // @ts-ignore: This is a hack to bypass maintaining a fork of vis-timeline
+    for (const uuid in this.timeline.itemSet.items) {
+      if (uuid.includes('-')) {
```
liam-lloyd, line 583: Also out of curiosity, why is this our condition?

liam-lloyd (approved):
I have a couple questions out of curiosity, but this looks good, and timeline view seems to still work.

PR #185, opened 2023-02-08
Remove createCredentials; do signup in one API call

cecilia-donnelly:
I put this together quickly so it's very possible I missed something, but the tests passed for me locally and I was able to sign up from the web app, so fingers crossed. This should be the front end counterpart to https://github.com/PermanentOrg/back-end/pull/406.


meisekimiu (approved):
Looks good, and I also got it working with the corresponding back-end branch! :+1:

liam-lloyd (approved):
Looks good to me too!

PR #187, opened 2023-03-10
Add Directives Display Screen Stories

meisekimiu:
This adds stories for the Directives Display Screen, as well as all of the corresponding code that goes along with it. There are a few parts of the design that will be implemented in the future, such as showing a proper edit button if a directive already exists for the user. Also the embedded links don't link anywhere right now.

This commit also introduces the Directive model, and the DirectiveRepo in the API Service. Currently the DirectiveRepo is unimplemented and will be hooked up to actual API endpoints in a future ticket.

One added feature that is not in the designs is an error message that shows up if API requests fail. Currently you can view this with the "API Error" story on Storybook.

**Stepz 2 Test:**
1. Run `npm run storybook`
2. View one of the "Directives Display" stories.
3. Play around with the controls on the story!

PR #189, opened 2023-03-13
Show "No Plan" Warning on Directives Display

meisekimiu:
This PR is building off of the work from #187.

In the design for the Directives Display screen, there is a warning that shows up and prohibits the user from continuing if the current account does not have a Legacy Contact set up. This commit includes the code to show that warning.

We've decided that this UI functionality may end up disabled when the feature goes live. As a result, all of this functionality can be switched on or off with an input in the `DirectiveDisplayComponent`.

**Steps to Test:**
1. Run `npm run storybook`
2. Check the "No Plan" story, and verify that the warning displays properly.
4. Toggle the **checkLegacyContact** property and verify that enables/disables the warning.
5. Verify the warning does not display in other stories.

### DIFF
--- a/src/app/directive/components/directive-display/directive-display.component.html
+++ b/src/app/directive/components/directive-display/directive-display.component.html
@@ -21,7 +21,19 @@
     <a href="#">here.</a>
   </p>
   <div class="archive-steward-header">Archive Steward: {{ archiveName }}</div>
-  <!--no plan warning will go here-->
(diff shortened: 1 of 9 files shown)


cecilia-donnelly (approved):
This looks great. I know you'll be rebasing on main when the other PR is merged anyway, but it looks like you ended up with a double "Expect API to return Steward Email in Directive" commit, so look out for that.

PR #190, opened 2023-03-13
Run Unit Tests on all PRs

meisekimiu:
The unit tests in #189 are not running since it's not currently set up to merge into `main`. But I'd still like to have unit tests run on this PR. This PR updates the workflow to run on all unit tests.

This PR is set up to merge into `directives-display` so we can test the functionality, but it also might be useful to merge it in that branch anyways so that the changes can make their way to #189.


cecilia-donnelly (approved):
Makes sense to me!

PR #192, opened 2023-03-14
Per 9157 reset redirect

cecilia-donnelly:
Redirect from the web app to FusionAuth, instead of linking directly to FA from "forgot password" emails. To test, click "forgot password" on local, see that the email you receive has an "ng.permanent.org:4200" link in it but redirects to FusionAuth for you to type in a new password.

I also need to update the FA template repo. This doesn't need to block on that, though, because the existing direct-to-FusionAuth link will not be affected by this change.

### DIFF
--- a/src/app/app.routes.ts
+++ b/src/app/app.routes.ts
@@ -78,6 +78,16 @@ const routes: RoutesWithData = [
       { path: 'app/forgot', redirectTo: 'app/auth/forgot', pathMatch: 'full' },
       { path: 'app/reset', redirectTo: 'app/auth/reset', pathMatch: 'full' },
       {
+        path: 'app/fa-reset',
+        loadChildren: () =>
+        new Promise( () => {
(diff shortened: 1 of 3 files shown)


src/app/app.routes.ts
```diff
@@ -40,6 +41,8 @@ export interface RouteWithData extends Route {
 
 export type RoutesWithData = RouteWithData[];
 
+const fusionauth_host = SecretsService.getStatic('FUSIONAUTH_HOST');
```
meisekimiu, line 44: Little nitpick here: Can we camelCase this variable name?

meisekimiu (approved):
This looks good to me! I verified that the URL sends me to FusionAuth dev with no secret, and sends me to whichever domain I put into the secret.

src/app/app.routes.ts
cecilia-donnelly: done!

PR #193, opened 2023-03-16
Per 8925 fix long archive name in left side nav

crisnicandrei:
When the text will be over 3 lines long it will automatically display the "...", and if the user wants to see it fully, I have added a hover functionality which will display the whole text, making sure to not overlap the next component

### DIFF
--- a/src/app/core/components/left-menu/left-menu.component.html
+++ b/src/app/core/components/left-menu/left-menu.component.html
@@ -19,7 +19,7 @@
           <div class="archive-name-access">
             <div class="archive-options-toggle" (click)="toggleArchiveOptions()"
             [class.reverse]="showArchiveOptions">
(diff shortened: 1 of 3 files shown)


cecilia-donnelly (commented):
This makes sense to me and the outcome looks good. My main feedback for future changes is to separate formatting changes from substantive changes. It looks to me like your IDE made some automatic adjustments. That's fine as long as its settings match the project's linting setup (and I'll let @meisekimiu weigh in on that). However, having those formatting changes in the same commits as changes that affect the project makes it harder to review. Can you keep formatting changes in their own commit or commits in the future? Thank you!

crisnicandrei:
> 

@cecilia-donnelly yes! noted! I will change my prettier settings to make sure it doesn't format on save anymore!

meisekimiu (approved):

PR #194, opened 2023-03-16
Per 8668 add public gallery link to left side menu bar

crisnicandrei:
Added a new route to the left menu component which redirects the user to the public gallery


cecilia-donnelly (approved):
Looks good; glad you were able to find the icon. Thank you for the clear commit messages! There are some formatting conventions we try to follow for them, notably keeping a 50 char limit on the first line. [This article](https://cbea.ms/git-commit/) is a good outline of what we strive for in our repos.

PR #196, opened 2023-03-17
Per 8829 allow changing archive type

crisnicandrei:
This pr serves as the implementation of the feature for changing the archive type.

This is done from the "Archive Settings" tab, under the show download button radio buttons.

Upon selection of a new type, a dialog appears asks for the user's confirmation. If the user does not confirm, the dialog closes, and the previous type is still set


src/app/core/components/archive-type-change-dialog/archive-type-change-dialog.component.ts
```diff
@@ -0,0 +1,48 @@
...
+    this.dialogRef.close();
+  }
+
+  public async onConfirmClick() {
+    this.archive.type = this.archiveType;
+    this.updating = true;
+    try {
+      await this.api.archive.update(this.archive);
+    } catch {
+      // fail silently
```
cecilia-donnelly, line 42: I think ideally we'd tell the user that an error occurred, though I defer to @meisekimiu on how we usually communicate that.

src/app/core/components/public-settings/public-settings.component.html
```diff
@@ -32,4 +32,13 @@
     Display a button that allows viewers of your published items to download a
     copy of the source file to their device.
   </small>
+  <div class="settings-item">
+   <label>Archive Type</label>
+   <select [(ngModel)]="archiveType" name="archiveType" id="archiveType" (ngModelChange)="onArchiveTypeChange()" [disabled]="archive.type === 'type.archive.nonprofit'" class="custom-select">
```
cecilia-donnelly, line 37: I know this is a little tricky since we don't have a specific design, but I don't think this version will work. It seems too confusing for a user looking at settings for a nonprofit archive -- "why is the type blank?" they'd be wondering. My idea would be to not show the dropdown at all and just show the word "Nonprofit" in that case. Open to other ideas!

cecilia-donnelly (commented):
Leaving this to @meisekimiu for final review, but here are some thoughts as I looked through it. Thanks for putting this together!

src/app/core/components/public-settings/public-settings.component.html
crisnicandrei: I think that is a great idea! The user might see a disabled dropdown and think there is something wrong with their account. I will change it now :)

src/app/core/components/archive-type-change-dialog/archive-type-change-dialog.component.html
```diff
@@ -0,0 +1,19 @@
+<div class="dialog-content">
+  <div class="header">
+    <span>My Archives</span>
+    <button class="btn" (click)="onDoneClick()">
+      <i class="material-icons">close</i>
+    </button>
+  </div>
+  <div class="'content">
```
meisekimiu, line 8: There's an extra `'` in this class name!

src/app/core/components/public-settings/public-settings.component.ts
```diff
@@ -12,11 +15,32 @@ export class PublicSettingsComponent implements OnInit {
   @Input() public archive: ArchiveVO;
   public updating: boolean = false;
   public allowDownloadsToggle: number = 0;
+  public supportLink:string = 'https://permanent.zohodesk.com/portal/en/newticket'
```
meisekimiu, line 18: Can you run prettier on this file? This line sticks out to me a bit, and I think prettier should have fixed that.

src/app/core/components/archive-type-change-dialog/archive-type-change-dialog.component.ts
meisekimiu: I think ideally we would at least like to call the `MessageService` to show an error in the UI (though no other action is really necessary).

meisekimiu (commented):
This looks pretty good! I should mention we already have a built in confirmation box (available via the `PromptService.confirm` method) that could have been used instead of a custom component. Overall, I'm fine with separating it into its own component for now since that at least also separates out the logic from the bigger Archive Settings Dialog component.

I did leave some comments on formatting which are pretty small.

src/app/core/components/public-settings/public-settings.component.ts
crisnicandrei: done

crisnicandrei:
@meisekimiu did all the changes! Waiting for you to look again :)

meisekimiu (approved):

PR #197, opened 2023-03-24
Added an extra route param

crisnicandrei:
Added an extra route param so when the user hits for example storage/promo then the promo tab will be selected, but if the route hits only storage then the first tab will be selected by default, this case also works with storage/add

The new tabs are:

/storage => storage with first tab open
/storage/add => same as above
/storage/promo => the redeem code tab
/storage/transaction => transaction history
/storage/file => file history.

### DIFF
--- a/src/app/core/components/left-menu/left-menu.component.html
+++ b/src/app/core/components/left-menu/left-menu.component.html
@@ -132,7 +132,7 @@
         Invitations
       </a>
       <a class="menu-item hide-desktop"
-        [routerLink]="[{ outlets: { dialog: ['storage']}}]">
+        [routerLink]="[{ outlets: { dialog: ['storage','add']}}]">
         <i class="ion-md-add-circle"></i>
         Add Storage
       </a>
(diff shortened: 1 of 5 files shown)


src/app/core/components/left-menu/left-menu.component.html
meisekimiu, line 1: Looks like prettier ran on this file even if there were no changes in the pull request. Since this file is currently creating a merge conflict, can you remove it from the commits in this PR (or rebase this branch onto main and fix the conflicts)?

src/app/core/components/nav/nav.component.html
meisekimiu, line 1: This file also doesn't actually have any relevant changes besides formatting. Though it isn't creating a merge conflict like the `left-menu.component.html` file, it would also be nice to remove this from the PR as well.

src/app/core/components/storage-dialog/storage-dialog.component.ts
```diff
@@ -1,9 +1,14 @@
+import { ActivatedRoute, ParamMap } from '@angular/router';
+// import { RouterStateSnapshot } from '@angular/router';
```
meisekimiu, line 2: Can you delete this commented out line?

meisekimiu (approved):
This looks good! There is a merge conflict as a result of prettier running on some files, so you'll have to deal with that before merging, but otherwise this looks good, and I tested it with each tab in the storage dialog! :+1:

PR #199, opened 2023-04-03
Per 4817 zoom functionality

crisnicandrei:


crisnicandrei:
@meisekimiu did what you asked! Was not that hard imo :) Waiting for your opinion!

src/app/file-browser/components/file-viewer/file-viewer.component.ts
```diff
@@ -222,6 +227,8 @@ export class FileViewerComponent implements OnInit, OnDestroy {
 
     const queuedThumbs = document.querySelectorAll('.thumb-wrapper.queue');
 
+    console.log('here')
```
meisekimiu, line 230: You missed a `console.log` here!

src/app/shared/components/thumbnail/thumbnail.component.ts
```diff
@@ -39,19 +41,23 @@ export class ThumbnailComponent
   private imageElement: Element;
   private resizableImageElement: Element;
 
+  private INITIAL_ZOOM_LEVEL = 0.2963835165168812;
```
meisekimiu, line 44: Does this variable end up getting used at all? If not can we delete it?

src/app/shared/components/thumbnail/thumbnail.component.ts
```diff
@@ -39,19 +41,23 @@ export class ThumbnailComponent
...
+
   private targetThumbWidth: number;
   private currentThumbWidth = 200;
   private currentThumbUrl: string;
   private dpiScale = 1;
 
   private lastItemDataStatus: DataStatus;
+  private initialZoom = 0;
```
meisekimiu, line 52: Same with this variable; it seems unused as well.

meisekimiu (commented):
Thank you for making this change! This works way smoother now!

I do have one more question about functionality that's similar to the last inquiry I had. If it ends up being too difficult we can always split it off into its own ticket!

Anyway, when I am zoomed out on an image and I do the swiping motion, I can also still pan around in the OpenSeaDragon viewer, which makes the image move in an unexpected way: it's both panned around in the viewer *and* moved via the swiping animation. I'd expect it to only swipe and not also pan if zoomed out. In other words, if swiping is disabled, we should be able to pan around in OpenSeaDragon, and if swiping is enabled we shouldn't be able to pan around. Can you try investigating to see whether it's possible to fix this as well?

src/app/shared/components/thumbnail/thumbnail.component.ts
crisnicandrei: that is the initial zoom level which i forgot about, sorry! I can always check if the zoom level is equal to the initial one then we can swipe the image, otherwise we can pan, what do you think?

crisnicandrei:
@meisekimiu done!

meisekimiu (approved):
Thanks! Can you spin off a ticket for the issue I noted in my previous comment after merging?

PR #201, opened 2023-04-12
Per 9184 build results page after ssearch

crisnicandrei:
This pull request adds the new design to the search results -> please be advised that this is still work in progress but i would like some input about the progress so far while i am waiting for the remaining icons 

This is the second PR that should be reviewed for the public search. 

This is still using dummy data so the search does not work on this pr yet

PR #202, opened 2023-04-12
Implement api search

crisnicandrei:
This serves as adding the api for the new search functionality

Please note that this is not the final version, but I am open to changes while i am waiting for the final icons so this does not stall :)

this is the final pr in the chain of prs for the public archive search task. This connects the api call to the page and displays the results


src/app/shared/services/api/search.repo.ts
```diff
@@ -1,39 +1,78 @@
...
+  ): Observable<SearchResponse> {
     const data = {
-      SearchVO: {
-        query,
-        TagVOs: tags,
-        numberOfResults: limit
-      }
+      query,
+      tags:'',
+      archiveId: 13,
```
cecilia-donnelly, line 68: I know this is in progress, but it looks like something hardcoded was left here by accident. Just mentioning it so it doesn't sneak through.

crisnicandrei: thank you!

cecilia-donnelly:
@crisnicandrei it looks to me like the unit tests are failing now, not a linting error as we talked about yesterday. Can you fix those? (It looks like you maybe forgot an import or something simple. :crossed_fingers: )

meisekimiu (changes requested):
Can you try to get unit tests working on the Archive Search component? If it ends up not working, then at the very least instead of leaving a file full of commented out code, we can just delete the file instead. To me it's a bit concerning that even just the default "it should create" test is failing, though that is most likely caused by some of the auto-generated testing set up. I think ideally I'd like this component to have a properly set up suite of tests.

One other thing we realized only after this ticket was started is that the ApiService repos and HttpService do not properly have code to call V2 GET endpoints! You added an implementation here (thanks for that! :+1: ), but one thing we've been planning on doing as we implement the Directives API is restructuring the HttpService code that gets the V2 calls. It's currently based heavily on the V1 API code that takes in the request and response VO objects, but we don't need to structure around that anymore.

As a result, I think we should wait on the Directives API code to be implemented and then rebase this series of branches onto main once it's merged. This will mean that some of the actual API code you've written will probably have to be changed (sorry about that!) but it should also simplify a lot of this code since the general V2 request methods will already be defined.

crisnicandrei:
@meisekimiu I have merged this branch with yours and implemented the new api that you have used. I also had to run npm install because of the new angular version that was not yet on this branch. And also, in the function you can see that I have used an <any> because I was gettings some errors when using SearchResponse as an interface and I wanted your input regarding that. Hope this looks good. Please let me know what you think! thanks

meisekimiu (changes requested):
Instead of merging it would have been better to rebase onto main instead. Otherwise this PR has 93 commits that make it now really hard to review. Because of the multiple branches used here, I think it'd specifically be advisable to rebase #198 onto the current version of main, since this branch will be merged into #201 which will then merge into #198's branch.

src/app/gallery/components/gallery/gallery.component.ts
```diff
@@ -23,6 +23,7 @@ export class GalleryComponent implements OnInit {
       SecretsService.hasStatic('FEATURED_ARCHIVES')
     ) {
       const envFeatured = SecretsService.getStatic('FEATURED_ARCHIVES');
+      console.log('envFeatured', envFeatured)
```
meisekimiu, line 26:
```suggestion
```
Leftover `console.log`

src/app/shared/services/api/search.repo.ts
```diff
@@ -62,18 +63,18 @@ export class SearchRepo extends BaseRepo {
...
+  ) {
     const data = {
       query,
       tags:'',
       archiveId,
       publicOnly:true
     };
 
-    return this.http.getV2Request<SearchResponse>(
+    return getFirst(this.httpV2.get<any>(
```
meisekimiu, line 74: Can you avoid using the `any` type here? Also, can  you update the call to `httpV2.get` to include an options parameter to disable passing in the Auth Token? It would look something like this:
```js
this.httpV2.get(
  '/search/folderAndRecord',
  data,
  null,
  {
    authToken: false
  }
)
```

This makes sure that if the `AuthToken` used in certain V2 requests is defined, it won't be passed into the request. Since I was working on Directives stuff earlier, I had the `AuthToken` defined and this specific endpoint will respond with a 401 error if you pass in an authorization token.

src/app/shared/services/http/http.service.ts
```diff
@@ -21,7 +21,7 @@ export class HttpService {
   constructor(private http: HttpClient, private storage: StorageService) {}
 
   public sendRequest<T = BaseResponse>(
-    endpoint: string,
+    endpoint: string,x§
```
meisekimiu: This seems like an error.

meisekimiu (changes requested):
I tested the search and it does work! I have found a few problems, but this is looking good.

crisnicandrei:
@meisekimiu do you have any idea why that test is failing?

meisekimiu:
Unit tests should be fixed when #223 is merged.

meisekimiu:
This PR has been combined with the other public search PRs into #226 for convenience. We'll close this PR once that one gets merged, but I want to keep this PR open just in case any commits/issues got lost in the process of combining all of the PRs.

meisekimiu:
Closing this since #226 was merged.

PR #204, opened 2023-04-20
Display text on hover when it overflows

crisnicandrei:


src/app/file-browser/components/file-list-item/file-list-item.component.scss
```diff
@@ -116,16 +116,31 @@ $bg-transition-length: 0.075s;
...
+  }
+
   > div {
     display: flex;
     line-height: 1.25;
     max-width: 100%;
 
-    span {
-      flex: 1 1 auto;
+    .article-name {
```
meisekimiu, line 128: Is `article-name` the best name for this class? I feel like something like `item-name` might make more sense, though ultimately it doesn't matter since this classname is only used here.

meisekimiu (commented):
This looks good, and the solution makes perfect sense! Since this is a style fix I would like to get this fix approved by the experience team before merging.

I'm noticing in Google Chrome that when hovering over text that fits in one line I still see a scrollbar appear next to the names, even if it doesn't do anything (this doesn't appear in Firefox though!). I do think we might want to fix that.

src/app/file-browser/components/file-list-item/file-list-item.component.scss
crisnicandrei: i think you are right!

cecilia-donnelly:
@crisnicandrei fyi when @meisekimiu says this needs experience team approval we usually take screenshots or a video and share those in slack, tagging @k8lyn6 .

crisnicandrei:
@meisekimiu I have pushed a new commit which hides the scrollbar but keeps the functionality. Let me know what you think!

meisekimiu (approved):
Thanks for fixing the scrollbar issue!

PR #207, opened 2023-04-24
Replaced My files with private

crisnicandrei:


meisekimiu (changes requested):
This currently doesn't work once I'm inside a nested folder. You can see here that the display still says "My Files > New Folder" instead of "Private > New Folder"



Additionally, instead of just checking for the display name "My Files" can we check if the folder is the private root directly? The `FolderVO` object that represents the "My Files" directory we want to rename has a property called `type` that is always going to be equal to `"type.folder.root.private"`, no other folders are going to be assigned this type so checking for this will only ever match the private root. Otherwise it will affect any other folders that happen to be called "My Files".

You can see in this image I am in a subfolder in my archive I named "My Files" but it shows up as "Private". (And due to the other breadcrumb error it ends up creating the effect of reversing the two names, this should say "Private > My Files" and not the other way around)

crisnicandrei:
@meisekimiu thank you for your comment! I was not aware of that and I will make the change :)

src/app/shared/services/api/folder.repo.ts
```diff
@@ -40,8 +40,18 @@ export class FolderRepo extends BaseRepo {
...
+    console.log(folderVO)
+    let response = {
+      ...folderVO
+    }
+    if(folderVO.type === 'type.folder.root.private'){
+      response.displayName = 'Private'
+    }
+
+    console.log(response)
+
```
meisekimiu, line 52:
```suggestion
    const response = {
      ...folderVO
    };
    if (folderVO.type === 'type.folder.root.private') {
      response.displayName = 'Private';
    }

```
There are some leftover console.log statements here. Also, since `response` is never reassigned, let's go ahead and use `const` instead of `let` to define it.

meisekimiu (approved):
Thanks for changing this! This looks good to me now!

PR #208, opened 2023-04-25
Per 8879 navigate with tab

crisnicandrei:
This pr implements the tab navigation for the metadata fields.

The first commit is basically useless because the whole implementation is contained in the second one


src/app/file-browser/components/sidebar/sidebar.component.ts
```diff
@@ -33,11 +34,12 @@ export class SidebarComponent implements OnInit, OnDestroy, HasSubscriptions {
   canEdit: boolean;
   canShare: boolean;
   canUseViews: boolean;
+  hasTabKeyNavigation: boolean = false;
```
meisekimiu, line 37:
```suggestion
```
This also seems to be an unused leftover.

PR #209, opened 2023-04-25
Upgrade to Angular 15

meisekimiu:
This PR upgrades the web-app's Angular version to Angular 14. ... And then it upgrades it to Angular 15. Initially, this ticket was intended to only upgrade to Angular 14, but upgrading to Angular 14 ended up creating some compatibility issues with Storybook making it impossible to run. After trying to upgrade Storybook, an issue with inconsistent dependency versions would cause Storybook to crash. Ultimately I decided to just try upgrading to Angular 15, hoping that this would bump up some Angular dependencies and allow Storybook to work. I don't believe it fixed the issue, but rather I fixed the issue in another way after nearly completing the Angular 15 upgrade.

Obviously a big question with this PR is: Should we keep the upgrade to 15 since the work has been done, or stick with an upgrade to 14 like the original ticket asks?

<details>
<summary>Should we upgrade to 15 or just 14?</summary>

I think I could go back and fix Storybook on Angular 14 now to make this PR *only* feature the intended Angular 14 upgrade. But I don't want to throw away the work involved in upgrading to Angular 15 since I've already completed it in troubleshooting the upgrade to Angular 14. I would be fine splitting this into two PRs if necessary, though I don't think it's really necessary to wait an arbitrary time between upgrades if the work has already been done and everything is confirmed to work.

Originally when I began attempting to upgrade Angular versions in #112, we decided that Angular upgrades should be done one major version at a time rather than skipping versions. While this PR does include two major version upgrades, the upgrade process for both are properly captured in the git history.
</details>

Anyway, a side effect of the migration work done in this PR is that the Storybook eslint plugin is properly installed, meaning some of the code formatting used in Storybook stories should no longer cause linting errors. In addition, compodoc is no longer installed as a dependency for use with Storybook.

Since Angular 14+ require node 16 or higher, <span title="For the first time, this lockfile version bump has been done intentionally. btw I use Arch">this PR will bump up the lockfile format to version 2.</span> Keep in mind that this may cause some issues when moving between branches.

**Steps To Test**
1. Run `npm install`
    * Make sure you have at least node 16 installed on your system, as Angular 14+ now require a node version of 16 or higher.
3. Run `npm run build` / `npm run build:local` to verify that a production build can be made without errors.
4. Run unit tests to verify that they can be run without errors.
5. Run `npm run storybook` to verify that Storybook can be run without errors
6. Run `npm run dev` to verify that the development server can be run without errors.
    * Click around in the web-app to make sure nothing's broken!

Resolves #205. Resolves #206. Resolves PER-9125, and PER-9245

**More Dependabot Autoresolves:**
Resolves #212. Resolves #213. Resolves #214.


meisekimiu:
This PR is suddenly failing the formatting check since I've added a new commit. I think #211 may fix this issue? I'm not sure how this happened, since I haven't rebased this branch at all.

liam-lloyd (approved):

meisekimiu:
(I think I won't be merging this until #199 is merged, since that's probably going to create some annoying conflicts in the lockfile at the very least. I think I'd rather deal with the conflicts in this PR then force them onto the zoom one!)

crisnicandrei (approved):

meisekimiu:
Added a quick run of Prettier to fix a formatting error that got merged into main. Created #216 as a separate issue to deal with these formatting issues that are getting through CI somehow.

liam-lloyd (approved):

PR #215, opened 2023-04-27
Display the item type in the sidebar

crisnicandrei:
This PR displays the item type in the sidebar in the read-only section.

If it is folder it sets the type as Folder, otherwise it picks from the newly created object the corresponding value for the type

If i have missed any types please do let me know.

### DIFF
--- a/src/app/file-browser/components/sidebar/sidebar.component.html
+++ b/src/app/file-browser/components/sidebar/sidebar.component.html
@@ -130,6 +130,18 @@
             </div>
           </div>
         </ng-container>
+        <div *ngIf="originalFileExtension" class="sidebar-detail">
+          <label>Original Format</label>
+          <div class="sidebar-detail-content">
+           .{{originalFileExtension}}
+          </div>
+        </div>
+        <div *ngIf="permanentFileExtension" class="sidebar-detail">
+          <label>Permanent Format</label>
+          <div class="sidebar-detail-content">
+           .{{permanentFileExtension}}
+          </div>
+        </div>
       </div>
     </div>
   </ng-container>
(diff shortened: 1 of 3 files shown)


meisekimiu (approved):
Looks good to me! Thanks!

meisekimiu:
Can we share this with the experience team before merging? I think they might have had a specific idea about this ticket that's different from the implementation here.

meisekimiu (changes requested):
This looks really good! However, I noticed something while testing out different file types! If a file is still processing (especially if it gets stuck), it will not have any `FileVOs` listed on the record. In this case we try to access properties and methods on an `undefined` object and this throws an error before the `originalFileExtension` and `permanentFileExtension` properties are changed. In this case, if you click on a file that has been processed, and then click on a file that is still processing, the previous File Extension information is shown which can be pretty confusing.

If you'd like to test this for yourself, I think currently all `.wmv` file processing is broken, since when I tried uploading one it got stuck forever.

crisnicandrei:
@meisekimiu and what should i display in that case? unknown?

cecilia-donnelly:
@crisnicandrei Either "unknown" or blank would work, I think. It seems like the main thing is making sure that the previous file's info is cleared.

meisekimiu:
I think leaving it blank would be best. Since we have the `ngIf` directive in place that only shows these fields if they're defined, it will completely hide them until you click on a file that is processed.

crisnicandrei:
@meisekimiu I have hidden the type if the file is not processed yet

meisekimiu (approved):
This looks good! Can you rebase this on main so that the tests pass? (If they still don't pass after rebasing then please do not merge)

meisekimiu (approved):

PR #219, opened 2023-05-01
PER-9216

liam-lloyd:


src/app/file-browser/components/sharing-dialog/sharing-dialog.component.html
```diff
@@ -387,7 +387,12 @@
           </ng-container>
           <ng-container *ngSwitchDefault>
             This link will expire
-            {{ shareLink.expiresDT | date : "MMM d, y 'at' h:mm:ss a" }}
+            {{
+              shareLink.expiresDT
+                | date
+                  : "MMM d, y 'at'
+            h:mm:ss a"
+            }}
```
cecilia-donnelly, line 395: This is what prettier picked, huh? A bit strange but oh well.

cecilia-donnelly (approved):
Nice find, thank you!

PR #221, opened 2023-05-04
Changed the image resolution for the zooming library

crisnicandrei:
I have changed the image in the library component, making it use the highest resolution available


cecilia-donnelly (changes requested):
I would use the `fileurl` under the `FileVO` since that's the full image. Even the larger thumbnail is likely lower quality than the original, and since we want to be able to zoom in we want the absolutely highest quality.

crisnicandrei:
@meisekimiu @cecilia-donnelly I have found a way to fix the tests

meisekimiu (approved):

crisnicandrei:
@cecilia-donnelly can i merge this?

cecilia-donnelly (approved):
Yes, looks good! Thanks for making this change.

PR #222, opened 2023-05-04
 Implement Directive API

meisekimiu:
This PR implements the methods in the `ApiService`'s `DirectiveRepo` for sending requests to stela's directives endpoints. Previously, we added new methods to the `HttpService` to call V2 endpoints with the POST method only. This PR creates a new `HttpV2Service` that separates the V1 and V2 code from each other and allows for callers to specify the needed HTTP method.

A question with this PR is: What do we want making V2 requests in the web-app to look like? I had a lot of potential ideas for what I wanted the method calls to look like, but I arrived to my current solution by just building up the class bit by bit. If you have suggestions on what the methods should look like instead, feel free to make suggestions! (I'm not particularly satisfied with the `options` parameter to control adding CSRF / Auth Tokens to requests).

Right now there's no UI to test these methods. I actually ended up just sticking the directives API calls in a random dialog's `ngOnInit` method to test things out, if that helps!

### DIFF
--- a/src/app/directive/components/directive-edit/test-utils.ts
+++ b/src/app/directive/components/directive-edit/test-utils.ts
@@ -60,7 +60,7 @@ export class MockDirectiveRepo {
     const testDirectiveId = '39b2a5fa-3508-4030-91b6-21dc6ec7a1ab';
     const newDirective: DirectiveData = {
       directiveId: testDirectiveId,
-      archiveId: directive.archiveId,
+      archiveId: parseInt(directive.archiveId.toString()),
       type: directive.type,
       createdDt: new Date(),
       updatedDt: new Date(),
--- a/src/app/directive/components/directive-edit/test-utils.ts
+++ b/src/app/directive/components/directive-edit/test-utils.ts
@@ -1,5 +1,5 @@
 import { HttpErrorResponse, HttpHeaders } from '@angular/common/http';
-import { Directive, DirectiveCreateRequest } from '@models/directive';
+import { DirectiveData, DirectiveCreateRequest } from '@models/directive';
 import { MockDirectiveRepo as DirectiveGetRepo } from '../directive-display/test-utils';
 
 export const createDirective = (email: string, note: string) => {
@@ -24,8 +24,8 @@ export class MockMessageService {
 }
 
 export class MockDirectiveRepo {
-  public static createdDirective: Directive = null;
-  public static editedDirective: Partial<Directive> = null;
+  public static createdDirective: DirectiveData = null;
+  public static editedDirective: Partial<DirectiveData> = null;
   public static failRequest: boolean = false;
   public static errorDelay: number = 0;
   public static accountExists: boolean = true;
@@ -38,7 +38,9 @@ export class MockDirectiveRepo {
     MockDirectiveRepo.accountExists = true;
   }
 
-  public async create(directive: DirectiveCreateRequest): Promise<Directive> {
+  public async create(
+    directive: DirectiveCreateRequest
+  ): Promise<DirectiveData> {
     if (MockDirectiveRepo.failRequest) {
       await new Promise((resolve) =>
         setTimeout(resolve, MockDirectiveRepo.errorDelay)
@@ -52,11 +54,11 @@ export class MockDirectiveRepo {
         },
         status: 404,
         statusText: 'Steward account not found',
-      }) as any as Directive;
+      }) as any as DirectiveData;
       // Simulate returning an unexpected error from the HttpClient
     }
     const testDirectiveId = '39b2a5fa-3508-4030-91b6-21dc6ec7a1ab';
-    const newDirective: Directive = {
+    const newDirective: DirectiveData = {
       directiveId: testDirectiveId,
       archiveId: directive.archiveId,
       type: directive.type,
@@ -77,7 +79,9 @@ export class MockDirectiveRepo {
     return newDirective;
   }
 
-  public async update(directive: Partial<Directive>): Promise<Directive> {
+  public async update(
+    directive: Partial<DirectiveData>
+  ): Promise<DirectiveData> {
     if (MockDirectiveRepo.failRequest) {
       await new Promise((resolve) =>
         setTimeout(resolve, MockDirectiveRepo.errorDelay)
@@ -91,10 +95,10 @@ export class MockDirectiveRepo {
         },
         status: 404,
         statusText: 'Steward account not found',
-      }) as any as Directive;
+      }) as any as DirectiveData;
       // Simulate returning an unexpected error from the HttpClient
     }
     MockDirectiveRepo.editedDirective = directive;
-    return directive as Directive;
+    return directive as DirectiveData;
   }
 }
--- a/src/app/shared/services/api/directive.repo.ts
+++ b/src/app/shared/services/api/directive.repo.ts
@@ -1,30 +1,46 @@
 import {
   AccountVO,
   ArchiveVO,
-  Directive,
+  DirectiveData,
   DirectiveCreateRequest,
   LegacyContact,
+  Directive,
 } from '@models/index';
 import { BaseRepo } from './base';
+import { getFirst } from '../http-v2/http-v2.service';
 
 export class DirectiveRepo extends BaseRepo {
   public async get(archive: ArchiveVO): Promise<Directive> {
-    console.warn('Directive API is currently unimplemented.');
-    return {} as Directive;
+    return getFirst(
+      this.httpV2.get(
+        `/v2/directive/archive/${archive.archiveId}`,
+        {},
+        Directive
+      )
+    ).toPromise();
   }
 
   public async getLegacyContact(account: AccountVO): Promise<LegacyContact> {
-    console.warn('Directive API is currently unimplemented.');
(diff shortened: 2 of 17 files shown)


src/app/directive/components/directive-edit/test-utils.ts
```diff
@@ -52,13 +54,13 @@ export class MockDirectiveRepo {
...
-      }) as any as Directive;
+      }) as any as DirectiveData;
       // Simulate returning an unexpected error from the HttpClient
     }
     const testDirectiveId = '39b2a5fa-3508-4030-91b6-21dc6ec7a1ab';
-    const newDirective: Directive = {
+    const newDirective: DirectiveData = {
       directiveId: testDirectiveId,
-      archiveId: directive.archiveId,
+      archiveId: parseInt(directive.archiveId.toString()),
```
cecilia-donnelly, line 63: Why do you have to cast it to a string and then an int?

src/app/shared/services/http-v2/http-v2.service.ts
```diff
@@ -0,0 +1,224 @@
...
+import { StorageService } from '../storage/storage.service';
+import { environment } from '@root/environments/environment';
+import { Observable } from 'rxjs';
+import { map } from 'rxjs/operators';
+import { Location } from '@angular/common';
+
+const CSRF_KEY = 'CSRF';
+const AUTH_KEY = 'AUTH_TOKEN';
+
+type HttpMethod = 'post' | 'get' | 'put' | 'delete';
```
cecilia-donnelly, line 12: Hurray for verbs.

src/app/shared/services/http-v2/http-v2.service.ts
```diff
@@ -0,0 +1,224 @@
...
+  csrf?: boolean;
+  authToken?: boolean;
+}
+
+const defaultOptions: RequestOptions = {
+  csrf: false,
+  authToken: true,
+};
+
+export function getFirst<T>(observable: Observable<T[]>): Observable<T> {
```
cecilia-donnelly, line 25: This is very useful and it's a little surprising that we need it - do we return everything as an array?

src/app/shared/services/http-v2/http-v2.service.spec.ts
```diff
@@ -0,0 +1,198 @@
...
+} from '@angular/common/http/testing';
+import { TestBed } from '@angular/core/testing';
+import { getFirst, HttpV2Service } from './http-v2.service';
+import { environment } from '@root/environments/environment';
+import { StorageService } from '../storage/storage.service';
+
+const apiUrl = (endpoint: string) => `${environment.apiUrl}${endpoint}`;
+
+class HealthResponse {
+  public status: 'unavailabe' | 'available';
```
cecilia-donnelly, line 14: typo on `unavailable`

cecilia-donnelly (commented):
This is super exciting, @meisekimiu ! Some questions and comments before I try testing it out.

src/app/directive/components/directive-edit/test-utils.ts
meisekimiu: One issue I ran into was that Stela actually expects the `archiveId` to be a string instead of a number. I'm not sure if that's intended, since in the current Permanent API archiveIds are always numbers.
<details>
<summary>Specific explanation of the cast</summary>
In order to not break the current directive components tests right now, I set the `archiveId` type on the Directive model to be a `number | string` to match the usage Stela expects as well as the usage of `ArchiveVO.archiveId` we use in the web-app right now. I didn't want the behavior of this test to change from updating types, so I manually cast the `archiveId` to a number by using `parseInt`... which always expects a string. Since this `archiveId` could either be a string or number, I have to first cast it to a string to ensure it *is* a string before then casting it back into a number.
</details>

I think it might be best just to update the types and tests to match the types that Stela expects, but I wasn't sure I wanted massive updates to these tests to be part of this PR.

src/app/shared/services/http-v2/http-v2.service.ts
meisekimiu: A tricky thing about designing all of this is that some endpoints will return an array of objects and some may only return a single object. In the "V1" code, we return objects that represent the returned `ResponseVO`. These objects will usually have separate methods to grab the data from the `ResponseVO` as either a single value or an array (something like `getArchiveVO()` vs `getArchiveVOs()`). Since Stela is returning data in a more plain format, there's not reason to package everything up in a "response object", since we don't need to write functionality to parse the response object for the data we actually need. But as mentioned earlier... some endpoints may return a single object of type `T` while others will return an array of type `T[]`.

The question then is: how do we want to deal with this? Ideally I don't want the caller to have to make a different functional call for getting a single object vs an array (having single and array versions of each HTTP method sounds like a lot), and I don't think there's a way to effectively do this through TypeScript's type magic. The closest we could get is having all of the `get`/`post`/`put`/`delete` methods return a type of `Observable<T | T[]>`... which ultimately always makes it the responsibility of the caller to do type guarding to determine whether they received an array of objects or a single object.

One of the goals of this service was to keep the typing of everything a bit more locked down and pleasant, so ultimately I decided that we should just *always* return type `T[]` so that everything's a bit more consistent. In the event that the API does return a single object, it will wrap it up into an array of length one. A consequence of this is that it is now slightly annoying if a method *is* fetching a single object (you always have to grab index 0 of the returned array), so I created this helper method to convert an `Observable<T[]>` to `Observable<T>`. That way, when this Observable is converted to a Promise, the Promise will only ever resolve with a single object instead of the array.

If anyone has any better ideas of how to handle this, please let me know!

(I also wish I could have extended the Observable class so calls could look like `this.httpv2.get(...).getFirst().toPromise(...)` instead of `getFirst(this.httpv2.get(...)).toPromise(...)`... but that would not really work)

crisnicandrei (approved):

src/app/directive/components/directive-edit/test-utils.ts
cecilia-donnelly: Thank you! That makes sense. I remember some conversation about stela expecting ids to be strings. I think the idea is that new stela-specific ids will be uuids, which are strings, but I don't remember exactly why we made `archiveId`s strings since they're always ints at the moment.

In any case, I agree with you that big test changes don't belong in this PR.

src/app/shared/services/http-v2/http-v2.service.ts
cecilia-donnelly: This is fascinating, thanks so much for the full explanation! I find it quite readable as is and it makes sense why you decided to do it this way. I don't think I have a better idea.

src/app/shared/services/api/directive.repo.ts
```diff
@@ -1,30 +1,46 @@
...
-    console.warn('Directive API is currently unimplemented.');
+    console.warn('Legacy Contact API is currently unimplemented.');
     return {} as LegacyContact;
   }
 
   public async create(directive: DirectiveCreateRequest): Promise<Directive> {
-    console.warn('Directive API is currently unimplemented.');
-    return {} as Directive;
+    return getFirst(
+      this.httpV2.post('/v2/directive', directive, Directive)
```
cecilia-donnelly, line 30: I'm assuming this is just because I didn't have everything set up locally as I should, but I wasn't sure how to pass the authorization header to `create` so I got a 401.

cecilia-donnelly (approved):
I like this! I was able to call the new endpoints from an `onInit`, as you suggested, @meisekimiu. Your explanations of your choices make sense. I don't think `options`, which you called out in the PR message, is bad at all, though I do have a question inline about passing the authorization header.

src/app/shared/services/api/directive.repo.ts
meisekimiu: The Authorization Header is automatically handled by the `HttpV2Service`, so it should be passed in automatically if you have logged into the web-app with this PR. This actually brings up a good point about calling the new endpoints: We get the Auth token from login, but what do we do if a user is already logged in from a previous session? They're authenticated for the V1 API, and the V2 API tries to cache the Authentication Token, but what if they haven't actually logged in through the web-app since this new change has come in (or they just deleted the Auth Token in browser dev tools)?

cecilia-donnelly: Ahh that makes sense and is a bit of a problem since most people presumably will still be logged in when we announce legacy planning. Maybe `isLoggedIn` should return the token? Or some other endpoint?

meisekimiu: I think `isLoggedIn` would make the most sense

cecilia-donnelly: So, looking at it more: we don't store the token anywhere on the backend and I don't think we should, since that would essentially be going back to storing passwords in the database. To get a token we have to log in to FusionAuth, which requires the user's password, again obviously. I think the right thing to do would be to force re-login, which gets the front end the token. I know we talked about that recently with the "spin forever" bug and there isn't an obvious way to do that in the web-app at the moment, right?

meisekimiu:
While there are some questions on what we want to do when the user doesn't have an Auth Token and they try to make a V2 request, we've decided to merge this for now and address it once the web-app officially calls code that requires an Auth Token (which will probably be when the Directives Modal is fully assembled).

PR #224, opened 2023-05-12
Fix login with MFA enabled

meisekimiu:
The `toPromise` method on RxJS Observables does not play well when there are multiple pipes on a single observable. Add the AuthToken code to the AccountService's pipe instead of having multiple.


cecilia-donnelly (approved):
Fascinating! Thanks for figuring this out. Works locally.

PR #226, opened 2023-05-15
Public search full PR

cecilia-donnelly:
Combine @crisnicandrei's three PRs into one for easier QA. Also rebase on main to make sure we're picking up all the test fixes and new HTTP calls.


cecilia-donnelly:
QA questions:
- Is there a design for what the "no results" page should look like?
- Do all the icons look okay?
- Are thumbnails appearing correctly? (Pretty sure this is just my local failing to process)
- In workspaces, search starts while the user types. In public it does not. Is it clear enough that you have to hit enter to search?
- What about searching in a mobile browser?
- Any styling issues with different screen widths?

cecilia-donnelly:
The tests are being unpredictable - maybe it's a race condition as in [this stackoverflow question](https://stackoverflow.com/questions/51543241/angular-6-unit-tests-an-error-was-thrown-in-afterall-nreferenceerror-cant-fin)? I only changed text so I'm not sure why they would succeed after my last commit but fail on @crisnicandrei's.

meisekimiu:
I've pushed up some changes to address some code issues I've found so we can get this out faster.

src/app/public-search-results/public-search-results.component.ts
```diff
@@ -1,26 +1,29 @@
-import { query } from '@angular/animations';
```
cecilia-donnelly: I take it this was unused?

src/app/public/components/public-search-results/public-search-results.component.ts
```diff
@@ -36,13 +36,13 @@ export class PublicSearchResultsComponent implements OnInit, OnDestroy {
     //get the query param from the route
     if (this.route.params) {
       this.paramsSubscription = this.route.params.subscribe((params) => {
+        this.query = params.query;
```
cecilia-donnelly, line 43: Nice, thank you!

src/app/public/components/public-search-results/public-search-results.component.html
```diff
@@ -1,37 +1,7 @@
...
-            y2="13.4901"
-            gradientUnits="userSpaceOnUse"
-          >
-            <stop stop-color="#131B4A" />
-            <stop offset="1" stop-color="#364493" />
-          </linearGradient>
-        </defs>
-      </svg>
-      <span class="back-to-archive">Back to archive</span>
-    </div>
```
cecilia-donnelly: Why get rid of this special image/styling? Or is it expressed some other way now?

cecilia-donnelly (commented):
Looks good to me, just that one styling question. Thanks for these final code changes, @meisekimiu !

src/app/public/components/public-search-results/public-search-results.component.html
meisekimiu: It's replaced with the `ng-container` line below it. This code was duplicated in both the "has results" and "no results" templates so I put it in an `ng-template` that can be referenced in both places.

cecilia-donnelly: ah thank you!

meisekimiu (approved):

PR #227, opened 2023-05-15
Create Directive Modal

meisekimiu:
This PR creates a hidden tab on the Archive Settings dialog that finally exposes the Legacy Planning UI in the web-app. To access it, the user just has to open the Archive Settings dialog and add `#legacy-planning` to the end of the URL.

This PR will be adjusted to conform to the API changes in https://github.com/PermanentOrg/stela/pull/30 when finished. <del>Right now it still is based on a version of the API that does not return an object for the steward.</del>

**Open Questions**
* This is the first screen that uses the Stela endpoints that require an Auth Token. If the Auth Token isn't loaded, do we forcibly log out the user?

**Steps To Test**:
1. Open the "Archive Settings" dialog
2. Add `#legacy-planning` to the end of the URL
3. Test out the Legacy Planning UI

### DIFF
--- a/src/app/core/components/archive-settings-dialog/archive-settings-dialog.component.html
+++ b/src/app/core/components/archive-settings-dialog/archive-settings-dialog.component.html
@@ -28,6 +28,14 @@
       >
         Archive Settings
       </div>
+      <div
+        class="dialog-tab"
+        (click)="setTab('legacy-planning')"
+        [class.active]="activeTab === 'legacy-planning'"
+        *ngIf="legacyPlanningEnabled"
+      >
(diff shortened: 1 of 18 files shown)


cecilia-donnelly:
Updated https://github.com/PermanentOrg/stela/pull/30, sorry to be a blocker!

liam-lloyd (commented):
I left one nit-level comment about variable names, and also from testing it appears that updating the steward account doesn't work; when I change the email it still sends the old one in the PUT request

liam-lloyd (approved):

PR #228, opened 2023-05-16
PER-9260: Always show jpg version of image

cecilia-donnelly:
If we have a converted image file, it will be a jpg. If we don't, then the original was a jpg. Choose the full size image to show in the zoom window accordingly.

This will fail if for some reason a record has multiple original files and no converted files. That should never happen but I should be a little more defensive about it.

Looking for feedback on all of it, but in particular whether this is even the right file for this function.

### DIFF
--- a/src/app/shared/components/thumbnail/thumbnail.component.ts
+++ b/src/app/shared/components/thumbnail/thumbnail.component.ts
@@ -79,10 +79,11 @@ export class ThumbnailComponent
       this.item instanceof RecordVO &&
       this.item.FileVOs
     ) {
+      const fullSizeImage = this.chooseFullSizeImage(this.item);
       this.viewer = OpenSeaDragon({
         element: resizableImageElement as HTMLElement,
         prefixUrl: 'assets/openseadragon/images/',
-        tileSources: { type: 'image', url: this.item?.FileVOs[0].fileURL },
+        tileSources: { type: 'image', url: fullSizeImage },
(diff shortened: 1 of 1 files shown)


crisnicandrei (approved):

src/app/shared/components/thumbnail/thumbnail.component.ts
```diff
@@ -199,4 +206,15 @@ export class ThumbnailComponent
...
   }
+
+  chooseFullSizeImage(record: RecordVO) {
+    if (record.FileVOs.length > 1) {
+      const convertedUrl = record.FileVOs.find(
+        (file) => file.format == 'file.format.converted'
+      ).fileURL;
+      return convertedUrl;
+    } else {
+      return record.FileVOs[0].fileURL;
```
meisekimiu, line 217:
```suggestion
      return record.FileVOs[0]?.fileURL;
```
This is pretty minor, but if a file is still processing it won't have any FileVOs at all and this line will throw an error. (I tested this by just uploading a bigger file and immediately clicking into it.) Though if we do add this, I think it might also be useful to return from `ngOnInit` if `chooseFullSizeImage()` returns an undefined value so that OpenSeaDragon doesn't need to be loaded in with a null URL.

cecilia-donnelly: Thanks!

meisekimiu (approved):

PR #229, opened 2023-05-17
Per 9221 allow users to choose a file type to download

crisnicandrei:
This pr servers as the implementation of the feature that allows the users to select the file type when downloading the highlighted item. The dropdown appears in the bar above the table, but also in the info side of the selected item.

The dropdown does not appear if there are multiple files selected or if the selected file is a folder.

It also appears in the file viewer coming from the archive or from the public archive.

There seem to be some issues with the downloadURL for some files (ex: docx, because for each converted value the download url is the same, so this might be a backend issue). and also for images coming from the public archive, the same behaviour happening there aswell

---
  - https://app.asana.com/0/0/1204672029895725

### DIFF
--- a/src/app/core/services/edit/edit.service.ts
+++ b/src/app/core/services/edit/edit.service.ts
@@ -208,7 +208,7 @@ export class EditService {
           if (items.length === 1) {
             const item = items[0];
             if (item instanceof RecordVO) {
-              this.dataService.downloadFile(item, '');
+              this.dataService.downloadFile(item);
             }
           }
           break;
--- a/src/app/core/services/edit/edit.service.ts
+++ b/src/app/core/services/edit/edit.service.ts
@@ -208,7 +208,7 @@ export class EditService {
           if (items.length === 1) {
             const item = items[0];
             if (item instanceof RecordVO) {
-              this.dataService.downloadFile(item);
+              this.dataService.downloadFile(item, '');
             }
           }
           break;
--- a/src/app/file-browser/components/download-button/download-button.component.scss
+++ b/src/app/file-browser/components/download-button/download-button.component.scss
@@ -1,0 +1,50 @@
+@mixin flex {
+  display: flex;
+  align-items: center;
+}
+
+
+.download-container{
+  cursor: pointer;
+  display: flex;
+  justify-content: space-between;
+  border-bottom: 1px solid black;
+  border-radius: 4px;
+  padding-bottom: 6px;
+
+  .download-icon-text{
+    display: flex;
+    align-items: center;
+    @include flex();
+    &>span{
+      margin-left: 8px;
+    }
+
+  }
+  .download-icon{
+    @include flex();
+   
+  }
+}
+
+.download-options{
+  width: 211px;
+  height: fit-content;
+  box-shadow: 0px 2px 2px rgba(0, 0, 0, 0.24), 0px 0px 2px rgba(0, 0, 0, 0.12);
+  padding: 0.5rem 0;
+  margin-left: 3rem;
+  &-file-viewer{
+    position: absolute;
+  }
+  
+  .download-option{
+    cursor: pointer;
+    &:hover{
+      background-color: #EAEAEA;
+    }
+    & > span{
+      font-size: 14px;
+      margin-left: 10px;
+    }
+  }
+}
--- a/src/app/file-browser/components/download-button/download-button.component.scss
+++ b/src/app/file-browser/components/download-button/download-button.component.scss
@@ -27,6 +27,7 @@
   }
 }
 
+/* @format */
 .download-options{
   width: 211px;
   height: fit-content;
--- a/src/app/file-browser/components/download-button/download-button.component.scss
+++ b/src/app/file-browser/components/download-button/download-button.component.scss
@@ -1,3 +1,4 @@
+/* @format */
 @mixin flex {
   display: flex;
   align-items: center;
@@ -26,8 +27,6 @@
    
   }
 }
-
-/* @format */
 .download-options{
   width: 211px;
   height: fit-content;
--- a/src/app/file-browser/components/download-button/download-button.component.scss
+++ b/src/app/file-browser/components/download-button/download-button.component.scss
@@ -4,8 +4,7 @@
   align-items: center;
 }
 
-
-.download-container{
+.download-container {
   cursor: pointer;
   display: flex;
   justify-content: space-between;
@@ -13,36 +12,34 @@
   border-radius: 4px;
   padding-bottom: 6px;
 
-  .download-icon-text{
+  .download-icon-text {
     display: flex;
     align-items: center;
     @include flex();
-    &>span{
+    & > span {
       margin-left: 8px;
     }
-
   }
-  .download-icon{
+  .download-icon {
     @include flex();
-   
   }
 }
-.download-options{
+.download-options {
   width: 211px;
   height: fit-content;
   box-shadow: 0px 2px 2px rgba(0, 0, 0, 0.24), 0px 0px 2px rgba(0, 0, 0, 0.12);
   padding: 0.5rem 0;
   margin-left: 3rem;
-  &-file-viewer{
+  &-file-viewer {
     position: absolute;
   }
-  
-  .download-option{
+
+  .download-option {
     cursor: pointer;
-    &:hover{
-      background-color: #EAEAEA;
+    &:hover {
+      background-color: #eaeaea;
     }
-    & > span{
+    & > span {
       font-size: 14px;
       margin-left: 10px;
     }
--- a/src/app/file-browser/components/download-button/download-button.component.ts
+++ b/src/app/file-browser/components/download-button/download-button.component.ts
@@ -93,18 +93,6 @@ export class DownloadButtonComponent implements OnInit {
   }
 
   bringDropdownIntoView() {
-    if (this.downloadButtonDropdown) {
-      this.downloadButtonDropdown?.nativeElement.scrollIntoView({
-        behavior: 'smooth',
-        block: 'end',
-        inline: 'nearest',
-      });
-    } else {
-      this.downloadButton.nativeElement.parentElement.scrollIntoView({
-        behavior: 'smooth',
-        block: 'end',
-        inline: 'nearest',
-      });
-    }
+      this.downloadButton.nativeElement.scrollIntoView();
   }
 }
--- a/src/app/file-browser/components/download-button/download-button.component.ts
+++ b/src/app/file-browser/components/download-button/download-button.component.ts
@@ -28,11 +28,10 @@ export class DownloadButtonComponent implements OnInit {
   @Input() isFileView = false;
 
   @ViewChild('downloadButton') downloadButton: ElementRef;
+  @ViewChild('downloadButtonDropdown', { static: false })
+  downloadButtonDropdown: ElementRef;
 
-  constructor(
-    private data: DataService,
-    private message: MessageService
-  ) {}
+  constructor(private data: DataService, private message: MessageService) {}
 
   ngOnInit(): void {}
 
@@ -50,6 +49,7 @@ export class DownloadButtonComponent implements OnInit {
     if (this.selectedItem instanceof RecordVO && this.selectedItem.FileVOs) {
       if (!this.displayDownloadDropdown) {
         this.displayDownloadOptions();
+        this.bringDropdownIntoView();
       } else {
         this.displayDownloadDropdown = false;
         this.downloadOptions = [];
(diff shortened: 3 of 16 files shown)


cecilia-donnelly (changes requested):
Thanks so much for putting this together, @crisnicandrei ! This is an initial QA pass - I haven't looked at the code yet. Overall you've matched the design beautifully. I have a few requests:

1. First off, rebase on main! This branch is very far behind, which I noticed because the filetypes aren't in the info tab. It's a good practice to always keep your local version of main up to date, or as up to date as possible, so you are branching from an up to date version and have fewer conflicts. We prefer rebasing over merging - it keeps the history simpler. (Though if you're collaborating on a branch with someone I would avoid it as a rule.)
2. 
Screenshot 2023-05-17 at 15-51-08 Private Files Permanent org
The biggest QA issue is shown in this screenshot (I hope), which is that the "Name / Type / Date" column headers should be in a row below the action buttons. Currently they're pushed over to the side.
3. A smaller issue is that the "?" for help overlaps with the download menu when a file is in full view mode: 
Screenshot 2023-05-17 at 15-52-06 Private Files Permanent org
4. You mentioned in the PR message that the download menu does not show up for folders or when multiple files are selected, which is awesome! I do see the downward facing caret when I have one folder selected, as here: 
Screenshot 2023-05-17 at 15-54-37 Private Files Permanent org
I'd prefer it if that didn't appear, since there is no drop down menu.

Thanks again! I'll look into the backend problem with download urls.

cecilia-donnelly:
Oh, and I tested on Firefox!

cecilia-donnelly (changes requested):
Thanks for making these changes! A couple small comments on the download dropdown on the full view screen. I think a little more padding between "Location" and "Download" would look better. Also, if the description is long and the download button shows up at the bottom of the page then it would be better to autoscroll down when the user clicks on it. Currently it looks like nothing happens when the button is clicked because the menu shows up below the end of the page.

Also, if you look at [the list of branches](https://github.com/PermanentOrg/web-app/branches) you can see that this is still 55 commits behind `main`, though I can see that you force-pushed and got some of the `main` changes in here. The order of operations is to make sure your local `main` is up to date, rebase onto it, then force push. Let's meet briefly tomorrow so I can look at it with you.

Over to you, @meisekimiu.

meisekimiu (changes requested):
Could you run formatting on all of this, including the `@format` pragma in the new files?

This looks good! There is one bug I've found that seems to come from the back-end and not the web-app that I think we should address. The back-end is returning an HTTP header to tell the browser that the file is to be downloaded and suggests a default filename. Unfortunately on converted files, it still uses the original filename for the suggested default filename, including the extension. This means if you just click download and a destination without paying attention to the filename, you'd be saving the converted file with the wrong extension. Very confusing... I thought there was a bug in the front end at first!

There exists a `download` attribute on `<a>` tags that I thought might help in suggesting a default filename more accurate to the type, but unfortunately HTTP Headers are prioritized over the value of the `download` attribute, so we will have to fix this on the back end.

I've left some comments on some specific code issues.

src/app/shared/services/data/data.service.ts
```diff
@@ -485,26 +485,31 @@ export class DataService {
...
-        downloadOriginalFile(item);
+        downloadFile(item,fileName);
       });
     }
 
-    function downloadOriginalFile(fileItem: any) {
-      const fileVO = getOriginalFile(fileItem) as any;
+    function downloadFile(fileItem: any,fileName?: string) {
+      let fileVO;
+      fileVO = getFile(fileItem,fileName) as any;
```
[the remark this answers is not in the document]
crisnicandrei: Yes, I did not find any for the FileVO, where could i find all the fields so i can create an interface?

src/app/shared/services/data/data.service.ts
```diff
@@ -485,26 +485,31 @@ export class DataService {
...
     } else {
       return this.fetchFullItems([item]).then(() => {
-        downloadOriginalFile(item);
+        downloadFile(item,fileName);
       });
     }
 
-    function downloadOriginalFile(fileItem: any) {
-      const fileVO = getOriginalFile(fileItem) as any;
+    function downloadFile(fileItem: any,fileName?: string) {
```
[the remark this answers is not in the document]
crisnicandrei: I don't know about why the function was written like that in the first place, should i refactor it?

src/app/core/services/edit/edit.service.ts
```diff
@@ -208,7 +208,7 @@ export class EditService {
           if (items.length === 1) {
             const item = items[0];
             if (item instanceof RecordVO) {
-              this.dataService.downloadFile(item);
+              this.dataService.downloadFile(item, '');
```
cecilia-donnelly, line 211: Why not pass the type here? I'm not totally sure what the edit service is so I'm sure there's a good reason, just trying to follow.

src/app/file-browser/components/download-button/download-button.component.ts
```diff
@@ -0,0 +1,110 @@
...
+      (item) => item.format === 'file.format.converted'
+    ).map((item) => ({
+      name: item.type,
+      extension: item.type.split('.').pop(),
+    }));
+    this.downloadOptions = [
+      { name: original?.type, extension: original?.type.split('.').pop() },
+      ...converted,
+    ];
+  }
```
cecilia-donnelly, line 93: This is a fun way to do it.

src/app/file-browser/components/file-list-controls/file-list-controls.component.html
```diff
@@ -89,3 +97,5 @@
     </ng-container>
   </ng-container>
 </div>
+
+
```
cecilia-donnelly, line 101: Can you get rid of this extra whitespace?

src/app/file-browser/components/file-list-controls/file-list-controls.component.ts
```diff
@@ -339,8 +361,15 @@ export class FileListControlsComponent implements OnInit, OnDestroy, HasSubscrip
...
-      this.data.downloadFile(this.selectedItems[0]);
+    if(this.selectedItems.length === 1 && this.selectedItems[0] instanceof RecordVO){
+      if(!this.displayDownloadDropdown){
+       this.displayDownloadOptions()
+      }
+      else{
+        this.displayDownloadDropdown = false
+        this.downloadOptions = []
+      }
+
```
cecilia-donnelly, line 372: I would think we'd need some extra spaces here between parens and brackets - did you run prettier on this file?

src/app/file-browser/file-browser.module.ts
```diff
@@ -25,7 +25,7 @@ export const LazyLoadFileBrowserSibling = () => import('../file-browser/file-bro
     FileBrowserRoutingModule,
     SharedModule,
     DialogModule,
-    NgbTooltipModule
+    NgbTooltipModule,
```
cecilia-donnelly, line 28: I'd probably remove this since it's the only change to this file in this PR.

cecilia-donnelly (commented):
Thanks so much for this @crisnicandrei ! I'm going to put this up for QA with the corresponding back-end changes. It seems to be working well for me so far.

src/app/file-browser/components/download-button/download-button.component.ts
crisnicandrei: Should I change it?

src/app/core/services/edit/edit.service.ts
crisnicandrei: Good idea, this way i could refactor the download function as well

src/app/file-browser/components/download-button/download-button.component.ts
cecilia-donnelly: Not at all, I was just saying I liked it.

src/app/core/services/edit/edit.service.ts
crisnicandrei: Although isn't the type nested inside the FileVO? Maybe I could just pass "type.format.original" because i think that case only downloads original files. What do you think?

crisnicandrei: Because that's what i was trying to achieve with the empty string :)

cecilia-donnelly: I'm fine with however you want to do it! I was just curious what your reasoning was.

cecilia-donnelly (commented):
It looks to me like you still need to run prettier on src/app/file-browser/components/file-list-controls/file-list-controls.component.ts

crisnicandrei:
@cecilia-donnelly done

k8lyn6 (changes requested):
Hey! This is really exciting!!

Testing on Safari:

- The menu looks like it is transparent, so it's really hard to read because you can see the headers for the file columns behind it at certain browser widths. See photo below:
Screen Shot 2023-06-01 at 10 20 06 AM
- download appears to be broken on the mobile width browser view. Clicking the download button doesn't appear to do anything. I think downloading different file types should be limited to desktop view, and the download button for iPad/mobile width browsers should just return the original file, if possible. 
- When I'm using the download drop down in the Info tab, when it opens the drop down, it appears to scroll back up so you can't immediately see the dropdown. A small issue, but a little confusing for members.
- Some file types that aren't recognizable say that the original format is ".null" but I think it would be an improvement if they said "unknown" instead.

Firefox
- jpeg files return .jpeg files as the Permanent format instead of .jpg. This happened in Safari too.
- the original file .m4v returns a .mp4 (permanent format) for the original format. This happened only on Firefox.

I'm not able to test on Chrome currently (I can't log into Staging for some reason there).

crisnicandrei:
@k8lyn6 the jpeg/jpg issue might be backend related, because I am using the same code for all the files, so I cannot see a way in which some might be correct others wrong on the front-end. What do you think @cecilia-donnelly ?

cecilia-donnelly:
@crisnicandrei yes, I wouldn't worry about the jpeg/jpg problem. It looks like you fixed the other requests that @k8lyn6 made, though I haven't confirmed.

cecilia-donnelly:
Note to self and other reviewers: the backend piece of this has been merged and deployed so this can be tested with only front end changes now (i.e., @k8lyn6 you could deploy this to dev for QA and it should work now).

src/app/shared/services/data/data.service.ts
```diff
@@ -485,24 +485,30 @@ export class DataService {
...
+        downloadFile(item, type);
       });
     }
 
-    function downloadOriginalFile(fileItem: any) {
-      const fileVO = getOriginalFile(fileItem) as any;
+    function downloadFile(fileItem: any, type?: string) {
+      const fileVO = getFile(fileItem, type) as any;
+      console.log('downloadFile', fileVO);
+      console.log('downloadFile', fileItem);
```
cecilia-donnelly, line 501: It would be better to remove these console logs.

cecilia-donnelly (changes requested):
I can see that this request from @k8lyn6 "Some file types that aren't recognizable say that the original format is ".null" but I think it would be an improvement if they said "unknown" instead." still needs to be resolved. I tested by uploading a file that has no extension.

I don't have Safari so couldn't test the transparency problem, but the other issues seem to be resolved.

src/app/shared/services/data/data.service.ts
crisnicandrei: yes you are right, i am sorry about this, i removed them

crisnicandrei: @cecilia-donnelly can you please test now with the file with no extension?

crisnicandrei:
@cecilia-donnelly i need further approvals to merge this, can you please also take a look?

cecilia-donnelly (changes requested):
I see you changed .null to .unknown in the download dropdown menu, thank you! It is still showing up as .null in the "File Info" sidebar. Can you change it there, too?

@meisekimiu, you still have a "Changes Requested" review here - can you re-review to see if @crisnicandrei has made the changes you asked for? Merging is blocked because we both have open change requests.

crisnicandrei:
@cecilia-donnelly i am seeing them as unknown, what type of file have you tested with? :)
Screenshot 2023-05-18 at 15 32 39

cecilia-donnelly:
Ah, I'm referring to the lines above the dropdown - "Original Format" and "Permanent Format" both show as .null in your screenshot.

crisnicandrei:
@cecilia-donnelly yes you are right, sorry, i was looking at the button and missed it right in front of my eyes, will fix asap

cecilia-donnelly:
No problem, I get it!

crisnicandrei:
@cecilia-donnelly done!

crisnicandrei:
@cecilia-donnelly done

cecilia-donnelly (approved):
Thank you!

src/app/file-browser/components/download-button/download-button.component.scss
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+  }
+  .download-icon{
+    @include flex();
+   
+  }
+}
+
+/* @format */
```
meisekimiu, line 30:
```suggestion
/* @format */
@mixin flex {
  display: flex;
  align-items: center;
}


.download-container{
  cursor: pointer;
  display: flex;
  justify-content: space-between;
  border-bottom: 1px solid black;
  border-radius: 4px;
  padding-bottom: 6px;

  .download-icon-text{
    display: flex;
    align-items: center;
    @include flex();
    &>span{
      margin-left: 8px;
    }

  }
  .download-icon{
    @include flex();
   
  }
}

```
Minor nitpick: can you run prettier on this file and move the `@format` pragma to the top?

meisekimiu (approved):
This looks good though I left a minor formatting request related to Prettier.

PR #230, opened 2023-05-18
PER-9258-display formatted text inside the description input

crisnicandrei:
Changed the type of the used textfield in the public archive.

The previous couldn't handle html tags, but the "textarea" type one can

@meisekimiu @cecilia-donnelly Now that i am using the other text area the height of the element is a bit bigger, is that a problem?


src/app/shared/components/inline-value-edit/inline-value-edit.component.html
```diff
@@ -77,8 +77,8 @@
           <div class="placeholder-line"></div>
         </div>
       </div>
-      <div *ngIf="!isEditing" (click)="startEdit()">
-        <p *ngIf="displayValue && displayValue.trim()">{{displayValue}}</p>
+      <div class="description-textarea" [class.can-edit]="canEdit" *ngIf="!isEditing" (click)="startEdit()">
+        <div class="description-textarea-value" *ngIf="displayValue && displayValue.trim()" [innerHtml]="displayValue"></div>
```
meisekimiu, line 81: Is there a reason why the `<p>` tag needed to be changed to a `<div>`? Just wondering because it introduces the same spacing issue (suddenly there's a lot of whitespace on descriptions that would have fit on one line) that was in the previous revision. I think ideally we don't want that whitespace to appear, especially on shares or public galleries where the content isn't editable by the user.

crisnicandrei: @meisekimiu I was thinking the <div> made more sense to hold inner html, but I will change it back to a paragraph tag

meisekimiu (approved):

PR #233, opened 2023-05-22
PER-9272 Added an additional check so other types of documents are displayed

crisnicandrei:


src/app/file-browser/components/file-viewer/file-viewer.component.ts
```diff
@@ -199,7 +199,7 @@ export class FileViewerComponent implements OnInit, OnDestroy {
 
     let url;
 
-    if (original?.type.includes('pdf')) {
+    if (original?.type.includes('pdf') || original?.type.includes('document')) {
```
cecilia-donnelly, line 202: This doesn't seem like the right approach to me since we already check that `this` is a document at the beginning of `getPdfUrl`.

cecilia-donnelly (changes requested):
This does not work for presentations. Let me know what the error you ran into on uploading those was so that we can debug it! On the other hand, it indirectly addresses https://permanent.atlassian.net/browse/PER-8783 by showing the pdf version of uploaded .txt files. I'm not sure that's how we want to manage them, but it could be a nice side effect.

Overall it seems to me that something about how we're checking that a given record is an image is broken, and that's where this change needs to happen.

crisnicandrei:
@cecilia-donnelly I am not sure but i think these are the errors :)
Screenshot 2023-05-24 at 10 26 27
Screenshot 2023-05-24 at 10 27 11

crisnicandrei:
@cecilia-donnelly these are the files that I could upload: pdfs, docx, images, video recordings
I couldn’t upload: ppt, pptx, xlsx 

Maybe I am missing something in my .env?

cecilia-donnelly:
@crisnicandrei and I met last week and my understanding is that the files he was uploading were not real ppt files, etc, because of limitations on his local computer. I pointed him to our folder of test files which he will use going forward. :crossed_fingers:

cecilia-donnelly (approved):
I'm interested in @meisekimiu's thoughts, since I would expect us to be checking whether a record is an image before we decide if it's resizable, which would make this redundant, but maybe not!

This looks like it's all working to me and is ready for QA, @k8lyn6 . (@crisnicandrei , please don't merge until we get the +1 from @k8lyn6 ).

k8lyn6 (changes requested):
It looks like this did fix files that aren't supported and need a thumbnail, but images are no longer zoomable. When I open a TIFF file, I momentarily see the zoom function buttons but then they disappear. They're not clickable when I am able to see them. On a .jpg, I don't see the buttons at all. 

I also think only the first page of converted spreadsheets and presentations is showing. If you open the example presentation in the test files, it only shows the first page of the presentation. I think these used to open in a PDF reader (though I can't totally confirm, since it's currently just broken on prod).

crisnicandrei:
@cecilia-donnelly @meisekimiu can you please confirm the above about the presentations? As far as i know iframes do not support them

meisekimiu:
> @cecilia-donnelly @meisekimiu can you please confirm the above about the presentations? As far as i know iframes do not support them

Documents and presentations are both converted to PDF format to be displayed in the browser. One of the main reasons we do conversion is to have them be displayable in the browser, so for files that have a converted filetype, at least one of the converted version should *always* be displayable in the browser (for presentations it's the one of type `application/pdf`).

k8lyn6 (approved):
This looks good to me!

PR #236, opened 2023-05-24
PER-9267: use show more for long descriptions

crisnicandrei:
For this functionality, I thought that 100 characters for the short text might be enough, but that could change upon request.

When the user clicks the new span with show more, the full text gets displayed, and when the user clicks on see less, the shortened version is being shown.

The short text is a new variable saved inside the component

### DIFF
--- a/src/app/public/components/public-archive/public-archive.component.html
+++ b/src/app/public/components/public-archive/public-archive.component.html
@@ -95,14 +95,14 @@
           <span
             class="profile-description"
             *ngIf="hasSingleValueFor('description', 'textData1') && !showShortText"
-            [innerHTML]="this.profileItems.description[0].textData1"
+            [innerHTML]="this.description"
           ></span>
           <span
           class="profile-description"
           *ngIf="hasSingleValueFor('description', 'textData1') && showShortText"
           [innerHTML]='this.shortText + "..."'
         ></span>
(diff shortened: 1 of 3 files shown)


meisekimiu (commented):
This looks good, although I had to rebase my copy of this branch onto main so that #234 took effect as well. I've added @k8lyn6 onto this PR for QA approval. Can you rebase this onto main so we can deploy it to dev with the previous fixes included as well?

k8lyn6 (changes requested):
Overall, the read more looks good! One major thing: when the "Tell Us About This [archive type]" description that we've implemented the read more on is blank, it currently removes all sections of the public profile (i.e. if there's no description, the public profile shows up as blank even if other sections are filled out). We need to fix this bug before we deploy this change. I think if the "Tell Us About This..." section is blank, there shouldn't be a "read more" or anything.


Another minor thing:
- Ideally, if a description is less than the cutoff length for the read more, the read more shouldn't show up. This could be confusing for members who think they'll get more text when really, that's all there is. (For instance, if the description reads "A short description", there shouldn't be a read more after that because it's short enough to fit.)

k8lyn6 (approved):
Comments were addressed. Looks good!

meisekimiu (approved):

PR #237, opened 2023-05-26
Add Legacy Contact Modal Stories

meisekimiu:
This PR adds the components that will make up the Legacy Contact modal, as well as corresponding Storybook stories to test them with. The final flow that connects the two components together, fully implemented API code, and finally linking the modal in the web app's UI will be finished in future tickets.

Because Storybook was updated in between this ticket and the previous one, the way the stories are written in this ticket differ from previous stories written for the web-app.

Resolves PER-9146 and PER-9147.

**Steps to Test:**
1. Run `npm run storybook`
2. View the Legacy Contact Display / Legacy Contact Edit stories
3. Play around with the various stories.

### DIFF
--- a/src/app/directive/components/directive-edit/directive-edit.component.scss
+++ b/src/app/directive/components/directive-edit/directive-edit.component.scss
@@ -12,9 +12,12 @@
     flex-direction: column;
     small {
       font-style: italic;
-      color: #444;
     }
   }
+}
+
+small {
+  color: #444;
 }
 
 .account-not-found {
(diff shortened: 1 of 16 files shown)


meisekimiu:
By the way, commit 2e9beea is showing up in the Github UI but is just an artifact from rebasing this onto main and is not actually part of the PR.

liam-lloyd (approved):
I left a comment on an accessibility issue related to the link to the knowledge base, but that link doesn't actually exist yet so that can be handled when it's added.

PR #239, opened 2023-05-31
PER-9259-fix pdf display on mobile

crisnicandrei:
I have fixed the pdf display on mobile by adding an additional check and removing a css rule

---
  - https://app.asana.com/0/0/1204725642430373

### DIFF
--- a/src/app/file-browser/components/file-viewer/file-viewer.component.html
+++ b/src/app/file-browser/components/file-viewer/file-viewer.component.html
@@ -33,7 +33,7 @@
           (disableSwipe)="toggleSwipe($event)"
           (isFullScreen)="toggleFullscreen($event)"
           [item]="currentRecord"
-          *ngIf="showThumbnail"
+          *ngIf="showThumbnail && !isDocument"
           [hideResizableImage]="false"
         ></pr-thumbnail>
         <pr-video [item]="currentRecord" *ngIf="isVideo"></pr-video>
--- a/src/app/file-browser/components/file-viewer/file-viewer.component.scss
+++ b/src/app/file-browser/components/file-viewer/file-viewer.component.scss
@@ -113,7 +113,7 @@ iframe {
   iframe {
     width: 100%;
     height: 100%;
-    
+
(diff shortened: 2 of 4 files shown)


crisnicandrei:
I have also assigned @k8lyn6 just now

k8lyn6 (changes requested):
Oops, I just realized this didn't go through code review yet -- sorry for jumping in too fast! However, I did take a look and it looks like PDFs are no longer showing up in the new zoom reader, but it's only showing the first page of the PDFs in a mobile browser. It also doesn't look like a document viewer. It seems to work fine in a desktop browser at different sizes, though (i.e. shows up properly in a document viewer and you can see all pages of the PDF).

crisnicandrei:
@k8lyn6 could you please send me some screenshots of how it used to look before?

k8lyn6:
@crisnicandrei I don't think I have any screenshots available of what it looked like before. To be honest, it could have been broken before we implemented the new zoom feature -- I'm not entirely sure. But ideally, the PDF would open like it does in a desktop browser or even the mobile app so you can view all the pages.

crisnicandrei:
@k8lyn6 

https://github.com/PermanentOrg/web-app/assets/62384997/16a7b824-d97e-429b-bb14-554e7853aeea

This is how it looks like to me, isn't this correct? If not can you please point out what needs to be changed because I couldn't understand. Thanks

cecilia-donnelly (approved):
This looks correct to me. I can scroll the pdf in the mobile-sized browser on Firefox and can still open and zoom an image.

k8lyn6:
@crisnicandrei it looks like it works fine in a web browser collapsed to mobile view, but if you look on an actual mobile browser, this is what happens. Not sure what's different about the two, but I tried on Safari and Chrome on my iPhone after deploying this branch to dev.

https://github.com/PermanentOrg/web-app/assets/16201121/4d5bb81e-31ad-4051-a0ce-42d83e623d67

k8lyn6 (approved):
This looks like it's working as we decided in Slack.

PR #241, opened 2023-06-01
Implement legacy contact API

meisekimiu:
Previously, the Legacy Contact API was stubbed out since it didn't need to be called for Storybook testing. This PR implements methods in the DirectiveRepo of the ApiService that lets us hit the Legacy Contact endpoints of stela.

Before the API was implemented in Stela, it was assumed that we might have to pass in the AccountID when calling these endpoints. Ultimately the real endpoints are able to get the currently authenticated account so this is unnecessary. As a result, we don't need to pass in an AccountVO on any Legacy Contact API methods and so previous calls to the stubbed out API methods that assumed this have been updated to remove AccountVO parameters.

A local helper class, `LegacyContactClass` is used by the DirectiveRepo to construct response objects, although since they implement the interface they are able to be passed around in place of the `LegacyContact` type.

Resolves PER-9162: Implement Legacy Contact API

**How to Test**
Since the Legacy Contact modal is not fully implemented in the UI yet, the only way to test this is to write code to call these endpoints. I ended up sticking this code in the `ngOnInit()` method of the AccountSettingsComponent to test it out, opening up the dialog in the UI and refreshing to verify that the calls were all successful:
```ts
this.api.directive.getLegacyContact().then((contact) => {
      const legacyContact = contact;
      console.log(legacyContact);
      if (contact) {
        legacyContact.name = 'Test Updated User';
        this.api.directive.updateLegacyContact(legacyContact).then((c) => {
          console.log('Updated Legacy Contact: ', c);
        });
      } else {
        this.api.directive
          .createLegacyContact({
            name: 'Test User',
            email: 'natalie+legacycontact@permanent.org',
          })
          .then((c) => {
            console.log('Created Legacy Contact: ', c);
          });
      }
    });
```
I then checked in my browser's dev tools to verify the `console.log` statements. I also manually verified the network requests: On the first page load I get a successful GET call to `/api/v2/legacy-contact`, then a successful POST call to the same endpoint. On the second page load, I get another successful GET call and then a successful PUT call to `api/v2/legacy-contact/{insert-id-here}`.

(Also you can look at the unit tests and compare them to the [API spec](https://github.com/PermanentOrg/stela/blob/main/API.md) if that helps)

### DIFF
--- a/src/app/directive/components/directive-display/directive-display.component.ts
+++ b/src/app/directive/components/directive-display/directive-display.component.ts
@@ -1,5 +1,5 @@
 import { Component, EventEmitter, Input, OnInit, Output } from '@angular/core';
-import { AccountVO, Directive } from '@models/index';
+import { Directive } from '@models/index';
 import { AccountService } from '@shared/services/account/account.service';
 import { ApiService } from '@shared/services/api/api.service';
 
@@ -34,9 +34,7 @@ export class DirectiveDisplayComponent implements OnInit {
 
   protected async getLegacyContact(): Promise<void> {
     try {
-      const legacyContact = await this.api.directive.getLegacyContact(
-        this.account.getAccount()
-      );
+      const legacyContact = await this.api.directive.getLegacyContact();
       if (!legacyContact?.name || !legacyContact?.email) {
         this.noPlan = true;
       }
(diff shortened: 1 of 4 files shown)


meisekimiu:
Note that this is set up to merge into the `legacy-contact-modal` branch. Ideally #237 will be merged first and this will merge into main, but ultimately it doesn't matter.

liam-lloyd (approved):

PR #242, opened 2023-06-05
Add optional STELA_DOMAIN secret

meisekimiu:
On local environments, the domain to access Stela endpoints is the same as the standard API. In dev/staging/prod, Stela will live on a different domain and this domain needs to be configured for these different environments. Allow the Stela domain to be configured with a secret, and then make the HttpV2Service use this domain for Stela requests.

One thing that has affected the design of the HttpV2Service is that there are two different types of V2 endpoints we hit: Those on the legacy backend and those on Stela. To avoid messing up any calls to the legacy backend, the HttpV2Service tries to detect that we're hitting Stela with the presence of the `/v2/` in the endpoint path, *and* I've also allowed callers to manually switch off using the Stela domain value just in case.

While this works, it did make me pause and think about how we really want to structure this. I also came up with an alternate idea where Stela paths could be specified with some kind of marker (e.g. calling the endpoint `stela:/directive/archive/{archiveId}`) which sounds somewhat appealing to me as it would make all the stela endpoint calls potentially more readable as a Stela endpoint vs a legacy backend endpoint. But I could also see how it could obfuscate where Stela endpoints actually live. Feel free to discuss in this PR!

PER-9287: Configure web-app with Stela API Domain

### DIFF
--- a/src/app/shared/services/http-v2/http-v2.service.ts
+++ b/src/app/shared/services/http-v2/http-v2.service.ts
@@ -156,10 +156,19 @@ export class HttpV2Service {
     endpoint: string,
     options: RequestOptions
   ): string {
-    if (options.useStelaDomain && endpoint.match(/^\/*v2\//)) {
+    if (
+      options.useStelaDomain &&
+      endpoint.match(/^\/*v2\//) &&
+      this.isStelaDomainDefined()
+    ) {
       return this.secrets.get('STELA_DOMAIN') ?? this.apiUrl;
     }
     return this.apiUrl;
+  }
+
+  protected isStelaDomainDefined() {
+    const stelaDomain = this.secrets.get('STELA_DOMAIN') ?? '';
+    return stelaDomain.length > 0;
   }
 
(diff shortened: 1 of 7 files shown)


meisekimiu:
Deployed this to dev, and requests are currently blocked by CORS issues!

src/app/shared/services/http-v2/http-v2.service.ts
```diff
@@ -134,8 +142,30 @@ export class HttpV2Service {
...
+      this.isStelaDomainDefined()
+    ) {
+      return this.secrets.get('STELA_DOMAIN') ?? this.apiUrl;
+    }
+    return this.apiUrl;
+  }
+
+  protected isStelaDomainDefined() {
+    const stelaDomain = this.secrets.get('STELA_DOMAIN') ?? '';
+    return stelaDomain.length > 0;
```
crisnicandrei, line 168: why didn't you use double negation ? I am just curious

PR #243, opened 2023-06-07
PER-9235: archive modal sometimes not opening

crisnicandrei:
When accesing the public archive, then navigating to the private page, and then clicking the archive settings modal, there was an error being thrown.

The public module was missing the resolver for the profile items. Instead the profile items were being initialized using the resolver for the public items.

I have added the missing resolver and renamed the profileItems variable to publicProfileItems.

### DIFF
--- a/src/app/core/components/profile-edit/profile-edit.component.ts
+++ b/src/app/core/components/profile-edit/profile-edit.component.ts
@@ -1,3 +1,4 @@
+/* @format */
 import { Component, OnInit, Inject, AfterViewInit } from '@angular/core';
 import { FolderVO, ArchiveVO, RecordVO } from '@models';
 import {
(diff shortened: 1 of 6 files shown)


src/app/core/components/profile-edit/profile-edit.component.ts
```diff
@@ -1,18 +1,36 @@
 import { Component, OnInit, Inject, AfterViewInit } from '@angular/core';
```
meisekimiu, line 1:
```suggestion
/* @format */
import { Component, OnInit, Inject, AfterViewInit } from '@angular/core';
```
Since Prettier was run on this file, can we add the `@format` pragma so this file is checked by Prettier?

meisekimiu (approved):
Thanks for fixing this bug! This seems like it was a bad one that would flood Sentry with errors every frame!

PR #245, opened 2023-06-08
PER-9251: Create Legacy Contact Modal

meisekimiu:
This PR creates a hidden tab on the Account Settings dialog that exposes the Legacy Contact UI in the web-app. To access it, the user just has to open the Account Settings dialog and add `#legacy-contact` to the end of the URL, just like with the Legacy Planning modal before it.

One thing to note about this PR: I took this opportunity to rename the `SettingsDialogComponent` to `AccountSettingsDialogComponent`, so the internal name matches what the whole Permanent team refers to it as. It was originally called `SettingsDialog` because it was the *only* settings dialog before we added a separate Archive Settings dialog in #83. I also ran Prettier on all files affected by this refactoring.

**Stepz To Test:**
1. ~~Make sure you log out and log back in to reset your auth token (if #240 is not merged)~~
2. Open the "Account Settings" dialog
3. Add #legacy-contact to the end of the URL
4. Test out the Legacy Contact UI

### DIFF
--- a/src/app/directive/components/directive-display/directive-display.component.html
+++ b/src/app/directive/components/directive-display/directive-display.component.html
@@ -19,7 +19,10 @@
   </ol>
   <p>
     We have additional support and resources about planning your legacy
-    <a href="#">here.</a>
+    <a
+      href="https://permanent.zohodesk.com/portal/en/kb/articles/legacy-planning"
+      >here.</a
+    >
   </p>
   <div class="archive-steward-header">
     Archive Steward: The {{ archiveName }} Archive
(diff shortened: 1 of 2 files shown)


cecilia-donnelly (approved):
This looks good to me, @meisekimiu ! Thanks for changing the account settings too, that definitely makes things clearer.

cecilia-donnelly:
Oh, actually, the link from "set a directive" to "set a legacy contact" doesn't work yet and I think should be in this PR - if it's going in the "go live" PR then that's also fine!

meisekimiu:
Fixed the link to the Legacy Contact modal and also enabled the "No Plan" check, both in the Directive Display Screen. See the commit message on 33a2f46 for notes on implementing the links.

crisnicandrei (approved):

cecilia-donnelly:
Thanks @meisekimiu ! Updates look good. I'm a bit inclined to merge this and do other followups as separate PRs - what do you think?

meisekimiu:
I think that's fine. I'd like a lot of the final touch ups to be part of the "add the link" (go live) PR if possible

cecilia-donnelly:
Exactly, that's what I'm thinking too.

PR #246, opened 2023-06-12
Save auth token when verifying MFA

meisekimiu:
Previously, the Auth Token was only recorded on the `login` endpoint, which fails when MFA is required to log in. Only once a user successfully verifies through the `verify` endpoint do they get an Auth Token. Because the verification step also gives a `trustToken`, we have to manually find the specific instance of SimpleVO or AuthSimpleVO with the key "authToken" which is probably a better way of looking for the token than relying on just finding the first SimpleVO in the data set.

Fortunately it looks like the token is grabbed when creating an account for the first time.

Resolves PER-9284: Store JWT in all authentication flows

### DIFF
--- a/src/app/shared/services/account/account.service.ts
+++ b/src/app/shared/services/account/account.service.ts
@@ -331,7 +331,7 @@ export class AccountService {
             }
             this.skipSessionCheck = true;
 
-            const authToken = response.getSimpleVO()?.value;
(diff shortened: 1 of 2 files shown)


liam-lloyd (approved):

crisnicandrei (approved):

cecilia-donnelly (approved):
Works for me! Thanks @meisekimiu .

PR #248, opened 2023-06-13
Per 9237 gift storage interface

crisnicandrei:
This pr's purpose is the implementation of the gift storage interface.

I have added a new tab to the storage dialog accessed from the dropdown for the gift component. I have created a new component which in which the new form lies where the user must input the email, amount and the message. Upon confirmation a new modal pops up for confirmation.


meisekimiu:
Can you rebase this on main and fix the conflicts?

cecilia-donnelly (changes requested):
This may have been preexisting, but I can now reach https://ng.permanent.org:4200/app/(private//dialog:storage/done) and see an empty screen in the modal. I think we always want something to appear in the modal, likely the "add storage" page by default.

More concerning, I tried gifting 1.2 GB of storage, which is not allowed since we are limiting to whole numbers of gigabytes. I got a 500 error back from the server with the message `SQLSTATE[22P02]: Invalid text representation: 7 ERROR:  invalid input syntax for type bigint: "-1288490188.8"CONTEXT:  unnamed portal parameter $1 = '...'`. but the user-facing message was "Success!" We should also remove the decimal point from the placeholder text and have it just read "0" or similar to make it clear that users need to send only whole numbers of gbs as gifts. [Here is the slack conversation where Robert suggested sticking to whole gbs for gifting](https://permanent-org.slack.com/archives/C01ARGVA849/p1684350429130289). Mind you, the server error should have been a validation issue, not a sql error, so I'm also looking into that.

crisnicandrei:
@cecilia-donnelly I have pushed some changes, could you please recheck? thanks

cecilia-donnelly (changes requested):
Thanks for validating the storage amount is a whole number! However, just disabling the button is very confusing. Can you add a red box around the problematic entry or something like that? (cc @k8lyn6 since this is really a design question - is there language you'd like to use for the error if a person tries to give 1.5 GB and we're only allowing them to gift integer gbs of storage?)

k8lyn6:
@cecilia-donnelly How about: "Storage can only be gifted in 1GB increments."

crisnicandrei:
@cecilia-donnelly @k8lyn6 should there be a message for invalid email addresses as well?

k8lyn6:
@cecilia-donnelly users can gift storage to people who aren't Permanent members, correct? I don't think we normally have a message for invalid email addresses, so I don't think we need one here.

crisnicandrei:
@k8lyn6 i have added a message if the string is not a valid message (e.g andrei instead of andrei@domain.com)

crisnicandrei:
@k8lyn6 @cecilia-donnelly @meisekimiu I have tested this with the backend changes, and it looks good so I would say it can be reviewed

cecilia-donnelly (commented):
A couple things I noticed when trying this out:
1. Storage doesn't update immediately - but I see that's fixed in https://github.com/PermanentOrg/web-app/pull/308, so we'll just have to keep an eye on any conflicts when we merge!
2. If I try to give away more storage than I have, the "Send Gift Storage" button grays out but there's no message. I think it would be helpful to include a message there to alert people to what's going on.

crisnicandrei:
@cecilia-donnelly where do you check to see if the storage was not updated? I am asking just so I know where I should look. In the dialog or the account dropdown?

cecilia-donnelly:
@crisnicandrei I checked both.

crisnicandrei:
@cecilia-donnelly it updates for me!
https://github.com/PermanentOrg/web-app/assets/62384997/ef40c82a-03c4-4661-ad13-105853d8584f

src/app/core/components/confirm-gift-dialog/confirm-gift-dialog.component.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+    },
+  };
+
+  const mockGiftResult = new BehaviorSubject<boolean>(false);
+
+  beforeEach(async () => {
+    const config: TestModuleMetadata = cloneDeep(Testing.BASE_TEST_CONFIG);
+
+    dialogData = {
+      email: 'test@email.com',
```
cecilia-donnelly, line 42: I usually use "example.com" for things like this, so I don't accidentally use a real email address (even though this won't send any emails). You'll see that in several other places in this codebase, too, so it's good for consistency.

src/app/core/components/gift-storage/gift-storage.component.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    instance.giftForm.updateValueAndValidity();
+    fixture.detectChanges();
+
+    const button: HTMLButtonElement = find('.btn-primary').nativeElement;
+
+    expect(button.disabled).toBe(false);
+  });
+
+  it('the button is disabled if the email is not valid', async () => {
```
cecilia-donnelly, line 49: Nitpicky, but it seems like the convention for test names is to use a more active sentence, like "disables the button if the email is not valid."

src/app/core/components/gift-storage/gift-storage.component.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+    spyOn(instance, 'submitStorageGiftForm').and.callThrough();
+
+    instance.submitStorageGiftForm(instance.giftForm.value);
+
+    expect(instance.submitStorageGiftForm).toHaveBeenCalledWith({
+      email: 'test@example.com',
+      amount: 5,
+      message: '',
+    });
+  });
```
cecilia-donnelly, line 81: Maybe add a test of trying to give more storage than the member has available?

src/app/core/components/storage-dialog/storage-dialog.component.ts
```diff
@@ -1,10 +1,17 @@
 import { ActivatedRoute, ParamMap } from '@angular/router';
```
cecilia-donnelly, line 1: It seems like you ran prettier on this file - can you add the pragma at the top so it's automatically checked going forward?

cecilia-donnelly (commented):
Some requests, but overall this looks good to me @crisnicandrei! I defer to @meisekimiu on actual code organization as usual.

src/app/core/components/confirm-gift-dialog/confirm-gift-dialog.component.ts
```diff
@@ -0,0 +1,45 @@
...
+    this.dialogRef.close();
+  }
+
+  public async onConfirmClick() {
+    try {
+      await this.api.billing.giftStorage(this.email, Number(this.amount));
+      this.giftResult.next(true);
+    } catch (e) {
+      this.msg.showError('Something went wrong! Please try again.');
+      this.giftResult.next(false);
```
meisekimiu, line 41: These lines are not covered by tests. We should probably write some tests that cover what happens if the giftStorage call goes wrong.

src/app/core/components/gift-storage/gift-storage.component.ts
```diff
@@ -0,0 +1,148 @@
...
+        const newAccount = new AccountVO({
+          ...this.account,
+          spaceLeft: remainingSpaceInBytes,
+          spaceTotal:
+            this.account.spaceTotal - giftedAmount * this.bytesPerGigabyte,
+        });
+
+        this.accountService.setAccount(newAccount);
+        this.isSuccessful = isSuccessful;
+      }
```
meisekimiu, line 86: This whole function that runs when the Gifting Result is successful is also not covered by tests.  Do we want to test this bit of functionality?

k8lyn6 (changes requested):
Looking good! I already mentioned some things in [this thread on Slack](https://permanent-org.slack.com/archives/C01ARGVA849/p1698164549123139), but a few minor front end things not included in that thread:
- the text says "they must log in to/create an account in order to claim their storage." Can you remove the "/" and add a space between "to" and "create"?
- once you've finished the gifting flow and return to the gifting screen, I think the fields should go back to being blank instead of filled in with the previous info, if possible

crisnicandrei:
@k8lyn6 done!

k8lyn6 (changes requested):
Sorry, one minor change! I've read that last sentence about 20 times and it's a little confusing. Can we change it to:

>Gift any amount of your unused storage to someone else. You can send storage to both current Permanent users and those who do not yet have an account. Those without an account must create an account in order to claim their storage.

Other than that, it looks great! Thanks Andrei!

k8lyn6 (approved):

PR #249, opened 2023-06-14
PER-9280: Show different welcome message for default signups

meisekimiu:
Show a welcome message reminding users who signed up with no special invite code to check their email to redeem their storage.

Note that this solution does have an edge case bug: If a user signs up with a specific promo code but does not complete onboarding and does so later on another device/browser, they will be shown this new message instead of the default message, which may be inaccurate. We've accepted this as a good enough solution for the time being since there's no way to get a user's invite code via the API right now.

Resolves PER-9280.

**Steps to Test:**
1. Sign up as a user with no promo code (or to just test this message you can also navigate to `/app/welcome`)
2. Verify that the welcome message seen after onboarding is correct.
3. Sign up as a user with a promo code. 
    * To do this, add `?inviteCode={INSERT ANYTHING HERE}` to the end of the URL on the `/app/auth/signup` page.
4. Try refreshing during the onboarding process to verify that the invite code persists between refreshes.
5. Verify that the welcome message seen after onboarding is different from the previous one seen.

<details>
<summary>Pro Testing Tip</summary>
You can also quickly simulate signing up with a promo code by running this line of code in the developer console:

```js
sessionStorage.setItem('INVITE_CODE', 'test');
```

And then navigating to `/app/welcome`. To reset this simply run:
```js
sessionStorage.removeItem('INVITE_CODE');
```

This can be useful if you want to swap between both versions of the dialog quickly, although obviously testing with actual sign ups is important too
</details>

### DIFF
--- a/src/app/shared/services/account/account.service.spec.ts
+++ b/src/app/shared/services/account/account.service.spec.ts
@@ -50,6 +50,10 @@ describe('AccountService', () => {
           get: () => {},
           set: () => {},
         },
+        session: {
+          get: () => {},
+          set: () => {},
+        },
       });
   });
 


k8lyn6 (approved):
Looks good!

PR #250, opened 2023-06-15
Per 9242 add payer archive settings UI

crisnicandrei:
PER-9242 - Add payer archive settings UI

I have added a new section to the archive settings tab where a logged user can set themselves as the payer of the archive.

I have created a new component ArchiveStorageComponent which lies in the archive settings. When the user toggles to toggle, a popup appears with a message,
asking the user if they want to set themselves as the payer of the archive. If the user clicks yes, the payer is set to the logged user.

If the user is already the payer, the user can remove themselves as the payer of the archive. Another message appears asking the user if they want to remove
themselves as the payer of the archive. If the user clicks yes, the payer is set to null.


src/app/core/components/confirm-payer-dialog/confirm-payer-dialog.component.html
```diff
@@ -0,0 +1,48 @@
...
+      <li>
+        When an archive does not have a designated payer, all materials uploaded
+        to that archive will count against the uploading user’s personal storage
+        balance.
+      </li>
+      <li>
+        Only a manager or owner of an archive can assign themself as the payer.
+      </li>
+    </ul>
+  </div>
```
cecilia-donnelly, line 41: This has a lot more whitespace than the design does at top and bottom of the window.

src/app/core/components/confirm-payer-dialog/confirm-payer-dialog.component.html
```diff
@@ -0,0 +1,48 @@
...
+      <li>
+        All materials uploaded to this archive will count against your endowed
+        storage balance, regardless of who uploads it.
+      </li>
+      <li>
+        Only a manager or owner of an archive can assign themself as the payer.
+      </li>
+      <li>Only one payer is allowed per archive.</li>
+    </ul>
+  </div>
```
cecilia-donnelly, line 26: Similarly, lots of whitespace here that could be removed.

src/app/core/components/archive-payer/archive-payer.component.html
```diff
@@ -0,0 +1,40 @@
...
+        archiveId: this.archive?.archiveId,
+        isPayerDifferentThanLoggedUser: this.isPayerDifferentThanLoggedUser,
+        handleAccountInfoChange: this.handleAccountInfoChange.bind(this),
+        cancelAccountPayerSet: this.cancelAccountPayerSet.bind(this)
+      }"
+      [isChecked]="!isPayerDifferentThanLoggedUser"
+      [change]="setArchivePayer"
+    ></pr-switcher>
+    <div>
+      <p class="assign-user">Assign myself as the payer</p>
```
cecilia-donnelly, line 36: This almost looks like a link to me -- is it intended to be clickable in the design, @k8lyn6? (I thought maybe I'd be able to click on it to assign myself as a payer.)

src/app/core/components/archive-payer/archive-payer.component.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    fixture = TestBed.createComponent(ArchivePayerComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
+});
```
cecilia-donnelly, line 26: We could use some more tests here, I'd say. Does it accurately show the payer when a payer has been assigned? What happens if it gets back some unexpected data?

src/app/core/components/archive-payer/archive-payer.component.html
cecilia-donnelly: More relevant for code review, @crisnicandrei: this says "Assign myself as the payer" even if the logged in user is already the payer. That's a small bug that is present in the design, too -- I'm not sure how it should look/what it should say, but I think it's confusing as is. What do you think, @k8lyn6 ?

src/app/shared/components/switcher/switcher.component.html
```diff
@@ -0,0 +1,12 @@
...
+  <label class="switch">
+    <input
+      #switch
+      (change)="change(data)"
+      [checked]="isChecked"
+      type="checkbox"
+    />
+    <span class="slider"></span>
+  </label>
+</div>
```
cecilia-donnelly, line 12: I didn't realize we'd use a new component to make a slider like this. Interesting!

cecilia-donnelly (changes requested):
Thanks @crisnicandrei! Looks close, with some styling and test improvements along with the questions for @k8lyn6 that might have some associated changes.

src/app/core/components/archive-payer/archive-payer.component.spec.ts
crisnicandrei: good idea!

src/app/core/components/confirm-payer-dialog/confirm-payer-dialog.component.html
crisnicandrei: @cecilia-donnelly can you check now please?

src/app/core/components/archive-payer/archive-payer.component.html
crisnicandrei: maybe remove the text if you are the payer? Or if you are the payer something like "Unassign yourself"?

crisnicandrei: Screenshot 2023-06-28 at 13 37 25

k8lyn6: I think you're right about both, @cecilia-donnelly.  I talked to Nathalie R. and I think the changes we should make here are:

- the "Assign myself as the payer" should be gray when the toggle is off, and blue when the toggle is on
- the "Assign myself as the payer" is clickable, and clicking it turns on or off the toggle
- change "Assign my as the payer" to "Remove myself as the payer" when the toggle is on

I think Nathalie made all of the above changes in the design if that description is confusing, too.

I haven't QA'ed this yet since it looks like there are some code review changes first. Let me know when those changes are made and I'll QA it!

crisnicandrei: @k8lyn6 shouldn’t the text be blue at all times? For me a gray text might suggest that the switch is disabled. Let me know what you think about this!

k8lyn6: @crisnicandrei ooh good point. Let's just keep it blue the whole time and clickable then. Sorry for designing on the fly!

cecilia-donnelly:
While testing https://github.com/PermanentOrg/back-end/pull/477, I saw the following occur:
1. Create an archive (with no payer initially)
2. Invite another user as a manager of that archive
3. Sign up as the manager account and set myself (the manager) as the payer on the archive
4. Log in as the owner and remove the manager account from the archive
5. Navigate to "archive settings" and see "This archive's payer is" (blank) and "Access:" instead of "This archive does not currently have an assigned payer."

src/app/core/components/archive-payer/archive-payer.component.spec.ts
```diff
@@ -0,0 +1,171 @@
...
+    component.hasPayer = true;
+    component.isPayerDifferentThanLoggedUser = true;
+    component.payer = payer;
+
+    fixture.detectChanges();
+    const div = fixture.nativeElement.querySelector('#is-not-payer');
+    expect(div).toBeTruthy();
+  });
+
+  it('displays the current payer data if the logged account is different than the payer', () => {
```
cecilia-donnelly, line 88: This should have a different name from the previous test. It looks like this one is meant to be the case where the payer is the logged in user?

src/app/core/components/archive-payer/archive-payer.component.spec.ts
cecilia-donnelly: Thanks for adding more!

src/app/core/components/archive-settings-dialog/archive-settings-dialog.component.ts
```diff
@@ -66,6 +68,12 @@ export class ArchiveSettingsDialogComponent implements OnInit {
         });
     }
     this.archive = this.account.getArchive();
+    this.api.archive.getMembers(this.archive).then((response) => {
+      const members = response.getAccountVOs();
+      this.payer = members.find(
+        (member) => member.accountId === this.archive.payerAccountId
+      );
+    });
```
cecilia-donnelly, line 76: What is this part doing? (Just curious.)

cecilia-donnelly (commented):
One small change but otherwise this looks good! Would love another check from @meisekimiu since there's a lot of Angular going on here that I'm not familiar with.

crisnicandrei:
@cecilia-donnelly it seems that the backend does not remove the payer id if the user gets removed, might i suggest checking when removing a user from the archive if that said user is the payer, to remove their id?

src/app/core/components/archive-settings-dialog/archive-settings-dialog.component.ts
crisnicandrei: It finds the payer objects, finding the memeber with the same account id as the payer account id of the archive

cecilia-donnelly:
Hey @crisnicandrei can you rebase this on main, please?

k8lyn6 (approved):
This all seems to be working well!

PR #253, opened 2023-06-21
Per 9262 arrow nav black screen on full screen

crisnicandrei:
have disabled the full screen arrow navigation mainly because of the behaviour of the open sea dragon library.

The library manipulates the dom and after navigating once and then clicking escape the app would be gone, only the div was still present.

Another issue with fullscreen was pdfs.

I have tried navigating back to them when an item was not an image, but the library did not have the router component so that was not possible.


crisnicandrei:
@cecilia-donnelly @meisekimiu I will squash the commits later

PR #255, opened 2023-06-23
Per 9296 using arrow keys to move cursor

crisnicandrei:
I have created a new service which will save the current state of the input (if it is editing or not) and then disable or enable the swiping accordingly

### DIFF
--- a/src/app/shared/components/inline-value-edit/inline-value-edit.component.ts
+++ b/src/app/shared/components/inline-value-edit/inline-value-edit.component.ts
@@ -111,6 +111,13 @@ export class InlineValueEditComponent implements OnInit, OnChanges {
     if (this.class) {
       this.extraClasses = this.class.split(' ');
     }
+
+    this.elementRef.nativeElement.addEventListener(
+      'keydown',
+      (event: KeyboardEvent) => {
+        event.stopPropagation();
+      }
+    );
   }
 
   ngOnChanges(changes: SimpleChanges): void {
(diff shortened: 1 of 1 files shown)


meisekimiu (changes requested):
Instead of creating a brand new service for this functionality, I think we can just have the `inline-value-edit` component stop further propagation of the `KeyboardEvent` to the rest of the page (where it gets picked up by the `file-viewer-component`).

I made a proof of concept of this by adding the following lines to `inline-value-edit.component.ts` at the end of the `ngOnInit()` method (specifically I edited it on a version of `main` rather than this specific branch):
```ts
this.elementRef.nativeElement.addEventListener(
  'keydown',
  (event: KeyboardEvent) => {
    event.stopPropagation();
  }
);
```

What do you think of this solution as an alternative? I think ideally the file-viewer shouldn't have to know whether or not any input fields are being edited, instead it should be oblivious to any keyboard input at all. Do you see any pros/cons between the two methods?

crisnicandrei:
@meisekimiu I think your method is better than mine. It makes more sense to handle it inside the component instead of having a service to do it, because if we have the same problems in the future we would have to use the service again and again.

With your method everything is taken care of inside the component, so I will update the code to use that. Thanks!

crisnicandrei:
@meisekimiu thanks for your suggestion! I have updated the code

k8lyn6 (approved):
This looks good to me!

PR #256, opened 2023-06-27
PER - 9235

crisnicandrei:
I have added another api call for the profile items and a spinner, this way the modal opens each time

Steps to test:If you are logged in:
 Go to any public archive then refresh the page,  then go to your private archive by clicking your round image in the top bar, then click the archive profile. It should open

If you are not logged in: When viewing any public archive, go to sign up, then after being authenticated go to archive profile.


meisekimiu (approved):

k8lyn6 (changes requested):
I think this fixes the issue, but it looks like it causes a few different issues. Now, after following the testing steps, I get the pop up dialog telling me to fill out my profile even though I've already done this and seen the pop up before (it's only supposed to show up once, when you first open the archive profile). Is there any way to prevent that pop up from coming up with this fix? I think it's okay if not, but I was just wondering (picture below).
Screen Shot 2023-07-18 at 3 30 22 PM


I also noticed that my profile now says it's 0% complete, even though I have a lot of information filled out in it. It only appears to say 0% after following the steps you outlined for testing -- if I check it before, it has the right percentage, and if I exit out and then go back to the archive profile, it has the right percentage too. Did something in your changes reset the profile completeness percentage temporarily? I think this is a bug we'll want to resolve before merging these changes.

crisnicandrei:
@k8lyn6 I have fixed the progress, as for the popup, it seems you are missing a cookie and i don't know why, but i will investigate

crisnicandrei:
@k8lyn6 did you by any chance clear your cookies when testing this? because i think that is why the popup showed up, at least that's how it happened for me, afterwards it did not appear again after retesting

k8lyn6 (approved):
@crisnicandrei the pop up must have just been a weird fluke. I'm not seeing in now, and the percentage looks good now so I'll approve this.

PR #259, opened 2023-07-05
PER-9279: pull all public tags

crisnicandrei:
Created a public service resolver that handles the api call to get all the tags for a public archive and then saving them to local storage


src/app/shared/services/api/archive.repo.ts
```diff
@@ -153,6 +153,10 @@ export class ArchiveRepo extends BaseRepo {
 
     return this.http.sendRequestPromise<ArchiveResponse>('/profile_item/delete', data, ArchiveResponse);
   }
+
+  public getArchiveTags(archiveId: string) {
+    return this.httpV2.get<ArchiveResponse>(`v2/archive/${archiveId}/tags/public`).toPromise();
```
meisekimiu, line 158:
```suggestion
    return this.httpV2.get<TagVOData>(`v2/archive/${archiveId}/tags/public`).toPromise();
```
The typing here is inaccurate; according to the [API documentation](https://github.com/PermanentOrg/stela/blob/main/API.md#archives) this endpoint currently returns an array of TagVOs. (To get this fix to work you'll also have to import the `TagVOData` type as well!)

src/app/public/resolves/public-tags-resolve.service.ts
```diff
@@ -0,0 +1,25 @@
...
+  constructor(private api: ApiService) {}
+
+  async resolve(route: ActivatedRouteSnapshot, state: RouterStateSnapshot) {
+    const archiveNbr = route.params.publicArchiveNbr;
+    const response = await this.api.archive.get([
+      new ArchiveVO({ archiveNbr }),
+    ]);
+    const archiveId = response.getArchiveVO().archiveId;
+    const res = await this.api.archive.getArchiveTags(archiveId);
+    localStorage.setItem('tags', JSON.stringify(res));
```
meisekimiu, line 23: Though this is a bit nitpicky, can you use the `StorageService` to access localStorage instead of accessing it directly? Just want to make sure that all calls to localStorage are consistent!

Also while we're on this topic, I know that the ticket says to store them in localStorage, and that's fine, but would sessionStorage make more sense for storing these tags? (I'm actually not sure since we're only storing them right now, but I try to default to sessionStorage unless I need to use localStorage)

src/app/shared/services/api/archive.repo.ts
```diff
@@ -153,6 +153,10 @@ export class ArchiveRepo extends BaseRepo {
 
     return this.http.sendRequestPromise<ArchiveResponse>('/profile_item/delete', data, ArchiveResponse);
   }
+
+  public getArchiveTags(archiveId: string) {
```
meisekimiu, line 157: Since this is specifically for getting public tags, can we change this method name to something like `getPublicArchiveTags` or something similar so the name isn't confusing?

meisekimiu (changes requested):
I tested it and it works! Thanks for this!

I left some comments about some preferred code changes, but overall it works!

src/app/public/resolves/public-tags-resolve.service.ts
crisnicandrei: I think session storage would be fine, mainly because the API call is being made each time a public archive is visited, so they would get overwritten on each visit anyway

src/app/shared/services/api/archive.repo.ts
crisnicandrei: @meisekimiu I have renamed the method according to your suggestion

src/app/shared/services/api/archive.repo.ts
crisnicandrei: @meisekimiu thanks!

meisekimiu (approved):

PR #260, opened 2023-07-06
Use multipart uploads for large files

meisekimiu:
The back end now has support for using multipart uploads for files bigger than 5GB. Add functions to support these new endpoints and methods in the web-app.

Right now this code hits each multipart upload URL sequentially, one at a time. In theory we should be able to hit batches of these endpoints at a time to potentially speed up multipart uploads, though I didn't want to go all in on that if we deemed it wasn't necessary. I did adjust some code so that in theory converting over to a system where we hit multiple upload URL at the same time should be a bit easier.

**Stepz 2 Test:**
1. Upload a small file and verify that it is uploaded through the standard method.
2. Add some storage to your account by using one of [Stripe's test cards](https://stripe.com/docs/testing) and then verify you can upload a file greater than 5GB.
3. Verify that the progress bar moves up properly for the multipart upload.

To make testing easier, you can modify the `multiPartSize` line in `uploader.ts` to be a smaller size. In testing I had it set to 1 byte so *all* uploads were multipart uploads.

### Deployment Notes
To deploy this to other environments, we need to adjust some settings on S3:
- The PUT method must be allowed in the CORS configuration for that environment.
- In the PUT method's configuration, the `"ETag"` needs to be added in the `ExposeHeaders` field.

Permanent developers do not need to do this for their local environment since all of our local environments use the same CORS config.

### DIFF
--- a/src/app/core/services/upload/uploader.ts
+++ b/src/app/core/services/upload/uploader.ts
@@ -21,6 +21,8 @@ const buildForm = (fields: object, file: File) => {
 
 @Injectable()
 export class Uploader {
+  protected readonly tenMB = 10 * 1024 * 1024;
+
   constructor(private api: ApiService, private httpClient: HttpClient) {}
 
   private getUploadData = async (item: UploadItem) => {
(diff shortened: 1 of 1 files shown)


crisnicandrei (approved):

src/app/shared/services/api/record.repo.ts
```diff
@@ -1,14 +1,50 @@
...
+];
+
+class MultipartUploadUrlsList {
+  public urls: string[] = [];
+  public uploadId: string;
+  public key: string;
 
-const MIN_WHITELIST: (keyof RecordVO)[] = ['recordId', 'archiveNbr', 'folder_linkId'];
-const DEFAULT_WHITELIST: (keyof RecordVO)[] = [...MIN_WHITELIST, 'displayName', 'description', 'displayDT'];
+  protected isInstance(obj: unknown): obj is MultipartUploadUrlsList {
```
crisnicandrei, line 30: I am just curious, what is the purpose of this? Obviously it cheks those keys, but why?

crisnicandrei: can there be instances when those keys are other type? Or just a safeguard?

meisekimiu: This is a TypeScript type guard since we're getting an object of `unknown` type.

liam-lloyd (commented):
In testing with a ~6 GB file, I discovered what I think are two timeout problems. One is that our presigned URLs are presently configured to expire after an hour, which is longer than that 6 GB upload takes. I raised this timeout in my local upload-service (and I'll put up a PR for that soon), but then I encountered what I think was another timeout problem. It's hard to say for sure, because I wasn't watching when it happened, but what I think happened is that when the upload had finally completed and the client went to call V2 /registerRecord, it found it's JWT was expired and logged me out. Does that sound plausible?

meisekimiu:
I think having the JWT expire in lower environments is definitely possible. Also while thinking about it a bit more, I think I should also probably add some logic to make sure to retry requests that fail here since I think if you drop internet connection once the entire upload fails.

liam-lloyd (approved):
After setting JWT timeouts in lower environments to match prod, I successfully uploaded a 6 GB file.

meisekimiu:
(Rebasing this on main before merging just to run the linter on it)

PR #263, opened 2023-07-10
Per 9282 check promo code

crisnicandrei:
This pull request adds the api to check for the promo code 

Steps to test: 

Create an account that has a promo code and the displayed message should appear inside the dialog.

Unfortunately i couldn't get stela to work on my system (I think a fusionauth issue, but I think this should work based on my previous conversation with @meisekimiu on friday when the code was tested locally and it work for @meisekimiu)

---
  - https://app.asana.com/0/0/1205075868663314

### DIFF
--- a/src/app/core/components/welcome-dialog/welcome-dialog.component.ts
+++ b/src/app/core/components/welcome-dialog/welcome-dialog.component.ts
@@ -27,7 +27,7 @@ export class WelcomeDialogComponent implements OnInit {
       const res = await this.api.auth.getInviteToken();
       this.isEarlyBird = res.token === 'earlyb1rd';
     } catch (error) {
-      console.log(error);
+      //fail silently
     }
 
     const archive = this.account.getArchive();
--- a/src/app/core/components/welcome-dialog/welcome-dialog.component.ts
+++ b/src/app/core/components/welcome-dialog/welcome-dialog.component.ts
@@ -27,7 +27,7 @@ export class WelcomeDialogComponent implements OnInit {
       const res = await this.api.auth.getInviteToken();
       this.isEarlyBird = res.token === 'earlyb1rd';
     } catch (error) {
-      //fail silently
+      this.isEarlyBird = false;
     }
 
     const archive = this.account.getArchive();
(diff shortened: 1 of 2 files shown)


src/app/core/components/welcome-dialog/welcome-dialog.component.ts
```diff
@@ -17,15 +18,21 @@ export class WelcomeDialogComponent implements OnInit {
...
+    private api: ApiService
   ) {}
 
-  ngOnInit(): void {
+  async ngOnInit(): Promise<void> {
+    try {
+      const res = await this.api.auth.getInviteToken();
+      this.isEarlyBird = res.token === 'earlyb1rd';
+    } catch (error) {
+      console.log(error);
```
meisekimiu, line 30:
```suggestion
```
Can you get rid of this `console.log`? Though that begs the question of what *should* be in that catch block if the API call fails. I do think we should still catch the error (I'm pretty sure throwing it to Sentry will cause the rest of the component to not work properly).

I think in the case of an API failure, we might want to fail silently and set `isEarlyBird` to some default value, since showing the user an error right when they log in for the first time would probably be jarring. But I think it might help if @k8lyn6 weigh in on what should happen if we can't get the user's invite code for whatever reason.

meisekimiu (approved):
This looks good, except for the small bit of feedback I left below. I'll preemptively approve this though since otherwise it works perfectly.

src/app/core/components/welcome-dialog/welcome-dialog.component.ts
k8lyn6: If I'm following correctly, I think what you suggested makes sense @meisekimiu . They should probably just get the default (earlybird) message if for some reason we can't get the invite code.

k8lyn6 (approved):
Looks good!

src/app/core/components/welcome-dialog/welcome-dialog.component.ts
cecilia-donnelly: For this ticket, @crisnicandrei will just set `isEarlyBird` to `false`. We'll discuss the 1GB separately.

PR #265, opened 2023-07-11
PER-9176-full-view-editing

crisnicandrei:
PER-9176: Full view editing

Added a subscription for the tags, because in the file viewer component they were being initialised only on component initialisation. Also used the same subscription in the edit tags component in case we are using a dialog

Steps to test: Select an item from the file viewer then try adding custom metadata and tags

---
  - https://app.asana.com/0/0/1205075868663308


meisekimiu (approved):

k8lyn6 (approved):
This looks good to me!

PR #267, opened 2023-07-12
Add new linting workflow with eslint

meisekimiu:
This PR adds a new workflow run on pull requests for linting. This workflow runs both prettier and eslint, removing prettier from the unit testing workflow so that it's more clear why automated checks are failing. In addition, I added the `no-console` rule to eslint. Initially it was to test linting failures with, but since it's a good rule that we ~~mostly~~ weren't breaking, I decided to leave it in so we can immediately benefit from turning on eslint.

Resolves PER-8554: Add eslint to CI

## Notes on Merging
Once this PR is merged I will add the `lint` workflow to be required to pass in order to merge a PR into `main`.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -53,6 +53,7 @@ module.exports = {
         'import/no-import-module-exports': 'off',
         'no-underscore-dangle': 'off',
         'max-classes-per-file': 'off',
+        'no-console': ['error', { allow: ['error'] }],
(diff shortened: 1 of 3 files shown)


meisekimiu:
Added a test commit to force the linting workflow to fail and it seems to be working:

liam-lloyd (approved):

PR #271, opened 2023-07-21
Make folder creation requests in small batches

meisekimiu:
This PR groups folder creation request into smaller batches rather than potentially try to create hundreds of folders all at once.

In addition, I've made some small tweaks to the upload process and UI to better communicate what is happening so it doesn't look like the upload process has timed out. While the back end can surely time out when creating too many folders, there is an additional UI issue where the upload progress component can disappear while subfolders are still being created.

To accomplish this, I've made it so all folders are created before any files are uploaded (which was causing this UI issue), and I also added a new "Preparing folders..." message to the upload progress component to make it more clear that folders are being created (previously it would just say "Connecting..." which might indicate a connection error was happening when actually folders were just taking a long time to create).

I made it create folders in batches of ten, but this was a pretty arbitrary number. I did an informal performance test to measure if changing the batch size would affect performance (testing with sizes of 3, 5, 10, and 99), and in general it seems like the batch size didn't matter that much as long as it was kept low. A batch size of 99 ended up creating the UI issues described above.

Resolves PER-8547: Make folder creation requests in small batches

**Steps to Test**
1. Prepare a folder full of a lot of subfolders. These can be mostly empty subfolders, just named things like `a`, `b`, `c`, etc.
2. Add files into some of these subfolders at different depths.
3. Drag and drop the parent folder into your Private Files and let them upload.
4. Verify that the upload progress UI says "Preparing folders" and never disappears until the upload process is fully complete.
5. Verify that network requests created folders in batches no more than 10 folders at a time.

---
  - https://app.asana.com/0/0/1205175187345741

### DIFF
--- a/src/app/core/components/upload-progress/upload-progress.component.html
+++ b/src/app/core/components/upload-progress/upload-progress.component.html
@@ -1,42 +1,56 @@
-<div class="upload-wrapper" [ngClass]="{visible: visible, 'fade': useFade}">
-  <div class="upload-progress" [ngStyle]="{transform: getProgressTransform()}" *ngIf="currentItem"></div>
-  <div *ngIf="status === UploadSessionStatus.Start">
-    Connecting...
-  </div>
+<!-- @format -->
+<div class="upload-wrapper" [ngClass]="{ visible: visible, fade: useFade }">
+  <div
+    class="upload-progress"
+    [ngStyle]="{ transform: getProgressTransform() }"
+    *ngIf="currentItem"
+  ></div>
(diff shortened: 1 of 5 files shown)


crisnicandrei (approved):

cecilia-donnelly (approved):
This is very tidy. I like the new "Preparing folders..." message. I didn't realize at first that the folders are created by depth, so to see the batching at work I'd need to make a folder that has more than ten subfolders - saying that out loud in case it helps @k8lyn6 or in case I've misunderstood.

PR #273, opened 2023-07-21
Enable and fix import/no-duplicates linting rule

meisekimiu:
This PR enables the `import/no-duplicates` linting rule. This rule makes sure all imports from the same module are imported together in the same `import` statement. Files are autofixed with eslint and formatted with prettier.


crisnicandrei (approved):

cecilia-donnelly (approved):

meisekimiu:
Looking to get a quick re-review since an extra fix has been added since approval.

liam-lloyd (approved):

PR #274, opened 2023-07-25
PER-9243

crisnicandrei:
Display the payer storage amount in the left bar of the archive

Steps to test: Set a payer for any archive
                        Check the left menu to see the storage meter appear

---
  - https://app.asana.com/0/0/1205218622395220

### DIFF
--- a/src/app/core/components/archive-payer/archive-payer.component.html
+++ b/src/app/core/components/archive-payer/archive-payer.component.html
@@ -23,7 +23,7 @@
       <p>Contact this member for archive storage requests.</p>
     </div>
   </div>
-  <div class="toggler">
+  <div *ngIf="hasAccess; else noAccess" class="toggler">
     <pr-switcher
       [data]="{
         archiveId: this.archive?.archiveId,
--- a/src/app/core/components/archive-payer/archive-payer.component.html
+++ b/src/app/core/components/archive-payer/archive-payer.component.html
@@ -23,7 +23,7 @@
       <p>Contact this member for archive storage requests.</p>
     </div>
   </div>
-  <div *ngIf="hasAccess; else noAccess" class="toggler">
+  <div *ngIf="hasAccess" class="toggler">
     <pr-switcher
       [data]="{
         archiveId: this.archive?.archiveId,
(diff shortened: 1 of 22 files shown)


meisekimiu:
Currently can't test this fully since [the Stela endpoint this depends on was reverted,](https://github.com/PermanentOrg/stela/pull/48) but it looks good for now! Will test it more thoroughly once the back-end changes are fixed.

meisekimiu (approved):

cecilia-donnelly:
I believe the stela changes are re-merged now.

k8lyn6 (changes requested):
This looks good! A very minor thing -- if new files are added or deleted, you have to refresh the page to get an updated storage bar. I don't think that's a dealbreaker for deploying this, but if there was a way to update it after a change, that would be great. CC'ing @meisekimiu to see what she thinks too. If that's not an easy or possible change, I think this is good to go and I can approve it.

Actually, I don't think this is this pull request, but I just noticed that I can see the "Assign myself as the payer" as a curator, which shouldn't happen. Only owners and managers should be able to see that part in Archive Settings.

crisnicandrei:
@k8lyn6 I have managed to update the storage when deleting an item and also hidden the archive payer for non owners or managers. But for now the storage display cannot be updated because of the way files are being uploaded and processed right now. cc @cecilia-donnelly

k8lyn6 (changes requested):
Ahh sorry for the confusion, @crisnicandrei . Non managers/owners should still be able to see the heading for the archive payer, and still see if there is an archive payer, but they shouldn't be able to see the toggle to set themselves as the payer (i.e. if I'm the owner and the archive payer and you're a curator of that archive, you should still be able to see in archive settings that I'm the archive payer, but you shouldn't be able to see the toggle that allows you to set the archive payer). I should've made that more clear, I'm so sorry!

crisnicandrei:
@k8lyn6 thanks! I will change it!

crisnicandrei:
@k8lyn6 can you check it now? I need to run prettier tomorrow but at least you can see if it works ok!

k8lyn6:
@crisnicandrei I'm trying to deploy this and the build is failing. Can you take a look and see if there are any errors that would be causing the build to fail?

cecilia-donnelly:
Specifically [this error](https://github.com/PermanentOrg/web-app/actions/runs/5870256908/job/15916941071#step:5:47). `noAccess` isn't a property of `ArchivePayerComponent`.

k8lyn6 (approved):
I think this is good now! Sorry for the confusion @crisnicandrei . It looks great!

PR #275, opened 2023-07-28
PER-8802

crisnicandrei:
Changed all the family strings to group. This does not change the types "type.archive.family" because this would be problematic with the current back-end and i think it would be easier once the back-end is changed to do that change too

---
  - https://app.asana.com/0/0/1205175187345745


meisekimiu (approved):

k8lyn6 (approved):
I think this looks good!

PR #277, opened 2023-07-31
Install and integrate MixPanel

meisekimiu:
Install the `mixpanel-browser` package and its corresponding type definitions, add an optional secret to take in a MixPanel project token, and identify users to MixPanel on login.

**Open Questions**
Do we want to have any additional configuration at all?

I decided to label users in testing environments with the environment they're on, since I'm not actually sure how MixPanel works and if there are other ways to filter out users from environments. Since I'm just passing `accountId`, there would be potential for overlap between environments otherwise. Would love some insight on this.

**Steps to Test**
1. Set MIXPANEL_TOKEN in your `.env` file to the project token.
2. Load up the web-app and verify that MixPanel events are being set. You should be able to see debug output in the developer console if you are in a non-production environment.


liam-lloyd:
> I was given the project token before working on this ticket, but how do other developers get access to it (preferably without using Slack DMs)?

If it's accessible by logging into MixPanel and it's feasible to give us MixPanel accounts, I think that's the way we should get it. If giving developers MixPanel accounts drives up costs for little benefit, the thing to do is probably keep it in Bitwarden.

src/app/shared/services/account/account.service.ts
```diff
@@ -95,6 +99,16 @@ export class AccountService {
     Sentry.configureScope((scope) => {
       scope.setUser({ id: this.account.accountId });
     });
+
+    // set account data on mixpanel
+    if (this.secrets.get('MIXPANEL_TOKEN')) {
+      if (newAccount.accountId) {
+        const mixpanelIdentifier = environment.analyticsDebug
+          ? `${environment.environment}:${newAccount.accountId}`
```
cecilia-donnelly: Oh, this is clever.

cecilia-donnelly (commented):
I love the separate commits! They make it so easy to see your thought process. Thank you.

Can you add `MIXPANEL_TOKEN` to .env.template?

I can see some Mixpanel calls in the network tab from my local copy, but I don't see them in the Mixpanel console - or maybe I just don't know where to look? Should I be able to see anything there, or would that require more setup?

meisekimiu:
I was able to see my requests in MixPanel although I think it might have taken a while for it to show up in the actual application. I think when I was testing on Wednesday I was really seeing my data from Monday so I think it takes at least a day for the data to show up easily?

cecilia-donnelly:
It looks like I can [see myself as a user](https://mixpanel.com/project/3063361/view/3579337/app/profile#distinct_id=%24device%3A189c21373e157e-01d0bb78af1f178-47380724-1fa400-189c21373e157e)!  So that's cool. I'm +1 on this if you add the new token to .env.template. Thanks, @meisekimiu !

cecilia-donnelly (approved):

PR #280, opened 2023-08-07
Fix environment typing errors

meisekimiu:
I decided to force the `environment` variables to be readonly for safety purposes since they should never be modified by code. Unfortunately this caused some typing errors that caused compilation to fail. Create an interface to hopefully fix these compliation errors.


cecilia-donnelly (approved):
Ah I see. Sorry I didn't catch that on the earlier review! :crossed_fingers:

PR #281, opened 2023-08-09
 Verify user is logged in before using cached login

meisekimiu:
This PR fixes the elusive "login spins forever" bug in the web-app. The bug was caused by the web-app caching the user's account infomation. When a user’s back end session expires, but the cached account info still exists in their browser, the web-app will assume the user is still logged in and try to fetch the user’s data even though they are not logged in, causing endpoints to error out (and so we get the spinning circle forever instead of the app). This can be fixed by verifying the user is logged in before proceeding to use the cached account info, and clearing that info if the user is not logged in.

**Steps to reproduce the bug:**
1. Log into an account
2. Open your browser's dev tools and access Local Storage:
    * To do this in Chromium-based browsers:
        1. Click on the "Application" tab in dev tools
        2. Open the "Local Storage" drop down under "Storage"
        3. Click into the proper domain for the environment you're testing in (dev.permanent, staging.permanet, etc)
    * To do this in Firefox:
        1. Click on the "Storage" tab in dev tools
        2. Open the "Local Storage" drop down
        3. Click into the proper domain for the environment you're testing in (dev.permanent, staging.permanet, etc)
3. Find the key named "account" in the local storage 
4. Rename it to something else temporarily, such as "account_test". You will need to find this key again.
5. Log out of your account in the web-app UI. You should get redirected back to the login page.
6. Go back into Local Storage via dev tools and locate the account key you renamed. Rename it back to "account", all lower case.
7. Refresh the page.

If you are on a branch that includes this bug fix, everything should work as expected and you'll find yourself at the login screen. Otherwise you should be able to see the loading spinner spinning away forever until you clear local storage.

<details>
<summary>Example of account key in devtools</summary>
Account key in devtools
</details>

PER-9233: Login spins forever

### DIFF
--- a/src/app/shared/services/account/account.service.ts
+++ b/src/app/shared/services/account/account.service.ts
@@ -187,16 +187,27 @@ export class AccountService {
   public refreshAccount() {
     return this.api.account
       .get(this.account)
-      .then((response: AccountResponse) => {
+      .then(async (response: AccountResponse) => {
         if (!response.isSuccessful) {
           throw response;
         }
-
-        const newAccount = response.getAccountVO();
-        this.account.update(newAccount);
-        this.storage.local.set(ACCOUNT_KEY, this.account);
+        // Verify that server agrees that the user is logged in
+        try {
+          const loggedIn = await this.checkSession();
+          if (loggedIn) {
+            const newArchive = response.getArchiveVO();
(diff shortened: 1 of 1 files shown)


liam-lloyd (approved):
:tada: :tada: :tada:

PR #282, opened 2023-08-11
Enable and fix import/order linting rule

meisekimiu:
This linting rules enforces a specific convention in the order of import statements. For the web-app, this typically means importing external modules (from dependencies) before importing internal modules (from within our project).

Because of how we have our project set up, I customized the "internal" and "parent" levels to be equivalent to each other, since our "internal" imports are just shorthands for parent imports anyway.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -59,9 +59,20 @@ module.exports = {
           'error',
           { devDependencies: true },
         ],
+        'import/order': [
+          'error',
+          {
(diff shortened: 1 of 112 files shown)


cecilia-donnelly (approved):
I'll confess that I didn't read every line, but this looks good! I'm thrilled to have an enforced order for imports. Thanks, @meisekimiu .

PR #284, opened 2023-08-22
Per 9228 create goal setting flow

crisnicandrei:
This pull request serves as the implementation of the new onboarding flow. I have opted to make one pr with all the changes instead of 3 pull requests, mainly because I would get confused when having to change things in all 3 of them if changes were needed. 

Steps to test: 

Without pending archive invitations: 

1.Create a new account using a valid email for mail chimp
2.Input all the needed informations and tick the corresponding squares
3. Open the network tab before creating the archive
4. After creating the archive, check the network tab, there should be an API call with a "tags" name that sends everything to mail chimp.

With pending archive invitations everything works as before, but the button that allowed you to create an archive instead of accepting the invitation takes you now to the new onboarding screen.


src/app/file-browser/components/file-list/file-list.component.ts
```diff
@@ -124,8 +125,10 @@ export class FileListComponent implements OnInit, AfterViewInit, OnDestroy, HasS
     @Optional() private drag: DragService,
     private renderer: Renderer2,
     public device: DeviceService,
-    private ngZone: NgZone
+    private ngZone: NgZone,
+    private api: ApiService
```
meisekimiu, line 129: Why is this added here? This seems like a remnant from earlier work. I think we can remove it and the ApiService import.

src/app/file-browser/components/file-viewer/file-viewer.component.ts
```diff
@@ -14,20 +14,15 @@ import { Key } from 'ts-key-enum';
...
-  RecordResponse,
-} from '@shared/services/api/index.repo';
 import { DataService } from '@shared/services/data/data.service';
 import { EditService } from '@core/services/edit/edit.service';
 import { DataStatus } from '@models/data-status.enum';
 import { DomSanitizer } from '@angular/platform-browser';
 import { PublicProfileService } from '@public/services/public-profile/public-profile.service';
 import type { KeysOfType } from '@shared/utilities/keysoftype';
-import { Subscription } from 'rxjs';
+import { from, Subscription } from 'rxjs';
```
meisekimiu, line 25: This new import seems unused.

src/app/shared/components/archive-type-select/archive-type-select.component.html
```diff
@@ -0,0 +1,28 @@
...
+<div class="archive-type-select">
+  <div class="archive-image">
+    <img
+      *ngIf="selectedValue"
+      [src]="icons[selectedValue.split('+')[0]]"
+      [alt]="altTexts[selectedValue.split('+')[0]]"
+    />
+  </div>
+  <div class="archive-dropdown">
+    <select
```
meisekimiu, line 11: Can you make this select match the style of other dropdowns in the application? I think it might need the `form-control` class or a related class? This one just looks a bit smaller than the usual input style.

src/app/shared/components/archive-type-select/archive-type-select.component.ts
```diff
@@ -0,0 +1,33 @@
+/* format */
```
meisekimiu, line 1: We're only ever expecting to use this component in the onboarding flow, right? In that case, can we just include it in the `OnboardingModule` instead of the `SharedModule`? Ideally it wouldn't even need to be in `exports`, since the rest of the application doesn't even need to access this component.

Though if we know we want to use this in multiple places, the shared module is fine.

src/app/onboarding/components/create-new-archive/create-new-archive.component.html
```diff
@@ -1,101 +1,123 @@
...
+      </div>
+      <div class="archive-name-container">
+        <div class="archive-name">
+          <h3>What is your new archive called?</h3>
+          <p>You can always edit your archive's name later.</p>
+          <div class="archive-name-input">
+            <span>The</span>
+            <input
+              [(ngModel)]="name"
+              [placeholder]="'archive subject'"
```
meisekimiu, line 31: Even though the design has "archive subject" in all lowercase, I think we should capitalize it here. (Though ultimately, I think this might be up to @k8lyn6 )

src/app/onboarding/components/create-new-archive/create-new-archive.component.spec.ts
```diff
@@ -24,66 +23,124 @@ let shallow: Shallow<CreateNewArchiveComponent>;
...
-    const { find, outputs } = await shallow.render();
-    checkRadio('type.archive.person', find);
-    clickButton('.next-button', find);
-    expect(outputs.progress.emit).toHaveBeenCalledWith(2);
+  // it('should be able to exit out and go back', async () => {
+  //   const { find, outputs } = await shallow.render();
+  //   expect(outputs.back.emit).not.toHaveBeenCalled();
+  //   clickButton('.back-button', find);
+  //   expect(outputs.back.emit).toHaveBeenCalled();
+  // });
```
meisekimiu, line 60: Instead of commenting out tests, let's just delete them.

src/app/onboarding/components/create-new-archive/create-new-archive.component.html
```diff
@@ -1,101 +1,123 @@
...
+        (click)="setScreen('goals')"
+      >
+        Back
+      </button>
+      <button
+        (click)="onSubmit()"
+        class="btn btn-primary next-button create-archive"
+        [disabled]="!selectedReasons.length"
+      >
+        Create my archive.
```
meisekimiu, line 116: Can we remove the "." here? It's not in the design.

src/app/onboarding/components/create-new-archive/create-new-archive.component.scss
```diff
@@ -78,53 +80,151 @@ form {
...
+  padding: 10px;
   display: flex;
-  flex-direction: column;
   align-items: center;
-  border: 1px solid #C8C8C8;
-  box-shadow: 0px 4px 4px rgba(0, 0, 0, 0.25);
-  border-radius: 8px;
-  padding: 20px 11px;
-  background-color: #fff;
-  cursor: pointer;
```
meisekimiu, line 126: Can we preserve the `cursor: pointer` rule so these grid squares look clickable on desktop?

meisekimiu (changes requested):
Thank you for working on this! This looks good, though I have a few changes requested.

I think ideally I would like the `create-new-archive` screens to be split up into multiple components, but it might be better to just bundle everything into one component since we'll be reusing a lot of code between screens anyway, so I won't push on that.

src/app/shared/components/archive-type-select/archive-type-select.component.ts
meisekimiu: I'm still leaning towards having this stay in the OnboardingModule for the time being; personally I want to try to prioritize having stronger boundaries in this codebase and as of right now there's no definite reason as to why the rest of the application that's importing the `SharedModule` should have access to this specific component. I think if we do want to use it later, than moving it back to the `SharedModule` should be pretty trivial.

This could be a question for the experience team though (pinging @k8lyn6 again), if they know we want to reuse this specific design later, then it can probably stay in the `SharedModule`.

k8lyn6: @meisekimiu I don't know for sure if/how we would reuse it in the future. I could see us reusing it, but I think for the time being your reasoning for it staying in OnboardingModule makes sense.

src/app/onboarding/components/create-new-archive/create-new-archive.component.html
k8lyn6: @meisekimiu I think that makes sense.

src/app/core/core.module.ts
```diff
@@ -1,10 +1,7 @@
-/* @format */
```
meisekimiu, line 1: Let's not get rid of the Prettier format pragma.

src/app/dialog/dialog.service.ts
```diff
@@ -1,5 +1,17 @@
 // eslint-disable-next-line max-len
```
meisekimiu, line 1:
```suggestion
/* @format */
```
Since this file is being formatted by Prettier, let's add the Prettier format pragma here.
(Also, we can get rid of that ugly eslint disable comment since... well, the line has been wrapped now and we don't want to ignore eslint!)

src/app/onboarding/components/archive-type-select/archive-type-select.component.html
```diff
@@ -0,0 +1,29 @@
+<!-- format -->
```
meisekimiu, line 1:
```suggestion
<!-- @format -->
```
This is how the format comment should be written.

src/app/onboarding/components/archive-type-select/archive-type-select.component.spec.ts
```diff
@@ -0,0 +1,23 @@
+/* format */
```
meisekimiu, line 1:
```suggestion
/* @format */
```
Another small issue with the format pragma.

src/app/onboarding/components/archive-type-select/archive-type-select.component.ts
```diff
@@ -0,0 +1,33 @@
+/* format */
```
meisekimiu, line 1:
```suggestion
/* @format */
```

src/app/onboarding/components/create-new-archive/create-new-archive.component.html
```diff
@@ -1,101 +1,123 @@
...
-    <p><strong>What type of archive will this be?</strong> You can always change this later.</p>
-
-    <div class="archive-type-selector">
-      <label>
-        <div [ngClass]="getArchiveTypeClasses('type.archive.person')">
-          <img src="assets/svg/onboarding/archive-person.svg" alt="Person Archive" />
-          <div>
-            <strong class="archive-type-name">Person Archive</strong>
-            <span class="archive-type-description">represents an individual person, such as yourself or someone else</span>
+<!-- format -->
```
meisekimiu, line 1:
```suggestion
<!-- @format -->
```

src/app/onboarding/components/archive-type-select/archive-type-select.component.ts
```diff
@@ -0,0 +1,33 @@
...
+import { Component, EventEmitter, Output, Input } from '@angular/core';
+import { archiveOptions, archiveDescriptions } from '@root/app/onboarding/shared/onboarding-screen';
+
+@Component({
+  selector: 'pr-archive-type-select',
+  templateUrl: './archive-type-select.component.html',
+  styleUrls: ['./archive-type-select.component.scss'],
+})
+export class ArchiveTypeSelectComponent {
+  @Input() selectedValue: string = '';
```
meisekimiu, line 11: Since this is always going to deal with Archive Type, this should really return the `ArchiveType` type that can be found in the `archive-vo.ts` models file instead of a string to make it more type safe.

src/app/onboarding/components/create-new-archive/create-new-archive.component.scss
```diff
@@ -1,4 +1,6 @@
+/* format */
```
meisekimiu, line 1:
```suggestion
/* @format */
```

src/app/onboarding/components/create-new-archive/create-new-archive.component.spec.ts
```diff
@@ -1,7 +1,6 @@
+/* format */
```
meisekimiu, line 1:
```suggestion
/* @format */
```

src/app/onboarding/components/create-new-archive/create-new-archive.component.ts
```diff
@@ -1,8 +1,11 @@
+/* format */
```
meisekimiu, line 1:
```suggestion
/* @format */
```

src/app/onboarding/components/onboarding/onboarding.component.html
```diff
@@ -1,8 +1,9 @@
+<!-- format  -->
```
meisekimiu, line 1:
```suggestion
<!-- @format  -->
```

src/app/onboarding/components/onboarding/onboarding.component.scss
```diff
@@ -1,3 +1,4 @@
+/* format */
```
meisekimiu, line 1:
```suggestion
/* @format */
```

src/app/onboarding/components/onboarding/onboarding.component.spec.ts
```diff
@@ -1,8 +1,8 @@
+/* format */
```
meisekimiu, line 1:
```suggestion
/* @format */
```

src/app/onboarding/components/onboarding/onboarding.component.ts
```diff
@@ -1,22 +1,22 @@
+/* format */
```
meisekimiu, line 1:
```suggestion
/* @format */
```

src/app/onboarding/components/welcome-screen/welcome-screen.component.html
```diff
@@ -1,19 +1,20 @@
+<!-- format  -->
```
meisekimiu, line 1:
```suggestion
<!-- @format  -->
```

src/app/onboarding/components/welcome-screen/welcome-screen.component.spec.ts
```diff
@@ -1,3 +1,4 @@
+/* format */
```
meisekimiu, line 1:
```suggestion
/* @format */
```

src/app/onboarding/components/welcome-screen/welcome-screen.component.ts
```diff
@@ -1,24 +1,27 @@
+/* format */
```
meisekimiu, line 1:
```suggestion
/* @format */
```

src/app/onboarding/onboarding.module.ts
```diff
@@ -1,23 +1,45 @@
-import { NgModule } from '@angular/core';
+import {
```
meisekimiu, line 1:
```suggestion
/* @format */
import {
```

Since this file has been run through Prettier, let's add the pragma to check it.

src/app/shared/services/account/account.service.ts
```diff
@@ -1,9 +1,7 @@
-/* @format */
 import { Injectable, EventEmitter } from '@angular/core';
```
meisekimiu, line 1:
```suggestion
/* @format */
import { Injectable, EventEmitter } from '@angular/core';
```

src/app/onboarding/shared/onboarding-screen.ts
```diff
@@ -1,5 +1,126 @@
 export enum OnboardingScreen {
```
meisekimiu, line 1:
```suggestion
/* @format */
export enum OnboardingScreen {
```

src/app/shared/services/api/account.repo.ts
```diff
@@ -1,7 +1,5 @@
-/* @format */
 import {
```
meisekimiu, line 1:
```suggestion
/* @format */
import {
```

src/app/shared/shared.module.ts
```diff
@@ -1,8 +1,6 @@
-/* @format */
 import { NgModule, ComponentFactoryResolver } from '@angular/core';
```
meisekimiu, line 1: Ideally this file should have no changes to it at all, but especially including getting rid of the format pragma for Prettier.

meisekimiu (changes requested):
I added a new pass of comments. Sorry for not catching all of this the first time, and thanks for addressing some of the requested changes, it's looking great!

src/app/core/core.module.ts
crisnicandrei: I got rid of it because prettier was complaining for some reason, and by removing them it worked. Do you know why that happened @meisekimiu ?

src/app/shared/services/account/account.service.ts
crisnicandrei: The same as above @meisekimiu

src/app/shared/shared.module.ts
crisnicandrei: Same as before regarding the pragma @meisekimiu

src/app/core/core.module.ts
meisekimiu: Can you show me the error? These lines are currently what enables Prettier to check the codebase in Github actions so if we remove them then we will not check the files.

src/app/onboarding/components/archive-type-select/archive-type-select.component.ts
crisnicandrei: @meisekimiu this is a concatenation of the archive type and the tag for mail chimp which is being handled in `create-new-archive` component. I needed both values from the dropdown.

crisnicandrei:
@meisekimiu 
you can check the errors here, I run prettier but nothing changes inside the files. Any ideas why?

src/app/onboarding/components/archive-type-select/archive-type-select.component.ts
meisekimiu: Right, I forgot this was being generated from a `<select>` tag. Still, I feel like the type we're emitting from this component shouldn't be just a formatted string, so that the rest of the onboarding flow doesn't have to work around the concatenation format from this component when it's creating the archive and sending mailchimp tags.

crisnicandrei: @meisekimiu I have updated the Output, what do you think?

meisekimiu: That's looks better. Thanks for updating this. I think ideally it'd be nice to constrain the tag type as well, but I can also see how the tag types might change over time, so it's probably fine to leave it as a string for now.

meisekimiu (approved):
Thanks for making the requested changes!

k8lyn6 (changes requested):
I attempted to create an archive, and I'm getting an error for any archive I try to create that uses the archive type picker (see video). It's happening regardless of what I choose in the archive type picker or what I choose in the onboarding questions. If I pick "create my archive for me" it does work. The console error I get if I look in devtools is `PUT https://api.dev.permanent.org/api/v2/account/tags 404`

However, the "create my archive" shouldn't skip the onboarding questions. It should just select a person archive with the person's name and then move on to the questions, and then create the archive.

https://github.com/PermanentOrg/web-app/assets/16201121/df228f6a-5dde-4738-8624-a005491a0652

crisnicandrei:
@k8lyn6 can you post the response from the error?

k8lyn6:
@crisnicandrei is this what you're looking for? 
Screen Shot 2023-08-29 at 12 21 04 PM

crisnicandrei:
@k8lyn6 that seems like a mail chimp error, @meisekimiu did you also get it?

crisnicandrei:
@meisekimiu do you know anything about the mailchimp error?

meisekimiu:
Copying from slack:
> My findings from researching the Mailchimp issue from the onboarding flow: The accounts aren't added to the e-mail list so when we try to add tags to them, we get a 404 since no those accounts don't exist in the list. The back-end only adds accounts to Mailchimp in prod (our dev list never is modified by the back end), so it seems like we'll have to modify the back-end to add to the dev list in other environments, or manually add e-mail addresses to the dev list in order to test the tagging functionality.

In other words, the functionality is a little broken right now (and while this should in theory work in prod, we don't want to deploy there just to test it!). I think it's a bit up for debate as to what next steps should be.

I think at the very least, if account creation is successful but the tags request fails, the onboarding flow should silently continue and redirect the user to their archive as if nothing happened. It's unfortunate if we don't get those mailchimp tags but failed analytics requests should never be a "show-stopping" error.

Also, while we're waiting on requests, we should probably disable the "Create My Archive" button so multiple archives can't be created.

crisnicandrei:
@meisekimiu @k8lyn6 I have updated the code, now the app works even if the tags fail locally. What do you think? I have also added the loading flag to the button to disable it while loading.

k8lyn6:
@meisekimiu does this mean we can't actually test the tagging until it goes on prod? Ideally, I'd like to be able to test that everything is tagged properly. You mentioned I could add the accounts to mailchimp first and that might work? It might be good for us to jump in a call sometime today and talk more about options so I can better understand them.

k8lyn6:
@crisnicandrei another minor thing I've found: "A community I'm part of" choice should be a group archive, not an organization archive, and should have the group icon when selected.

crisnicandrei:
@k8lyn6 thanks for pointing that out! As you can see in the design it was showing in the organization icon, so that's why I did it that way!
Screenshot 2023-08-30 at 20 17 23

k8lyn6:
@crisnicandrei Sorry about that design error! A few other things I found relating to accepting a member invitation (some of this didn't exist in the design and I just noticed now, I'm sorry!):

- there should be a title on the archive invitation page that says Pending Invitations before the paragraph that already exists
- accepting an invitation should still take you to the onboarding questions before it takes you to your archive
- There's currently no way to go back to the archive invitations should you decide you want to accept the invitation after clicking Create a new archive. We should add a back button in the same place that we've put it in other screens and move the Make my archive for me button to the middle, just for invitations (so it will look like the buttons on the bottom of the goal setting screen)

crisnicandrei:
@k8lyn6 should i remove the already existent "pending invitations" text from below the paragraph then?
Screenshot 2023-08-31 at 11 53 32

crisnicandrei:
@k8lyn6 another thing that came to my mind right now with the following scenario: 

-User x has more than 1 pending archives on sign up
-User selects first archive then goes to the goals screen
-User decides to go back because they made a mistake by selecting that archive and wants to choose another one 

The question is: Should we disable the select button next to the archive that was previously selected? This way we can avoid confusion, so the user will not choose the same archive twice. What do you think?

crisnicandrei:
@k8lyn6 also, the onboarding with pending archives can be tested now :)

k8lyn6:
> @k8lyn6 should i remove the already existent "pending invitations" text from below the paragraph then?

Actually, let's call the title "Welcome! Let's join your first archive together."

> The question is: Should we disable the select button next to the archive that was previously selected? This way we can avoid confusion, so the user will not choose the same archive twice. What do you think?

I think so. Would a user be able to change their mind on which archive they selected? If so, I think we should just disable all the select buttons. Having multiple archive invitations is a pretty edge case, so I'm not super worried about someone not being able to change their mind on which archive to join.

crisnicandrei:
@k8lyn6 sorry but I am a bit confused about “disabling all the select buttons”. can you explain that part again pleas?

k8lyn6:
@crisnicandrei sorry about that! I don't know how difficult it would be to change the archive the member joins if they change their mind after selecting one. If you think that's something we can do easily, I think we can proceed as you outlined. If not, I think if a member tries to go back and join a different archive, they should be unable to select and choose a different archive.

k8lyn6 (changes requested):
- A small change -- the tag "why:collaborate" is misspelled as "why:colalborate."
- Selecting "I'm not sure yet" as the type should create a person archive with the archive name The [Member Name] Archive. Right now, if you choose it, it does create a person archive, but we should have the archive subject field get filled in with the member's name so they can proceed. 

I'm currently having some mysterious issues with testing the mailchimp integration, so I'm going to go ahead and submit this for now and I might add some more things I find once I figure out what's going on.

crisnicandrei:
@k8lyn6 and why “other” should not be the same? i am just curious.

k8lyn6:
@crisnicandrei good question! I think if someone chooses other, they probably have an archive subject in mind but we just don't have the type listed for what they're thinking about. They probably have an archive name in mind as well. If someone selects "I'm not sure" then they probably don't know what kind of archive they're going to make and we should make it for them.

crisnicandrei:
@k8lyn6 thanks that makes sense!

crisnicandrei:
@k8lyn6 I found something thatI don't have the answer to: user accepts an invitation to an archive, the type (ex: type:myself) is the type of the accepted invitation right? I mean this does make sense, but most archives do not have a type yet because of the old onboarding.

k8lyn6:
> @k8lyn6 I found something thatI don't have the answer to: user accepts an invitation to an archive, the type (ex: type:myself) is the type of the accepted invitation right? I mean this does make sense, but most archives do not have a type yet because of the old onboarding.

If a member accepts an invitation, we don't need to pass a type: tag to mailchimp.

crisnicandrei:
@k8lyn6 i have updated the header

k8lyn6 (changes requested):
Last few changes, I think! Sorry about all the back and forth -- this was a big one.

- can we add some padding under the buttons on the first onboarding page? On a smaller screen, it cuts off right at the bottom of the button, and looks a little funny. See screenshot.
Screen Shot 2023-09-07 at 2 38 40 PM
- can we unbold "person" in the "Make my archive for me" modal?
- Choosing "no, cancel" in the "Make my archive for me" modal selects "Myself" as the archive type. It should not change the archive type (so if it was blank before, it should be blank now, and if something else was selected, it should stay selected)

crisnicandrei:
@k8lyn6 pushed the changes

k8lyn6 (approved):
I think this is it! Thanks @crisnicandrei , you're awesome!

crisnicandrei:
@k8lyn6 thank you! i will merge this when i get home!

PR #286, opened 2023-08-24
Add sign up MixPanel event

meisekimiu:
Fire a "Sign up" event when a user has signed up for Permanent. This happens before any onboarding is completed, as adding analytics to onboarding will be handled in a separate ticket.

I picked "Sign up" for the event name as it matches the MixPanel events table made in confluence, and there is no other established standard for event names right now. This can be changed easily though.

**Steps To Test:**
1. Have MixPanel enabled by making sure the MixPanel token is added to the `.env` file
2. Sign up for an account with the developer console open. Verify that, while in debug mode, Mixpanel outputs a "Sign up" event after initially creating an account.

### DIFF
--- a/src/app/shared/services/account/account.service.ts
+++ b/src/app/shared/services/account/account.service.ts
@@ -532,6 +532,7 @@ export class AccountService {
             const newAccount = response;
             newAccount.isNew = true;
             this.setAccount(newAccount);
+            this.mixpanel.track('Sign up', { accountId: newAccount.accountId });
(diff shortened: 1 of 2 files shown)


liam-lloyd (approved):

crisnicandrei (approved):

k8lyn6 (approved):
I'm still new to mixpanel, but this looks like it's working! I was able to sign up for a new account and saw the event in mixpanel.

PR #287, opened 2023-08-24
PER-8964

crisnicandrei:
Log out the user when refreshing if keep me logged in is not checked


meisekimiu (approved):

k8lyn6 (changes requested):
Refreshing does log you out when the "Keep me logged in" box isn't checked, but it also logs you out when it is checked. It should keep you logged in if you refresh or close the tab if the box is checked.

crisnicandrei:
@k8lyn6 interesting. I did test it and it worked for me, but maybe you found something that I missed. I will look again! Thanks!

k8lyn6:
@crisnicandrei if it helps, I was using a Chrome browser!

crisnicandrei:
@k8lyn6 it worked for me on chrome 

https://github.com/PermanentOrg/web-app/assets/62384997/a27ff5d2-d97c-4934-be1b-bcc8567f3e46

crisnicandrei:
@k8lyn6 I have tested on safari too, and it works. Could you post a screen recording? Maybe i am missing something that you discovered

meisekimiu:
I'm looking at this again and I have a question about this behavior: Should we really log out the user on a *refresh* if they don't have the box checked? I think the "keep me logged in" functionality should keep the user logged in for a prolonged period of time, but if they don't have it checked, they should be logged out when the browser session ends. Logged in state should persist through a refresh, but it shouldn't persist through closing a tab/window.

crisnicandrei:
@meisekimiu I see your point, but I could not find anything on how to differentiate between refreshing or closing a browser. From what I found they are similar events

meisekimiu:
I'm specifically referring to using `sessionStorage` to store the account/archive information instead of `localStorage` when the checkbox is left unchecked, although I don't want to act on this until it's confirmed by the experience team.

crisnicandrei:
@k8lyn6 what do you think? Also, do you have any documentation on how to setup MFA?

k8lyn6:
@crisnicandrei If it's possible, I'd prefer a user to be able to stay logged in even if the browser is refreshed. I'll ping you on slack for setting up MFA.

crisnicandrei:
@k8lyn6 @meisekimiu it seems that when we use the `verify` endpoint with mfa, the account comes back in the response, but the field is null regardless of `keepLoggedIn` being `true` or `false`. I think in the case of mfa, this is a backend issue too.

See video attached

https://github.com/PermanentOrg/web-app/assets/62384997/31610305-154d-4803-9a57-3bc5384502b3

k8lyn6 (changes requested):
This doesn't appear to be working for accounts with MFA activated. I had "keeped me logged in" checked and then closed out of the browser, and it logged me out. Here's a video of it.

https://github.com/PermanentOrg/web-app/assets/16201121/0ef71c5a-0745-433c-a283-4af5bcff5dfe

meisekimiu:
> it seems that when we use the `verify` endpoint with mfa, the account comes back in the response, but the field is null regardless of `keepLoggedIn` being `true` or `false`. I think in the case of mfa, this is a backend issue too.

I think we shouldn't have to rely on the back-end's response to `keepLoggedIn`. The back-end *may* have erroneous behavior here, but since the front end is the client with the actual checkbox on it (and the one implementing the behavior), it should be able to remember if the checkbox is clicked. In other words, when verifying on MFA, the AccountService should still remember if the checkbox is clicked and override the back-end value of `null`.

meisekimiu (commented):
Left some additional comments on this.

crisnicandrei:
@meisekimiu what do you think about the latest changes? I have implemented that function you said about, hope it looks good! And I have also handled the boolean in the front-end

src/app/shared/services/storage/storage.service.ts
```diff
@@ -11,6 +12,14 @@ export class StorageService {
     this.session = new BaseStorage(window.sessionStorage);
     this.local = new BaseStorage(window.localStorage);
   }
+
+  getStorageByType({ key, condition }: { key?: string; condition?: boolean }) {
+    const isLocal = this.local.get(key) || condition ? true : false;
```
meisekimiu, line 17: I think we can get rid of the ternary statement here:
```suggestion
    const isLocal = this.local.get(key) || condition;
```

Although you might have to do this to ensure type safety:
```ts
const isLocal = !!(this.local.get(key) || condition)
```

Additionally, I'm not sure about this function. I feel like it can potentially be a source of confusion since it's meant to *get* a value, but adding a `condition` value to the parameter object essentially has the effect of *setting* which storage is being read and written to in additional calls. Essentially this function has the ability to do multiple things, but is only communicating one of those things (that it gets storage by type). Ideally the calls to set which storage is being used are different from the calls to get the preferred storage type.

Initially my idea was just to keep the preferred storage type and method on AccountService, since this functionality is only for the "keepLoggedIn" behavior right now. I think it can stay on StorageService. But I'd like to have a method to set the preferred storage location that is separate from the get method.

meisekimiu (commented):
Left a comment below on some code, feel free to discuss.

src/app/shared/services/storage/storage.service.ts
crisnicandrei: @meisekimiu thank you for explaining it again! I will change it tomorrow. Your suggestion is cleaner and easier to grasp.

crisnicandrei: @meisekimiu just to clear my confusion further, I will clear the new function that I have created and add a new one to handle the storage mainly for the AccountService, is that correct?

meisekimiu: You can handle it in AccountService, but if you think it makes more sense in the StorageService then you can keep it in there.

crisnicandrei:
@meisekimiu I have moved it into the account service. I don't see other use out of my mind where we would have to choose between local or session storage so I think it made more sense to move it.

k8lyn6 (approved):
Looks good!

PR #288, opened 2023-08-25
PER-9331: Display alt text

crisnicandrei:
- Added a new field to the sidebar and the file viewer where the user can see and edit the alt text of a file.
- Display the alt text using aria label for images, audio and video files, but also as titles for iframes.

Steps to test: 

1.Upload a file
2.Select it in the file viewer and update the field.


src/app/file-browser/components/file-viewer/file-viewer.component.html
```diff
@@ -38,7 +38,15 @@
         ></pr-thumbnail>
         <pr-video [item]="currentRecord" *ngIf="isVideo"></pr-video>
         <pr-audio [item]="currentRecord" *ngIf="isAudio"></pr-audio>
-        <iframe [src]="documentUrl" *ngIf="isDocument && documentUrl"></iframe>
+        <iframe
+          [title]="
+            currentRecord.altText
+              ? currentRecord.altText
+              : currentRecord.displayName
```
meisekimiu, line 45: This pattern is used multiple times throughout this PR. Maybe it would make sense to extract this out into a Pipe? I'm not sure if it's fully necessary, but I think it'd probably read easier, be testable, and allow us to change the altText algorithm later in one place. Then we could just write:
```
[title]="currentRecord | getAltText"
```
Or something similar.

meisekimiu (changes requested):
Thanks for working on this! I was able to verify that an `aria-label` attribute was appended to an image! I am noticing that the alt text input looks a bit weird in the fullscreen metadata viewer, including in public archives:


Notice the large font size and border that continues to exist even when I'm not editing it.

I think this is more of an experience team question (tagging @k8lyn6 again!), but I personally think it would be more useful to have alt text show up as a multi-line field like the description rather than as a single line field. Ideally we want to be able to accurately transcribe text that is multiple lines without compromises.

src/app/file-browser/components/file-viewer/file-viewer.component.html
crisnicandrei: thanks! that is a good idea!

crisnicandrei:
> Thanks for working on this! I was able to verify that an `aria-label` attribute was appended to an image! I am noticing that the alt text input looks a bit weird in the fullscreen metadata viewer, including in public archives:
> 
>  Notice the large font size and border that continues to exist even when I'm not editing it.
> 
> I think this is more of an experience team question (tagging @k8lyn6 again!), but I personally think it would be more useful to have alt text show up as a multi-line field like the description rather than as a single line field. Ideally we want to be able to accurately transcribe text that is multiple lines without compromises.

@meisekimiu I tried to keep the design like this. To me it looked that this was an empty input that was not highlighted, but maybe you are right! @k8lyn6 what do you think?

k8lyn6:
> I think this is more of an experience team question (tagging @k8lyn6 again!), but I personally think it would be more useful to have alt text show up as a multi-line field like the description rather than as a single line field. Ideally we want to be able to accurately transcribe text that is multiple lines without compromises.

I think this is right.

meisekimiu (changes requested):
For elements with the `aria-label` attribute on them, can you also supply the proper `aria-role` for them as well? Especially in the case of images, they should probably have `aria-role="img"` on them.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/img_role

crisnicandrei:
@meisekimiu @k8lyn6 this has been updated

meisekimiu (commented):
Can you replace the final occurrences of the `currentRecord.altText ? currentRecord.altText : currentRecord.displayName` logic with the Pipe you made? I think that's the only issue I've found. It seems like the ARIA metadata is set correctly, but I'll wait to see if Kaitlyn can properly detect it with plugins to see if it needs other adjustments.

crisnicandrei:
@meisekimiu Sorry for missing those! I have replaced them. Will pay more attention

meisekimiu (approved):

k8lyn6 (commented):
@meisekimiu @crisnicandrei it looks like the screen reader is still not picking up the image itself. It skips right over it from the left arrow to the right arrow.

crisnicandrei:
@k8lyn6 does this happen only for images or documents videos etc?

crisnicandrei:
@k8lyn6 I have tried using the mac os voice over and it detects the alt text for images, what tool are you using to test?

k8lyn6:
@crisnicandrei PDFs - The screen reader sees it, but it's read as "Application - Blocked"
Videos - works for videos

crisnicandrei:
@k8lyn6 thank you! I will install it tomorrow and test with it

crisnicandrei:
@k8lyn6 I did some changes and images should work now.

Regarding the pdfs with Applciation Blocked, I have tested using Silktide and I got the same response, but when testing with Voice Over from macOS and [this one](https://chrome.google.com/webstore/detail/screen-reader/kgejglhpjiefppelpmljglcjbhoiplfn/related) the title was being read correctly. Could you try testing also with the one I sent?

k8lyn6 (approved):
It appears to be working for images, audio, video, and documents using Voice Over by macOS. I tested across firefox, microsoft edge, and google chrome and everything looks good. Great job @crisnicandrei !

PR #291, opened 2023-08-28
PER-9345-mixpanel-events-to-new-onboarding-flow

crisnicandrei:
Track the events in the onboarding flow for all screens

Steps to test: Go through the new onboarding flow, and on each screen check mixpanel to see if an event has fired. (This can be also checked in the browser console)


meisekimiu (changes requested):
I know this at least somewhat based on #286, which passes the `accountId` into the Mixpanel tracking event, but after looking into the reported data a bit more, it seems like we don't need to pass the `accountId` into the event since Mixpanel should already have the authenticated user's information at this point. Instead of importing the `AccountService` into these components I think we can just import the Mixpanel service and provide no extra data on the event.

Although, it might be best to have each onboarding event be the same and send the current screen (or progress) into the event. This might be a question for @k8lyn6 who will be looking at the data though. What do we want this tracking data to look like? Should each screen be a separate event, or the same event with different parameters representing progress?

k8lyn6:
> What do we want this tracking data to look like? Should each screen be a separate event, or the same event with different parameters representing progress?

This is a great question. I'm still figuring out what things in Mixpanel look like. I'm leaning towards each page being an event, because I think it'll be easier to track dropoff and progress that way.

crisnicandrei:
@meisekimiu from @k8lyn6 ‘s answer, I guess my current implementation right?

meisekimiu (approved):

PR #293, opened 2023-08-29
PER-8785: Search results not appearing

crisnicandrei:
I have updated fuse.js to the latest version and also updated the fuse config objects, setting the ignoreLocation property to true. This will consider the patter a match anywhere in the string, as per fuse.js documentation.

Steps to test:

1. Search any term in the search bar considering there are records uploaded that contain that value.
2. The items should appear in the "In this folder" section.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -39,7 +39,7 @@
         "debug": "4.1.1",
         "firebase": "^9.9.3",
         "firebase-functions": "^3.6.0",
-        "fuse.js": "^5.2.3",
+        "fuse.js": "^6.6.2",
         "gsap": "^3.6.0",
         "hammerjs": "^2.0.8",
         "handlebars": "^4.7.7",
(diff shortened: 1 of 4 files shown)


meisekimiu (approved):
This looks good and it seems to fix the bug described in the ticket.

I'm noticing that SearchService doesn't have any tests (other than commented out boilerplate left 3 years ago). Do you think we should write tests for it while we're here? I'd be happy to jump in to write tests in this branch while we're updating this section of the code, but I'll leave that up to you.

crisnicandrei:
@meisekimiu that sounds good! If you have time feel free to do it, otherwise I can continue with the tests on monday!

k8lyn6 (approved):
This seems to have fixed the issue. Looks good to me.

meisekimiu:
I've added some tests and also went on to refactor some of the methods in the SearchService now that they're covered. (Though the `parseSearchTerm` method is still looking a bit big and scary, and it is covered by tests so we could refactor it, but I'm leaving it for the time being since this is a small ticket.)

I've also found some interesting behavior in the SearchService that we might want to investigate: it looks like we can't currently search for tags that contain quotation marks, even though they can be created just fine in the application. For the time being I've covered this behavior in tests but in the future we might want to change that test and the functionality in the production code.

I don't think you can approve my changes @crisnicandrei since you opened the PR, but if it looks good then you can feel free to merge.

PR #294, opened 2023-08-30
Fix scroll bug after clicking share link dialog

meisekimiu:
The current dialog system locks the browser from scrolling while a dialog is active. This caused a bug in the new dialog that pops up on the Share Link page. Normally dialogs are closed out in the app and can reset this scrolling lock, but this specific dialog has routerLinks that navigate away from the page without closing the dialog, thus leaving the scrolling lock active. This prevents users from completing onboarding if the page contents need to be scrolled, since they can't click buttons to progress. Fix this by closing the dialog when either of the links is clicked.

**Steps To Test:**
1. Go to an already existing account and share something (create a share link)
2. Follow that link in an incognito browser and sign up from the link
3. Progress to the choose archive type page of onboarding. On the current main branch, observe that you cannot see the buttons to move forward or backwards and you cannot scroll. On this branch, observe that the scrolling issue is fixed.

### DIFF
--- a/src/app/share-preview/components/create-account-dialog/create-account-dialog.component.html
+++ b/src/app/share-preview/components/create-account-dialog/create-account-dialog.component.html
@@ -21,10 +21,13 @@
         Learn more about Permanent
       </a>
     </p>
-    <a [routerLink]="['/app', 'signup']" class="btn btn-primary"
+    <a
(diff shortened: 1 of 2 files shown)


crisnicandrei (approved):

k8lyn6 (approved):
Looks good to me!

PR #298, opened 2023-09-12
PER-9350 Change storage checkout screen

crisnicandrei:
Updated the storage checkout screen by changing texts to the button and new links.

Steps to test: Go to the storage tab in the account dropdown -> The new text should be visible.

### DIFF
--- a/src/app/core/components/storage-dialog/storage-dialog.component.html
+++ b/src/app/core/components/storage-dialog/storage-dialog.component.html
@@ -1,3 +1,4 @@
+<!-- @format -->
 <div class="dialog-content">
   <div class="header">
     <span>Storage</span>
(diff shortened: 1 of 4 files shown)


meisekimiu (approved):
From a code perspective this all looks good. Thanks for renaming the name of the input to match the new behavior!

I do have some design suggestions, but I'll leave it up to @k8lyn6 as to whether these are needed or not:
- We should probably make the "publish my name" label bold to stand out from the description text below it.
- The link in the "publish my name" description text should probably be a different color since I find it hard to read against the background (probably just make it white to match the rest of the text, but keep the underline so people know it's a link?) 
- I know that Robert (@omnignorant) provided the description text but I'd like to suggest a correction. Instead of "we add your name to our contributors list at permanent.org/donate" I would rework it to "we add your name to our [contributors list.](https://www.permanent.org/donate)" since that flows better and has more descriptive link text.

k8lyn6 (changes requested):
Looking good! A few changes:

- Instead of "Publish my name with my contribution," can we use the text "Let the world know that I am a Permanent contributor!" We should also bold that text so it stands out from the other text.
- The link in the "when you make a donation..." description text should be white and still underlined, and we should make the change that @meisekimiu suggested below:

> I know that Robert provided the description text but I'd like to suggest a correction. Instead of "we add your name to our contributors list at permanent.org/donate" I would rework it to "we add your name to our [contributors list.](https://www.permanent.org/donate)" since that flows better and has more descriptive link text.

crisnicandrei:
Screenshot 2023-09-15 at 16 46 13
@k8lyn6 the stripe image is being pushed down with the new text. Do I have the permission to increase the width of the whole checkout container so it fits nicely?

crisnicandrei:
Screenshot 2023-09-15 at 16 55 21
@k8lyn6 what do you think?

k8lyn6:
@crisnicandrei Did you put the help text underneath because putting a tooltip isn't possible here? I don't like that the form is too long to see fully on the screen now. If it's possible to move the help text to a tooltip instead of putting it underneath, I think we should probably do that. If not, I think this works.

crisnicandrei:
@k8lyn6 as per Robert: "A question mark icon at the end would probably be the most obvious to users, but we could do both? We could also just expand that grey space vertically and have the descriptive text visible at all times under the option. In that case we could try the following text next to the checkbox:"

I have tried using a tooltip but I couldn't add an anchor element inside it, and personally the text was too long for a tooltp.

k8lyn6 (approved):
@crisnicandrei ok, that's fine then. I think this works for now. Thanks!

PR #299, opened 2023-09-15
Per 9348 storage bar does not update after upload or deletion

crisnicandrei:
This fixes the issue when a user uploads or deletes a file and the account storage bar does not update.

Steps to test: 

1.Upload/Delete any file folder.
2. Check the storage bar

@meisekimiu you might wonder why the request here: 
`getFileOrFolderSize(): number {
    let size = 0;
    if (this.item instanceof FolderVO) {
      this.api.folder.get([this.item]).then((response: FolderResponse) => {
        const folder = response.getFolderVO();
        size = folder.FolderSizeVO.allFileSizeDeep;
      });
    } else {
      size = this.item.size;
    }
    return size;
  }`

For some reason the get all function that retrieves all the items does not populate the FolderSizeVO inside a folder, only this request does ( as far as I have found out), so this was the only workaround I could have think of. If you have any other ideas in mind, please let me know!

### DIFF
--- a/src/app/shared/services/account/account.service.ts
+++ b/src/app/shared/services/account/account.service.ts
@@ -621,7 +621,7 @@ export class AccountService {
     return this.storage.local.get(key) || this.storage.session.get(key);
   }
 
-  public updateAccountStorage(amount) {
+  public deductAccountStorage(amount: number) {
     this.account.spaceLeft -= amount;
     this.setAccount(this.account);
   }
--- a/src/app/shared/services/account/account.service.ts
+++ b/src/app/shared/services/account/account.service.ts
@@ -620,4 +620,10 @@ export class AccountService {
   private getStorage(key) {
     return this.storage.local.get(key) || this.storage.session.get(key);
   }
+
+  public updateAccountStorage(amount) {
+    console.log(amount)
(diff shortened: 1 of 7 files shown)


src/app/shared/services/account/account.service.ts
```diff
@@ -607,4 +607,9 @@ export class AccountService {
   private getStorage(key) {
     return this.storage.local.get(key) || this.storage.session.get(key);
   }
+
+  public updateAccountStorage(amount) {
```
meisekimiu, line 611:
```suggestion
  public deductAccountStorage(amount: number) {
```
I think we should probably rename this method to make it more clear what's happening. My first thought was that `updateAccountStorage` would be a simple setter instead of using the `-=` operator. I think if we are using a relative setter it might be best to make it `addAccountStorage` instead, so that it's positive by default and only negative when we're taking things away. It's a bit confusing to pass a negative into the function if we're *adding* storage, but it would also just be clear if we named it `deductAccountStorage` instead.

meisekimiu (approved):
Besides my one code comment, I think this looks good. Thanks for writing tests!

k8lyn6 (approved):
Looks good to me!

PR #300, opened 2023-09-18
PER-8633: Arrow nav

crisnicandrei:
Allow the user to use the arrows to navigate through tags

I have also disabled the navigation with the arrows when the file viewer is open and the tags dialog is showing.

Steps to test:

1.Open the tags in the side bar or the file viewer.
2. Highlight the input with tab or by clicking it.
3. Use the arrows to navigate.

### DIFF
--- a/src/app/file-browser/components/edit-tags/edit-tags.component.spec.ts
+++ b/src/app/file-browser/components/edit-tags/edit-tags.component.spec.ts
@@ -1,7 +1,7 @@
 /* @format */
-import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+import { async } from '@angular/core/testing';
 import { Shallow } from 'shallow-render';
-import { Subject, Observable } from 'rxjs';
+import { Observable } from 'rxjs';
 
 import { ItemVO, TagVOData, RecordVO } from '@models';
 import { ApiService } from '@shared/services/api/api.service';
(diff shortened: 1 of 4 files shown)


src/app/file-browser/components/edit-tags/edit-tags.component.ts
```diff
@@ -269,4 +270,46 @@ export class EditTagsComponent
   close() {
     this.dialogRef.close();
   }
+
+  onArrowNav(event: KeyboardEvent, index: number) {
+    if (event.key === 'ArrowDown') {
```
meisekimiu, line 275: We should probably use `event.stopPropagation()` here, so that we don't even have to edit `file-viewer.component.ts` at all, as the event gets stopped before it propagates upward to parent components. We should do the same for the `setFocusToFirstTagOrButton` method as well.

src/app/file-browser/components/edit-tags/edit-tags.component.spec.ts
```diff
@@ -195,4 +195,56 @@ describe('EditTagsComponent', () => {
...
+
+    spyOn(componentInstance, 'setFocusToCurrentIndex');
+
+    const testIndex = 0;
+    componentInstance.onArrowNav(
+      new KeyboardEvent('keydown', { key: 'ArrowDown' }),
+      testIndex
+    );
+
+    // expect(componentInstance.currentIndex).toBe(testIndex + 1);
```
meisekimiu, line 211:
```suggestion
```
We can probably delete this comment

src/app/file-browser/components/edit-tags/edit-tags.component.spec.ts
meisekimiu, line 1: I'm not sure if it's possible, but might it be more effective to test if the proper elements are focused in DOM rather than spying on function calls? Sometimes spies are the most effective way to do testing, but I feel like in this situation if we can test focus directly (I think through `document.activeElement`... but I could be unsure about that!) it'd be the best because we could refactor the function calls later and still have tests pass. I think testing focus might be tricky though, so if it isn't possible these tests are a good compromise.

meisekimiu (changes requested):
I left a few comments on some minor things in the code I'd like changed. In terms of the behavior though, this works great from the testing I did!

crisnicandrei:
@meisekimiu I have implemented the changes you asked for! And wrote the tests!

meisekimiu (approved):
Thanks for adding these, I left a small note, but this seems good to me now!

k8lyn6 (approved):
This is great and seems to be working as described!

PR #301, opened 2023-09-21
PER-9353: Search with qutation marks

crisnicandrei:
Replaced the regex match that was removing all the qutation marks with a new function which removes only the first and last qutation marks.

@meisekimiu please let me know if this current behaviour is the wanted one. This is what I understood. Thanks!

### DIFF
--- a/src/app/search/services/search.service.spec.ts
+++ b/src/app/search/services/search.service.spec.ts
@@ -81,14 +81,6 @@ describe('SearchService', () => {
       expectSearchToBe(service.parseSearchTerm('Potato'), 'Potato', []);
     });
 
-    it('should remove tag tokens from search', () => {
-      expectSearchToBe(
-        service.parseSearchTerm('tag:"NonExistentTag"'),
-        undefined,
-        []
-      );
-    });
-
     it('should load tag tokens into the tags array', () => {
       tags.setTags([{ name: 'Potato' }]);
       expectSearchToBe(service.parseSearchTerm('tag:"Potato"'), undefined, [
@@ -108,28 +100,6 @@ describe('SearchService', () => {
         'Hello World',
         [{ name: 'Potato' }, { name: 'Potato' }]
       );
-    });
-
-    it('removes quotation marks from tag names', () => {
-      /*
-        This test probably represents erroneous behavior that we want to change
-        at some point. For now, this represents the original functionality of
-        the SearchService as it was written.
-      */
-      tags.setTags([
-        { name: 'Potato', tagId: 0 },
-        { name: '"Potato"', tagId: 1 },
-      ]);
-      const searchTokens = service.parseSearchTerm('tag:""Potato""');
-      expect(searchTokens[1].length).toBe(1);
-      expect(searchTokens[1][0].tagId).toBe(0);
-    });
-
-    it('handles tag edge cases', () => {
-      tags.setTags([{ name: 'tag:Test' }]);
-      expectSearchToBe(service.parseSearchTerm('tag:"tag:"Test""'), undefined, [
-        { name: 'tag:Test' },
-      ]);
     });
 
     function expectSearchToBe(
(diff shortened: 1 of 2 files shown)


meisekimiu (changes requested):
So I have a bit of a weird, probably very rare, edge case that breaks with this specific fix. It's a bit odd, but here's a test you can plug into the `ParseSearchTerm` describe block that will fail unexpectedly:

```ts
it('can handle tags with tags with quotation marks followed by search terms with quotation marks', () => {
  tags.setTags([{ name: '"A Multiword Tag"', tagId: 0 }]);
  const searchTokens = service.parseSearchTerm(
    'tag:""A Multiword Tag"" "potato"'
  );
  expect(searchTokens[1].length).toBe(1);
});
```
As previously established in other tests, right now the `parseSearchTerm` method allows for adding the `tag:` syntax in the middle of search terms and will even remove them from the actual search query. But this suddenly breaks when you add quotation marks into the normal search terms. This is just because it's a bit vague as to where tags end and begin with the current method of identifying tags.

I don't know if we want to fix this specific weird edge case as part of this ticket, since I highly doubt many people will run into it, but if we do want to fix it, then I think the best solution is to allow searching for tags with quotation marks using *escaped* quotation marks in the tags so that there is no uncertainty where tags begin and end. Even if we don't want to fix this issue in this specific ticket, I think I'd like to add a test that captures this erroneous behavior.

Regardless, I added some additional comments on the code for some other requested changes.

crisnicandrei:
> So I have a bit of a weird, probably very rare, edge case that breaks with this specific fix. It's a bit odd, but here's a test you can plug into the `ParseSearchTerm` describe block that will fail unexpectedly:
> 
> ```ts
> it('can handle tags with tags with quotation marks followed by search terms with quotation marks', () => {
>   tags.setTags([{ name: '"A Multiword Tag"', tagId: 0 }]);
>   const searchTokens = service.parseSearchTerm(
>     'tag:""A Multiword Tag"" "potato"'
>   );
>   expect(searchTokens[1].length).toBe(1);
> });
> ```
> 
> As previously established in other tests, right now the `parseSearchTerm` method allows for adding the `tag:` syntax in the middle of search terms and will even remove them from the actual search query. But this suddenly breaks when you add quotation marks into the normal search terms. This is just because it's a bit vague as to where tags end and begin with the current method of identifying tags.
> 
> I don't know if we want to fix this specific weird edge case as part of this ticket, since I highly doubt many people will run into it, but if we do want to fix it, then I think the best solution is to allow searching for tags with quotation marks using _escaped_ quotation marks in the tags so that there is no uncertainty where tags begin and end. Even if we don't want to fix this issue in this specific ticket, I think I'd like to add a test that captures this erroneous behavior.
> 
> Regardless, I added some additional comments on the code for some other requested changes.

@meisekimiu I am. having a bit of trouble comprehending the issue, sorry about that! Could you please give me an example using search terms and the expect one versus what pops up? I tried searching in the app, but the behaviour does not seem strange to me, or i am missing something and I want to know what. Thanks!

meisekimiu:
The issue is present in the test I provided but here's a more in-depth explanation.

Right now the `tag:` syntax can be placed anywhere in a search. Someone can search for "Test" and then add a tag to the search with "Test `tag:"testtag"`". But the order of these in the search string doesn't matter, and they can add the tag first. With the way we're parsing quotation marks, this can be an issue since the regular expression based approach is a bit greedy. If we search for a tag and then something with quotation marks, it will mess up. <q>_`tag:"testtag"`  "additional quote"_</u>.

I think we could see this as a separate issue though, which is why I'm not fully recommending a fix right now. If we choose to ignore it for now, I would at least prefer a test be written (something similar to the test I wrote although it can actually be simplified even more to my example above) to capture this behavior.

crisnicandrei:
@meisekimiu I wrote a test for the behaviour that you mentioned, what do you think?

meisekimiu:
That looks fine if we want to make another ticket to fix that issue later.

crisnicandrei:
@meisekimiu I am okay with creating a new ticket if you are.

k8lyn6 (approved):
This looks like it works as expected!

crisnicandrei:
@meisekimiu can you take a look at this and approve if everything looks good please?

meisekimiu (approved):

PR #302, opened 2023-09-22
PER-9361-Email and phone verification are broken;

crisnicandrei:
Changed the way isLoggedIn is handled in the verify component and made sure that keepLoggedIn is not being set to null after phone or mail verification.

Steps to test: 

1.Log in or create a new account.
2. Verify your email.
3. Refresh.

The button should not appear a second time now.

### DIFF
--- a/src/app/shared/services/account/account.service.spec.ts
+++ b/src/app/shared/services/account/account.service.spec.ts
@@ -140,6 +140,7 @@ describe('AccountService', () => {
 
     await instance.verifyEmail('sampleToken');
     expect(instance.getAccount().emailStatus).toBe('status.auth.verified');
+    expect(instance.getAccount().keepLoggedIn).toBeTrue();
   });
 
   it('should handle successful phone verification', async () => {
@@ -156,5 +157,6 @@ describe('AccountService', () => {
 
     await instance.verifyEmail('sampleToken');
     expect(instance.getAccount().phoneStatus).toBe('status.auth.verified');
+    expect(instance.getAccount().keepLoggedIn).toBeTrue();
   });
 });
--- a/src/app/shared/services/account/account.service.spec.ts
+++ b/src/app/shared/services/account/account.service.spec.ts
@@ -41,6 +41,32 @@ describe('AccountService', () => {
           },
           get: (account: AccountVO) => Promise.reject({}),
         },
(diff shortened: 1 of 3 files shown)


src/app/shared/services/account/account.service.spec.ts
```diff
@@ -99,4 +125,36 @@ describe('AccountService', () => {
...
+      primaryEmail: 'test@permanent.org',
+      fullName: 'Test User',
+      keepLoggedIn: true,
+      emailStatus: 'status.auth.unverified',
+    });
+
+    instance.setAccount(account);
+
+    await instance.verifyEmail('sampleToken');
+    expect(instance.getAccount().emailStatus).toBe('status.auth.verified');
```
meisekimiu, line 142:
```suggestion
    expect(instance.getAccount().emailStatus).toBe('status.auth.verified');
    expect(instance.getAccount().keepLoggedIn).toBeTrue();
```
Let's add a check for the `keepLoggedIn` behavior since that's what this particular ticket is all about. These tests otherwise just check for email/phone verification, and so pass even when copied over to the main branch. I'd like to see these tests fail if we run them if the specific bug addressed in this ticket isn't fixed, if that makes sense.

src/app/shared/services/account/account.service.spec.ts
```diff
@@ -99,4 +125,36 @@ describe('AccountService', () => {
...
+      primaryEmail: 'test@permanent.org',
+      fullName: 'Test User',
+      keepLoggedIn: true,
+      phoneStatus: 'status.auth.unverified',
+    });
+
+    instance.setAccount(account);
+
+    await instance.verifyEmail('sampleToken');
+    expect(instance.getAccount().phoneStatus).toBe('status.auth.verified');
```
meisekimiu, line 158:
```suggestion
    expect(instance.getAccount().phoneStatus).toBe('status.auth.verified');
    expect(instance.getAccount().keepLoggedIn).toBeTrue();
```

meisekimiu (approved):
Functionally these look good, I just have some notes about a small adjustment to the tests I'd like to see fixed. Thanks for fixing this bug!

k8lyn6 (approved):
This appears to be working now, thanks Andrei!

PR #303, opened 2023-09-25
PER-8369 Add "add tags" button/guide after tags have been created

crisnicandrei:
Added a new button that says Click to add when there are tags selected


meisekimiu (approved):

k8lyn6 (approved):
This looks good to me!

PR #304, opened 2023-09-28
PER-9254 Use branded icons for the image viewer

crisnicandrei:
Replaced the previous images for the open sea dragon library with permanent branded ones.


meisekimiu (approved):
There's really no code changes here, so I'm going add my approval already.

k8lyn6 (approved):
Looks good and everything appears to be working as previously

PR #305, opened 2023-09-29
Per 9139 publish modal shows extra slash

crisnicandrei:
Stepts to test: 

1.Select a record.
2. Click share
3. The generated url should have only one slash before p/archive

### DIFF
--- a/src/app/shared/pipes/public-link.pipe.spec.ts
+++ b/src/app/shared/pipes/public-link.pipe.spec.ts
@@ -1,3 +1,4 @@
+/* @format */
 import { FolderVO, RecordVO } from '@models';
 import { PublicLinkPipe } from './public-link.pipe';
 import { PublicRoutePipe } from './public-route.pipe';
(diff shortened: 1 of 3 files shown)


src/app/shared/pipes/public-link.pipe.spec.ts
```diff
@@ -32,10 +33,33 @@ describe('PublicLinkPipe', () => {
...
     });
     const route = pipe.transform(record);
     expect(route).toBeDefined();
-    expect(route.endsWith('/p/archive/0001-0000/0001-meow/1234/record/0001-00gp')).toBeTruthy();
+    expect(
+      route.endsWith('/p/archive/0001-0000/0001-meow/1234/record/0001-00gp')
+    ).toBeTruthy();
+  });
+
+  it('adds only one slash in the url before p/archive', () => {
```
meisekimiu, line 45: I think this test might be a bit too specific; what we probably  want is to make sure that there are no duplicate slashes in general. Instead of testing in one place (before the `p/archive`), I think it might make more sense to test the entire URL for double slashes (or rather the pathname of the URL since the `https://` bit at the beginning contains double slashes).

To make this testing easy we can take the string returned from `pipe.transform()` and put it into a URL object and test the pathname:
```ts
const url = new URL(pipe.transform(folder));
expect(url.pathname).not.toContain("//");
```

meisekimiu (approved):
This seems to work well! I left some notes about changes I'd like to be made to the test.

k8lyn6 (approved):
Looks good.

PR #306, opened 2023-10-02
PER-9247

crisnicandrei:
I have increased the duration of the message popup from 3 seconds to 9, and I have also added a button which will close the message when clicked on.

Steps to test: 

1. Any action that would display the message component.
2. The message should appear on the screen for 9 seconds.
3. If the user clicks on the icon, the message should clear.


meisekimiu (approved):

k8lyn6:
@crisnicandrei This looks good, but doesn't seem to apply to the phone and email verification messages. Is there a reason for that? Can we incorporate them as well?

crisnicandrei:
@k8lyn6 what messages did you test with? This might be something different because email verification was working for me

k8lyn6 (approved):
@crisnicandrei sorry, I must've not refreshed after I deployed to dev and tried the verify email and phone number messages, because it's working now! I think this will be a really good improvement.

PR #312, opened 2023-10-11
Per 9374 show amount invitation

crisnicandrei:
Display the gifted amount in the invitations table.

Steps to test: 
1.Send any amount a storage to an email that does not have a permanent account associated to it.
2. Go to the invitations dialog, pending invites.
3. It should be displayed there.

Note: This can be tested after #248 is merged, or by switching between the branches.


meisekimiu:
I've been waiting for #248 to be merged so that this can be QA'ed immediately afterwards. I think I would like to see a test that looks for the new invite amount being displayed in the UI though.

crisnicandrei:
@meisekimiu added more tests.

meisekimiu (approved):

PR #313, opened 2023-10-12
PER-9362 tags dissapearing for folders

crisnicandrei:
Added an additional check so fetchLeanItems does not get called each time preventing empty folders to be fetched.

@meisekimiu considering that records are being fetched if they do not have the thumbURL yet, shouldn't there be a condition for folders too? I will write tests for this but I just want to get the answer to this question so i can test that scenario.

### DIFF
--- a/src/app/models/folder-vo.ts
+++ b/src/app/models/folder-vo.ts
@@ -216,6 +216,7 @@ export interface FolderVOData extends BaseVOData {
   AccessVOs?: any;
   posStart?: any;
   posLimit?: any;
+  isFolder?: boolean;
 }
 
 export interface FolderSizeVOData {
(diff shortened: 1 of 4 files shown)


meisekimiu (approved):
I think the DataService should ideally work differently, since the root cause of this bug is that the DataService is overwriting a "fully fetched" item with a "lean" item and so it clears out the TagVO data field (as Lean items do not have defined TagVOs among many other data fields). The DataService should know not to do this, but after some serious research it seems like fixing this is a much larger architectural issue.

So I think for now this is a fine workaround. Thanks for fixing this! I'm not sure I understand your question, since this fix actively prevents folders from being fetched if they have no thumbnails. I think this solution is fine as is.

crisnicandrei:
@meisekimiu My question is this: I understand that for records that do not have a thumb url, to fetch them again. Shouldn't be there a case where folders are fetched again? For example if they are not somehow processed for some reason? I hope it makes sense now.

crisnicandrei:
@meisekimiu I think that can be a separate ticket because I don't know when we should refetch folders

PR #318, opened 2023-10-20
PER-9339: Add human readable transaction types

cecilia-donnelly:
With two new backend purchase types, we need corresponding human-readable names to appear in the transaction history. Also fix a (presumed) copy-paste error and fix the error message that appears if the transaction history doesn't load.

Needed for the changes in https://github.com/PermanentOrg/back-end/pull/514.


meisekimiu (approved):
This would be a good place to potentially write tests for the transaction history component, but that is a lot of work considering this is just a text change related to a back end fix. So this looks good to me!

PR #319, opened 2023-10-25
Per 9373 enter multiple emails gifting screen

crisnicandrei:
This pr adds the bulk gifting functionality to the gifting interface. Compared to the previous gifting interface, the user can now add multiple emails in the input separated by commas, gifting each user the same amount of GBs. 

Stepts to test: 

1. Go to the gifting storage tab
2. Add as many valid emails as you would like. Make sure to not exceed your storage
3. Add the gift amount
4. Send the storage.


crisnicandrei:
@cecilia-donnelly done

k8lyn6 (approved):
This looks good! Yay!

PR #321, opened 2023-11-02
Per 9385 long titles should expand on public archive full screen view

crisnicandrei:
Added a new flag which adds a class to the div on hover. This expands the name using css.

Steps to test: 
1..Go to any public file on full screen with a long name
2. Hover over it.

### DIFF
--- a/src/app/shared/components/inline-value-edit/inline-value-edit.component.spec.ts
+++ b/src/app/shared/components/inline-value-edit/inline-value-edit.component.spec.ts
@@ -386,7 +386,7 @@ describe('InlineValueEditComponent', () => {
     expect(Number(utcDt.format('h'))).toEqual(3);
   });
 
-  it('should expand the title when hovering with the mouse over it', () => {
+  it('should expand the title when hovering with the mouse over it', async () => {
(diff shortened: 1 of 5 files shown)


src/app/shared/components/inline-value-edit/inline-value-edit.component.spec.ts
```diff
@@ -357,4 +385,20 @@ describe('InlineValueEditComponent', () => {
...
+
+    const nameContainer = fixture.debugElement.query(
+      By.css('.inline-value-text')
+    ).nativeElement;
+    nameContainer.dispatchEvent(new Event('mouseenter'));
+    fixture.detectChanges();
+
+    fixture.whenStable().then(() => {
+      expect(nameContainer.classList).toContain('is-name-hovered');
+    });
```
meisekimiu, line 402: Just for consistency can you make this test `async` and use `await fixture.whenStable()` instead?

meisekimiu (approved):
The code looks good to me!

k8lyn6 (approved):
Looks great!

PR #322, opened 2023-11-07
Per 9248 panning issues

crisnicandrei:
Disable the panning when the zoom level of the image in the viewer is the first level or if it is zoomed out.

Stepts to test: 
1. Choose any image and display it full-screen.
2. Zoom out or leave it at the first zoom level
3. It should not pan when dragging it around.


k8lyn6:
@meisekimiu I don't know what issues you were seeing, but it looks like for me, when I zoom out and then try to swipe, it doesn't work. I don't know if that's something we want to fix in this ticket? The panning issue seems to be resolved.

crisnicandrei:
@k8lyn6 swipe as in swipe to the next image or moving the image around?

k8lyn6:
@crisnicandrei swipe to the next image

crisnicandrei:
@k8lyn6 fixed the swiping issue too

PR #324, opened 2023-11-09
PEr-9396 Send the email to mixpanel

crisnicandrei:
Steps to test:
1. Log in/register any account.
2. Check mixpanel.
3. It should send the email now.

### DIFF
--- a/src/app/shared/services/account/account.service.spec.ts
+++ b/src/app/shared/services/account/account.service.spec.ts
@@ -180,7 +180,7 @@ describe('AccountService', () => {
     await instance.verifyEmail('sampleToken');
     expect(instance.getAccount().emailStatus).toBe('status.auth.verified');
     expect(instance.getAccount().keepLoggedIn).toBeTrue();
-    expect(trackerSpy).toHaveBeenCalled()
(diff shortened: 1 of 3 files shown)


src/app/shared/services/account/account.service.spec.ts
```diff
@@ -182,6 +215,39 @@ describe('AccountService', () => {
...
+    const trackerSpy = spyOn(instance, 'trackAuthWithMixpanel');
+    const account = await instance.logIn(
+      'test@permanent.org',
+      'password123',
+      true,
+      true
+    );
+
+    expect(trackerSpy).toHaveBeenCalled();
+  });
```
meisekimiu, line 250: These are some particularly fragile tests since they are relying on calling a specific function in the AccountService that are highly coupled to Mixpanel. Since they've already been written, I'm fine with keeping them, but for the future we don't really need to fully cover the code that calls out to Mixpanel in particular unless it's doing something particularly complicated.

meisekimiu (approved):

k8lyn6 (approved):
Looks good, thanks!

PR #325, opened 2023-11-10
Display the promo code in the gift form

crisnicandrei:
Steps to test: 

Access the url ending with ?promoCode=TellYourStory. The promo code should be displayed in the form.

### DIFF
--- a/src/app/core/components/storage-dialog/storage-dialog.component.ts
+++ b/src/app/core/components/storage-dialog/storage-dialog.component.ts
@@ -68,8 +68,8 @@ export class StorageDialogComponent
 
       this.route.queryParamMap.subscribe((params) => {
         const param = params.get('promoCode');
-        if (this.activeTab === 'promo' && param === this.TELLYOURSTORY) {
-          this.promoForm.setValue({ code: this.TELLYOURSTORY });
+        if (this.activeTab === 'promo' && param) {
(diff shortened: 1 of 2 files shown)


src/app/core/components/storage-dialog/storage-dialog.component.ts
```diff
@@ -47,15 +54,28 @@ export class StorageDialogComponent implements OnInit, IsTabbedDialog {
...
+        } else {
+          this.activeTab = 'add';
+        }
+      }),
+
+      this.route.queryParamMap.subscribe((params) => {
+        const param = params.get('promoCode');
+        if (this.activeTab === 'promo' && param === this.TELLYOURSTORY) {
+          this.promoForm.setValue({ code: this.TELLYOURSTORY });
+        }
```
meisekimiu, line 72: Instead of hardcoding it to only work on the `TellYourStory` code, do we want to generalize it so it can take any code? That way in the future if we want to supply a link to another code we don't have to update the code at all.

meisekimiu (approved):

src/app/core/components/storage-dialog/storage-dialog.component.ts
crisnicandrei: good point!

k8lyn6 (approved):
Looks good!

PR #326, opened 2023-11-13
PER-9219 Allow sftp deletion tab.

crisnicandrei:
Create new component and a new tab for advanced settings.

In the advanced settings tab handle the sftp deletion toggler. The toggler is similar to the display download button toggler.

Steps to test:

1. Open the account dropdown and navigate to the new "Advanced Settings" tab.
2. The toggle should work setting the option on or off.

### DIFF
--- a/src/app/core/components/account-settings-dialog/account-settings-dialog.component.html
+++ b/src/app/core/components/account-settings-dialog/account-settings-dialog.component.html
@@ -48,9 +48,20 @@
       >
         Delete Account
       </div>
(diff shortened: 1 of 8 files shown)


meisekimiu (approved):

k8lyn6 (approved):
Noting that Filezilla does not allow syncing, so could not test this in action using Filezilla. However, we checked the database and the toggle is correctly changing the sftp deletion on/off appropriately. Looks good!

PR #327, opened 2023-11-14
Per 9185 update public archive design

crisnicandrei:
Updated the design of the public archive. 

@k8lyn6 if you see something that is off please let me know and i will take care of it.


k8lyn6 (changes requested):
It's a little hard to tell what was changed, but here are a few things I found (they could've been existing, I'm not sure):

- Did the banner get larger? I think it's taking up too much room. I think we should go back to what it was if it was changed.
- The search bar is getting cut off a little when the browser is a certain size. See photo below:
Screen Shot 2023-11-15 at 1 11 17 PM
- When you search for something and get a result that has a very long title, the image thumbnail is getting cut off. See photo below:
Screen Shot 2023-11-15 at 1 13 42 PM

PR #328, opened 2023-11-15
PER-9015 Changed the click event of the modals to mousedown

crisnicandrei:
Changed the click event of the modals to mousedown

Steps to test:
1. Open any modal.
2. Click inside it, holding the button down, then dragging outside of the modal
3. It should not close the modal when the button is release outside of it.


meisekimiu (approved):

k8lyn6 (approved):
Looks good!

PR #329, opened 2023-11-22
PER-8826-archive-members-with-viewer-role-cannot-get-link

crisnicandrei:
Enable the get link button for viewers, but disable the publish to internet archive button.

Steps to test:
1. Invite an account as a viewer to your archive.
2. Switch to that account
3. Go to the public files tab
4. The get link button should work


k8lyn6 (approved):
Looks good!

PR #331, opened 2023-11-23
PER-8400-toggle-person-info-public-archive

crisnicandrei:
PER-8400-toggle-person-info-public-archive
Added a chevron that expands the profile information on mobile devices.

Steps to test:

1.Go to any public archive on mobile view.
2. Click the arrow expanding the information

### DIFF
--- a/src/app/public/components/public-archive/public-archive.component.html
+++ b/src/app/public/components/public-archive/public-archive.component.html
@@ -33,6 +33,7 @@
             class="icon-expand"
             [class.reverse]="showProfileInformation"
             [icon]="['fas', 'chevron-down']"
+            *ngIf="!(isViewingProfile$ | async)"
           ></fa-icon>
         </div>
         <div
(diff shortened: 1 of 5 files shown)


meisekimiu (approved):

k8lyn6 (changes requested):
I'm so sorry for the delay! I think when you switch over to the archive profile, we don't need a drop down because the user is there to see the profile and it isn't getting in the way of anything.

crisnicandrei:
@k8lyn6 did the changes!

k8lyn6 (changes requested):
I don't know if it's something on my end, but it looks like the "Established" and the summary section don't immediately show up unless you click to the archive and then back to the profile. It also looks like the "see more" isn't working. Here's a video to show you what I mean.

https://github.com/PermanentOrg/web-app/assets/16201121/5090af06-35e3-44cf-b732-3c1b5696d0a5

crisnicandrei:
@k8lyn6 thanks! I found the issue and fixed it.

k8lyn6 (approved):
Looks good now!

PR #332, opened 2023-11-23
PER-8254: show error message no byytes

crisnicandrei:
Display a custom error message if the file size is 0 bytes.

Steps to test:
1. Upload a file that has zero bytes
2. The custom message should be displayed.


k8lyn6 (approved):
Looks good to me!

PR #333, opened 2023-11-27
Added the missing mixpanel events

crisnicandrei:
This pr adds the missing mixpanel events from the confluence document.

I have decided to make one big pr instead of many small ones because I have added 2 new methods to the mixpanel and device services which i needed, so instead of branching multiple times I made one bigger pr.

### DIFF
--- a/src/app/core/components/left-menu/left-menu.component.ts
+++ b/src/app/core/components/left-menu/left-menu.component.ts
@@ -16,7 +16,6 @@ import { find } from 'lodash';
 import * as Sentry from '@sentry/browser';
 import debug from 'debug';
 
-import { DeviceService } from '@shared/services/device/device.service';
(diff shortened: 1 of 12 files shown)


crisnicandrei:
@meisekimiu you are right, I forgot to remove the imports from some places

crisnicandrei:
@meisekimiu in the future i will keep prs smaller, but i needed that service method, and I did not want to keep rebranching for all the other prs.

meisekimiu:
The pull request itself doesn't need to be smaller, I just think separating each event into its own git commit within the PR would keep things more organized.

crisnicandrei:
@meisekimiu i get it now! Thanks!

k8lyn6 (approved):
Everything looks good now! Thanks!

PR #334, opened 2023-11-28
PER-8957 Drag and drop file to the public workspace

crisnicandrei:
Display the prompt. with the confirmation message (same as with upload) when dragging file into the public workplace.

Steps to test: 
1. Drag a file, or more, to the public workspace
2. The prompt with the confirmation message should appear.


k8lyn6 (approved):
This looks good!

PR #336, opened 2023-12-04
PER-8919: Public Archives Landing Page

crisnicandrei:
Steps to test:
1. Go to the app gallery tab from the left side nav bar.
2. A list of public archives should be displayed, or a message if no public archives exist.

- Added public archives component in which a list of public archives is displayed, with the option to click on an archive and getting redirected to the archive page.
- Added access role pipe to display the access role of the user in the public archives list.
- Added unit tests for the new pipe and the new component.


src/app/gallery/components/public-archives/public-archives.component.scss
```diff
@@ -0,0 +1,108 @@
...
+    flex-direction: column;
+  }
+}
+
+.public-archives-list {
+  display: flex;
+  flex-direction: column;
+  gap: 10px;
+  max-height: 320px;
+  overflow: scroll;
```
meisekimiu, line 20: In the design I only see vertical scrollbars (and no scrollbars in mobile view). I think that at the very least this should have either
```css
overflow-x: auto;
/* OR */
overflow-x: hidden;
```
rules enabled so that the horizontal scrollbars disappear. It isn't clear in the design if we want scrollbars always visible even if there are not enough archives in the list to result in scrollbars; we should get some clarification on that from the experience team.

src/app/gallery/components/public-archives/public-archives.component.ts
```diff
@@ -0,0 +1,36 @@
...
+import { AccountService } from '@shared/services/account/account.service';
+import { Router } from '@angular/router';
+import { MessageService } from '../../../shared/services/message/message.service';
+
+@Component({
+  selector: 'pr-public-archives',
+  templateUrl: './public-archives.component.html',
+  styleUrls: ['./public-archives.component.scss'],
+})
+export class PublicArchivesComponent implements OnInit {
```
meisekimiu, line 13: Although I know it's difficult to rename components in Angular, I really think this should have a different name. The name "PublicArchivesComponent" isn't very meaningful since we have a lot of code dealing with Public Archives in the web-app. Maybe something like "MyPublicArchivesComponent" or "PublicArchivesListComponent" could be better names for this. From my initial skimming of the code I wasn't able to tell at a glance whether this was indeed a new component or just an old component we had refashioned since the name was so generic.

src/app/shared/pipes/access-role.pipe.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+    expect(pipe).toBeTruthy();
+  });
+
+  it('returns the correct access role', () => {
+    const value = 'access.role.viewer';
+    const result = pipe.transform(value);
+    expect(result).toEqual('Viewer');
+  });
+
+  it('returns the correct access role', () => {
```
meisekimiu, line 20: Though these are really basic tests that shouldn't ever fail, I think we should give each test a unique name so that when one fails it communicates which access role isn't being converted properly.

meisekimiu (changes requested):
Left some comments on a few basic style/naming issues. This does work from my initial testing though. I think it might be useful to go ahead and add @k8lyn6 as a reviewer to comment on the style issue I bring up in the comments.

crisnicandrei:
@k8lyn6 please let us know what do you think!

src/app/gallery/components/public-archives/public-archives.component.scss
k8lyn6: @crisnicandrei I don't think we want horizontal scrollbars (which I'm not seeing) and probably don't want the scrollbars if there's not enough archives in the list to result in scrollbars. I also think the scrollbar on mobile would be a little hard to use. I think the design for mobile shows a "see more" after the first three archives, that would expand to a list of all your public archives. Could we make that change?

crisnicandrei:
@k8lyn6 I did the changes!

k8lyn6 (approved):
Looks good!

crisnicandrei:
@meisekimiu can you take a look at this please?

src/app/shared/pipes/access-role.pipe.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+    pipe = new AccessRolePipe();
+  });
+  it('create an instance', () => {
+    expect(pipe).toBeTruthy();
+  });
+
+  it('returns the correct access role for the viewer', () => {
+    const value = 'access.role.viewer';
+    const result = pipe.transform(value);
+    expect(result).toEqual('Viewer');
```
meisekimiu, line 17: Can you fix the lint warnings in this file? You only have to do them in the new files and changes.

meisekimiu (approved):

PR #339, opened 2023-12-14
Don't redirect invalid session on Public Gallery

meisekimiu:
Fix a bug where an invalid session on the public gallery could redirect the user to the login screen. Since authentication is not needed in the public gallery, they do not need to be redirected.

While this fixes some bug that deals with redirections to login, it's unclear if this is actually fixing the bug referred to in PER-9409. The invalid session redirect is one of the few pathways available for redirection to the login screen on the public gallery, and it seems to be the only one that does not provide an error message as seen in the one video recording we have of the bug in action. However, it's unclear how this specific code would be triggered in a brand new browser that doesn't have any session data in it to begin with, which is what was observed in the video.

To keep the AccountService tests more organized, the tests for this change were put in a separate file since the main AccountService spec file was already getting very long. Because of the way Angular unit testing works, there's a limit to how much code duplication we can eliminate, so we may have to split additional tests into their own files to keep things more organized.

**Steps to Recreate Bug:**
1. Log into Permanent
2. Open a page on a public archive.
3. Open up developer tools or browser settings and delete *only* the cookies for the current Permanent environment.
4. Refresh the page and verify that you are redirected to login instead of seeing the public archive.
5. Also verify that no error message is displayed to the user

**Steps to test this PR:**
1. Recreate the bug as described below.
2. Verify that the user is not redirected.

Resolves PER-9409.

### DIFF
--- a/src/app/shared/services/account/account.service.ts
+++ b/src/app/shared/services/account/account.service.ts
@@ -12,6 +12,7 @@ import {
   ArchiveResponse,
   FolderResponse,
 } from '@shared/services/api/index.repo';
+import { LocationStrategy } from '@angular/common';
 import { Router } from '@angular/router';
 import { Dialog } from '@root/app/dialog/dialog.module';
 import {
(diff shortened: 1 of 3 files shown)


meisekimiu:
I'm investigating why the second `catch()` block isn't being run in the tests. It should be, but if somehow it's never run despite a promise rejection from `api.account.get`, we can just delete that block entirely.

liam-lloyd (approved):

crisnicandrei (approved):

PR #343, opened 2023-12-19
Per 8562 indicate fields that are editable

crisnicandrei:
Steps to test:
1.Open any uploaded file in the full screen viewer.
2. Notice that an icon and Click to add text is present for tags and location (description already had a helper text and the name cannot be deleted.)

### DIFF
--- a/src/app/file-browser/components/file-viewer/file-viewer.component.ts
+++ b/src/app/file-browser/components/file-viewer/file-viewer.component.ts
@@ -49,7 +49,6 @@ export class FileViewerComponent implements OnInit, OnDestroy {
   public documentUrl = null;
 
   public canEdit: boolean;
-  public canEditFieldsOnFullscreen = false;
 
   // Swiping
   private touchElement: HTMLElement;
@@ -121,8 +120,6 @@ export class FileViewerComponent implements OnInit, OnDestroy {
         this.currentRecord.accessRole,
         AccessRole.Editor
       ) && !route.snapshot.data?.isPublicArchive;
-
-    this.canEditFieldsOnFullscreen = this.canEdit && this.canEditOnFullScreen();
 
     this.tagSubscription = this.tagsService
       .getItemTags$()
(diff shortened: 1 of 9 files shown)


src/app/file-browser/components/file-viewer/file-viewer.component.ts
```diff
@@ -121,6 +122,8 @@ export class FileViewerComponent implements OnInit, OnDestroy {
         AccessRole.Editor
       ) && !route.snapshot.data?.isPublicArchive;
 
+    this.canEditFieldsOnFullscreen = this.canEdit && this.canEditOnFullScreen();
```
meisekimiu, line 125: I'm a bit confused on the purpose of this variable. Is there ever a case where `canEdit` and `canEditFieldsOnFullscreen` are going to be different values? I'm not sure if there's a special case but I would think that anything that fullscreen editing shouldn't have different permissions from normal editing (especially since `canEdit` already checks if the file in question is a Public Archive or not).

meisekimiu (commented):
Left a question about some of the new inputs introduced in the code. Also, while the ticket mentioned that we might want to include an icon, I will say that adding that icon creates visual inconsistency between the fullscreen and standard metadata editors; the fullscreen one has icons and the "normal" editor doesn't. This is more of a thing for @k8lyn6 to discuss during review but I wanted to bring it up.

src/app/file-browser/components/file-viewer/file-viewer.component.ts
crisnicandrei: @meisekimiu  good point, but the ticket was suggesting adding the text only in fullscreen mode, not in the file list viewer, that's my reasoning. That is why I am making use of the routes. Should I change it?

cecilia-donnelly: Outcome: we'll use `canEdit` everywhere.

k8lyn6 (changes requested):
This looks good! I think we should probably remove the icon -- I like it better without them, and it looks like they don't show up consistently anyway (I wasn't seeing it on the tags if you already have tags entered)

meeksie (approved):
Works well!

k8lyn6 (approved):
Looks good!

PR #344, opened 2024-01-09
PEr-8276 inform user about upload destination

crisnicandrei:
Display a different message when dragging a file onto a folder.

I have noticed that when dragging a file onto a folder, not the root, the `parentFolder`'s `pathAsArchiveNbr` value is always an empty array, so when the array is empty, I am displaying a different message.

Steps to test:
1Drag a file onto an uploaded folder.
2. The message should be "Uploading 'file' to 'folder'" instead of just the file's name

### DIFF
--- a/src/app/core/components/upload-progress/upload-progress.component.html
+++ b/src/app/core/components/upload-progress/upload-progress.component.html
@@ -21,11 +21,7 @@
   >
     <div class="current-file">
       Uploading {{ currentItem.file.name }} to
-      {{
-        isUploadingFolder
-          ? folderTargetName + '/' + currentItem.parentFolder.displayName
-          : currentItem.parentFolder.displayName
-      }}
+      {{ folderTargetName }}
     </div>
     <div class="file-count">
       <strong>{{ fileCount.current }}</strong> of
--- a/src/app/core/components/upload-progress/upload-progress.component.html
+++ b/src/app/core/components/upload-progress/upload-progress.component.html
@@ -21,7 +21,11 @@
   >
     <div class="current-file">
       Uploading {{ currentItem.file.name }} to
-      {{ currentItem.parentFolder.displayName }}
+      {{
+        isUploadingFolder
+          ? folderTargetName + '/' + currentItem.parentFolder.displayName
+          : currentItem.parentFolder.displayName
+      }}
     </div>
     <div class="file-count">
       <strong>{{ fileCount.current }}</strong> of
--- a/src/app/core/components/upload-progress/upload-progress.component.html
+++ b/src/app/core/components/upload-progress/upload-progress.component.html
@@ -1,10 +1,9 @@
 <!-- @format -->
 <div
-  class="upload-wrapper"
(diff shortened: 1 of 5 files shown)


src/app/core/components/upload-progress/upload-progress.component.html
```diff
@@ -13,7 +20,13 @@
...
     *ngIf="status === UploadSessionStatus.InProgress && currentItem"
   >
-    <div class="current-file">{{ currentItem.file.name }}</div>
+    <div *ngIf="!isUploadingInNestedFolder" class="current-file">
+      {{ currentItem.file.name }}
+    </div>
+    <div *ngIf="isUploadingInNestedFolder" class="current-file">
+      Uploading {{ currentItem.file.name }} to
+      {{ currentItem.parentFolder.displayName }}
+    </div>
```
cecilia-donnelly, line 29: @crisnicandrei I think this can be simplified. We can show the destination folder for all uploads, not just nested folders.

cecilia-donnelly (commented):
Just what it says below. The nested folder code seems to add some complexity that's unnecessary.

crisnicandrei:
@cecilia-donnelly thank you! Did the change just now!

meisekimiu (approved):

omnignorant:
I tested in Chrome, Firefox and Safari using individual and multiple files, individual and multiple folders and a mix of files and folders.

Results matched the expectation for individual and multiple files but not with folders.

- When a folder (source) with contents (file) is dragged from the user’s computer onto a folder in their permanent archive (destination), the expected upload message should be “uploading file to destination/source” 
- it currently shows “uploading file to source” instead
- Presumably this is because the message is using the subfolder name where the files are going, e.g. source is created as a subfolder of destination on Permanent and then files are uploaded to this new subfolder and so the message uses the subfolder's name instead of the parent folder which is the intended destination for source and file.
- This behavior persisted for multiple folders with the message cycling through source 1, source 2, etc.

@crisnicandrei can you fix this to match the expected behavior for folders or should we close this as done and open a new ticket for improvement? cc @k8lyn6

crisnicandrei:
@omnignorant Thank you for your message! I have made the changes asked! Please let me know if it works for you now!

omnignorant (approved):
looks good now! great work @crisnicandrei and I would merge this and move on.

that being said, two new tickets to open @k8lyn6 :
1. when you hover over the window to drag/drop upload, the window is highlighted in yellow/orange with the destination folder appearing in the upload message at the top. When you withdraw (if you've changed your mind or whatever) and move your cursor off the window, the window remains highlighted and it cannot be undone.
2. The highlighted upload message for drag/drop in Private workspace still reads "My Files" and it should be "Private".

crisnicandrei:
@omnignorant "The highlighted upload message for drag/drop in Private workspace still reads "My Files" and it should be "Private"." I have fixed this in this ticket now since it made sense to me and it was a quick fix. So we need only one ticket now.

PR #346, opened 2024-01-11
PER-9438-search-results-in-subfolders

crisnicandrei:
Added the archive number of the parent folder and the link id when navigating in the search results into a nested record, thus forcing the data service to change the current folder.

Steps to test: 
1. Publish a folder with any record inside of it.
2. Go to the public archive and search for the nested record
3. It should open the record in the file viewer.


meisekimiu (approved):

emlesienk (changes requested):
The search wasn't working on dev when I tried. Branch is out of date so you might need to update for the new code to fix this issue.

crisnicandrei:
@emlesienk did this work?

emlesienk:
> @emlesienk did this work?

Yes, thank you.

PR #347, opened 2024-01-17
Upgrade to Angular 17

meisekimiu:
Upgrade Angular to version 17, and upgrade all dependent packages to compatible versions as well.

This upgrade was actually quite small; I mainly ran into issues with testing it and running/upgrading Storybook. Only one file had to be manually changed and only two files were upgraded by automatic Angular migrations.

**This upgrade now requires Node 18 to be installed.**

**Steps to Test:**
1. Run `npm install`
2. Run `npm run build` / `npm run build:local` to verify that a production build can be made without errors.
3. Run unit tests to verify that they can be run without errors and no functionality covered by tests is broken.
4. Run `npm run storybook` to verify Storybook can be run without errors <small>(I always forget this one!)</small>
5. Run `npm run dev` to verify that the local development server can be run without errors
  - Click around in the web-app to make sure nothing's broken!

Resolves #345. Resolves PER-9430.


meisekimiu:
I will include a new type definition file for the `vis-timeline` types before merging this PR.

meisekimiu:
After looking at the type definitions for vis-timeline more, I think it makes sense to write a helper method instead of rewriting the type information for the vis-timeline package. While in practice it seems like we're only ever getting `number` type for the `DateType` type (a union of `string`, `number`, and `Date` types), I think it's best to just write a conversion method that handles all three cases instead.

liam-lloyd (approved):

PR #348, opened 2024-01-18
Per 9422 mobile banner

crisnicandrei:
I have implemented a banner, that when clicked, would redirect the user to the app installed on their device, or to the store.

Steps to test if testing on a computer or laptop:
1. Open the app in the log in, onboarding or any workspace
2. Inspect the app with devtools
3. Go into responsive mode and choose either an android or ios phone
4. Refresh the page and the banner should show up
5. By clicking the banner after 1.5 seconds the user should see a prompt (this is for security reasons to please Chromium browsers) and by clicking yes they would be redirected to the store

Steps to test if on a mobile phone:
1. Open the app in the log in, onboarding or any workspace
2. Go into responsive mode and choose either an android or ios phone
3. By clicking the banner after 1.5 seconds the user should see a prompt (this is for security reasons to please Chromium browsers) and by clicking yes they would be redirected to the store, or if the app is installed, the app should open when clicking the banner without the prompt appearing.


meisekimiu:
I don't think we should be redirecting users manually if the deep link doesn't work, since that would just make the entire web app completely unusable on mobile.

crisnicandrei:
@meisekimiu do you have any suggestions? Should we look for a library that could handle this?

crisnicandrei:
@meisekimiu https://github.com/ain/smartbanner.js

src/app/shared/services/banner/banner.service.spec.ts
```diff
@@ -0,0 +1,53 @@
+import { NoopAnimationsModule } from '@angular/platform-browser/animations';
+/* @format */
```
meisekimiu, line 2:
```suggestion
/* @format */
import { NoopAnimationsModule } from '@angular/platform-browser/animations';
```

src/app/shared/services/banner/banner.service.ts
```diff
@@ -0,0 +1,43 @@
...
+
+@Injectable({
+  providedIn: 'root',
+})
+export class BannerService {
+  private _isVisible = false;
+  private _appStoreUrl =
+    'https://apps.apple.com/app/permanent-archive/id1571883070' as const;
+  private _playStoreUrl =
+    'https://play.google.com/store/apps/details?id=org.permanent.PermanentArchive' as const;
```
meisekimiu, line 13: Why the `as const` here? Could we use `private readonly` instead?

meisekimiu (commented):
I think the `pr-banner` component should be renamed to something a bit more descriptive, such as `pr-mobile-banner`, since "banner" is a bit too vague on its own.

meisekimiu (approved):

k8lyn6 (changes requested):
This is looking really good! A few things (cc @yeslikesolo):

- On Safari on iOS, when you tap Go to App, it pops open a dialog from Safari saying that the address is invalid. Is there anything you can do to fix that for Safari? See photo.
IMG_2070
- If you do have the app installed on iOS and tap Go to App, it takes you to the app store rather than the app. Is there a way to actually redirect them to the app itself rather than the app store? Is that a dev vs prod issue? It's ok if we can't redirect to the app itself -- the app store is fine.
- Some text changes: For the headline, can we say "Unlock more with our app." And then for the text, can it say "Experience full functionality – download our app to enjoy all features seamlessly."

crisnicandrei:
@k8lyn6 for the app to redirect to the app instead of the store we would need to first merge this pr so the mobile team to be able to make their changes! That "invalid page" message on safari is because the web app is trying to open the installed app, but i can try to make it disappear somehow

k8lyn6:
@crisnicandrei gotcha! Thanks for looking into the safari issue. I think it's okay if you aren't able to make that change.

cc @yeslikesolo

k8lyn6 (changes requested):
Small wording change: can we change the description to say "download the Permanent app to enjoy all features seamlessly"?

And a minor spacing thing: Are we able to increase the spacing between "Experience full functionality - download the Permanent app to enjoy all features seamlessly" and the "Go to App" button to make it equal to the others? It looks a little smaller than the spacing between the others.

I also deleted the app and still got the "invalid link" on Safari. Were you able to look into that error message?

cc @yeslikesolo

k8lyn6 (approved):
Looks awesome! Thanks so much Andrei!

PR #350, opened 2024-01-30
PER-9444 toggle component

crisnicandrei:
Created a toggle component with 3 states: enabled, checked and not checked.

If the toggle is enabled it emits back to the parent 'true', otherwise it emits 'false'. the toggle component looks and behaves just like the ios one.

@meisekimiu I am still unsure if the toggle should act as a way to choose between to values, if the toggle is enabled pick value a, otherwise pick value b? I think for that we should use checkboxes instead and the toggler should be used only for true/false? This just crossed my mind and wanted your opinion


meisekimiu (commented):
I think toggles should only be for true/false values. We have a lot of instances of having a "on / off" radio button in our app and I think eventually we might want to replace those with toggles for a more modern user experience.

Can you create a new module for the component library? Maybe something like `ComponentsModule` or something? I think we should aim to separate these from the rest of the things in the `SharedModule` if we can. The idea behind having a separate UI components module is that it is at one end of the dependency chain; these UI components can be used and inherited by all other modules but they don't have any dependencies themselves so can be isolated.

src/app/auth/auth.routes.ts
```diff
@@ -63,6 +64,7 @@ const routes: Routes = [
     SharedModule,
     AnnouncementModule,
     RecaptchaModule,
+    ComponentsModule,
```
meisekimiu, line 67: Since we're not using these components yet, I don't see a reason to include this module in any other modules. Hopefully they will only be imported when actually used to count down on unused imports.

meisekimiu (changes requested):

src/app/shared/shared.module.ts
```diff
@@ -80,6 +81,7 @@ import { AccessRolePipe } from './pipes/access-role.pipe';
     NgbDropdownModule,
     NgbPaginationModule,
     FontAwesomeModule,
+    ComponentsModule,
```
meisekimiu, line 84:
```suggestion
```
Let's get rid of all ComponentsModule imports for now until they're actually needed.

meisekimiu (approved):

PR #353, opened 2024-02-05
PER-8458-profile-completion

crisnicandrei:
I have removed some of the values from the profile completion checklist, because some of them were not used in the profile, preventing it to reach 100%. I have also made it easier for the profile to reach 100%, needing only 1 entry for each category.

Steps to test: 
1. Go to the Archive profile tab
2. Add one entry to each section
3. The profile percentage should then reach 100%

### DIFF
--- a/src/app/shared/services/profile/profile.service.spec.ts
+++ b/src/app/shared/services/profile/profile.service.spec.ts
@@ -1,16 +1,16 @@
-import { RecordVO } from './../../../models/record-vo';
 /* @format */
 import { FolderPickerService } from '@core/services/folder-picker/folder-picker.service';
 import { ArchiveVO, FolderVO } from '@models/index';
 import { SharedModule } from '@shared/shared.module';
 import { Shallow } from 'shallow-render';
+import { ProfileItemVOData } from '@models/profile-item-vo';
+import { RecordVO } from '../../../models/record-vo';
+import { ArchiveResponse } from '../api/archive.repo';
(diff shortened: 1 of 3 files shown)


src/app/shared/services/profile/profile.service.spec.ts
```diff
@@ -0,0 +1,199 @@
...
+      .dontMock(ProfileService);
+  });
+
+  it('should exist', async () => {
+    const { instance } = await shallow.createService();
+
+    expect(instance).toBeTruthy();
+  });
+
+  it('should return the correct completion value', async () => {
```
meisekimiu, line 79: Please give each test a unique name.

meisekimiu (changes requested):
Left some feedback on the tests.

Have you tested this on different archive types? (Personal vs Group vs Organization). These different types have differing metadata fields.

crisnicandrei:
@meisekimiu I did test on different archive types

src/app/shared/services/profile/profile.service.spec.ts
crisnicandrei: @meisekimiu done

meisekimiu (approved):

k8lyn6 (approved):
I think this looks good and solves the problem of never getting to 100%. There are some further small improvements I think we could make, but I'll make separate tickets for those. Thanks Andrei!

PR #354, opened 2024-02-05
PER-9445-button component

crisnicandrei:
PER-9445-button component

Implementation of a reusable button component. This still needs a disabled design.


meisekimiu (changes requested):
I am also requesting a Storybook story for this PR (see #350). I'd also like to see this component be included in the `ComponentsModule` made in #350, although obviously that one has to be merged first (or this one based on that branch) so that might have to wait a while. Sorry for that, in hindsight I think we should have had a ticket for creating the Component Library Module first so the other tickets wouldn't be blocking each other like this.

meisekimiu (changes requested):
Now that #350 has been merged, can you move the ButtonComponent to be in the components module?

meisekimiu (approved):

PR #356, opened 2024-02-07
PER-9459 broken public search

crisnicandrei:
Replaced the default value of the tags in the api call.

Steps to test:
1. Go to any public archive
2. Search for any file or folder
3. It should display the search results.

### DIFF
--- a/src/app/shared/services/api/search.repo.spec.ts
+++ b/src/app/shared/services/api/search.repo.spec.ts
@@ -1,0 +1,48 @@
+/* @format */
+import { TestBed } from '@angular/core/testing';
+import {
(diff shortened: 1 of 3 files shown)


crisnicandrei:
@meisekimiu I did write some tests a while ago using arrays instead of strings. The string was a mistake of mine which slipped back then

crisnicandrei:
@meisekimiu sorry, I have confused the repo with the service! Will do!

meisekimiu (approved):

emlesienk (approved):
Search within archives yields results that match titles of files or folders. Search yields results of files that are nested within folders as well.

PR #358, opened 2024-02-09
PER-9446-form input

crisnicandrei:
Implementation of the form input component from the new component library.

I have kept the old implementation and added a way to also use it without a form control, also using a debounce + style changes.

This pr also includes the new fonts for the new design system.


cecilia-donnelly:
Looks like the linting is failing, probably because of the conflicts @crisnicandrei -- can you rebase and fix any lint errors? Thanks!

meisekimiu (approved):

PR #359, opened 2024-02-13
PER-8903 pass group instead of family

crisnicandrei:
Change all the instances of family to group. Now the frontend handles group archives, creating dispalying editing. This will work once the back-end is done.

Steps to test: 
1. Open devtools network tab.
2. Create or change the type of an archive to be a group.
3. The request should be successful .


src/app/core/components/archive-switcher/archive-switcher.component.ts
```diff
@@ -145,12 +145,12 @@ export class ArchiveSwitcherComponent implements OnInit, AfterViewInit {
...
           {
-            text: 'Person',
-            value: 'type.archive.person'
+            text: 'Group',
+            value: 'type.archive.group'
           },
           {
             text: 'Family',
-            value: 'type.archive.family'
+            value: 'type.archive.group'
```
meisekimiu, line 153: This seems a bit incorrect.

crisnicandrei: @meisekimiu thanks Natalie

meisekimiu (approved):
This looks good, although as you note it should not be merged until back end work is completed to support this change fully.

liam-lloyd (changes requested):
It looks to me like this PR is going beyond what PER-8903 covers, and in addition to passing `type.archive.group` to the API where we now pass `type.archive.family`, it starts expecting to _receive_ `type.archive.group` where we now receive `type.archive.family`. This is skipping a step, because if we updated the API to return `type.archive.group` where ever it now returns `type.archive.family` we would have to deploy that at the same time as this _and_ the equivalent changes on the mobile apps, and given that we'd have to contend with users on outdated versions of the mobile apps, that would still be breaking for people. Instead, prior to updating what the API returns for archive type, we should update all clients to treat `type.archive.family` and `type.archive.group` equivalently (this is what https://permanent.atlassian.net/browse/PER-8901 is for). Once that change has been live for long enough that we expect a sufficient number of mobile users to have it, we can change what the API returns, and then remove the handling for `type.archive.family` from the clients.

crisnicandrei:
@liam-lloyd Thank you! I made it so it now handles both cases. :)

src/app/onboarding/shared/onboarding-screen.ts
```diff
@@ -110,6 +110,21 @@ export const archiveOptions = [
...
+  // {
+  //   value: 'type.archive.group',
+  //   text: "My family's history",
+  //   type: OnboardingTypes.famhist,
+  // },
+  // {
+  //   value: 'type.archive.group',
+  //   text: "A community I'm part of",
+  //   type: OnboardingTypes.community,
+  // },
```
liam-lloyd, line 127: Is this meant to be deleted or uncommented?

src/app/shared/components/new-archive-form/new-archive-form.component.ts
```diff
@@ -14,6 +14,10 @@ const ARCHIVE_TYPES: { text: string, value: ArchiveType }[] = [
     text: 'Person',
     value: 'type.archive.person'
   },
+  // {
+  //   text: 'Group',
+  //   value: 'type.archive.group'
+  // },
```
liam-lloyd, line 20: Same question as above.

cecilia-donnelly:
I saw two "Group" entries in the dropdown when I tried this with https://github.com/PermanentOrg/back-end/pull/583. One sent "family" and one sent "group."

cecilia-donnelly:
It also looks like this didn't change the options for creating an archive, which presumably should also use "Group" instead of "Family."

liam-lloyd (commented):
A couple questions about commented out code, and it looks like the archive type dropdown in Archive Settings > Archive Settings isn't properly populating for family type archives. 
image

crisnicandrei:
Thanks @liam-lloyd @cecilia-donnelly! I have addressed the issue you have pointed.

liam-lloyd:
I'm still seeing Archive Type fail to auto-populate in Archive Settings > Archive Settings for family type archives

crisnicandrei:
@liam-lloyd done

liam-lloyd (approved):

yeslikesolo (approved):

PR #361, opened 2024-02-20
PER-8783-support-txt-files

crisnicandrei:
Display the .txt files in the file viewer using the same approach as the pdf, with the file Url from the FileVOs.

Steps to test:

1.Upload a .txt files
2.Wait for it to process
3.Open it in the viewer.
4.It should display the contents of the file instead of the previous icon

### DIFF
--- a/src/app/file-browser/components/file-viewer/file-viewer.component.spec.ts
+++ b/src/app/file-browser/components/file-viewer/file-viewer.component.spec.ts
@@ -372,6 +372,12 @@ describe('FileViewerComponent', () => {
       expectSantizedUrlToContain(instance, 'original');
     });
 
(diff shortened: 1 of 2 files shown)


meisekimiu (approved):
Code-wise this seems fine although I think from a user experience it would be nicer to have some form of padding on the .txt documents, since most builtin browser PDF viewers have padding built into them. I'll let the experience team decide on that though.

k8lyn6 (approved):
Oh this is so exciting! Thanks for working on this @crisnicandrei ! 

I tested with an existing .txt file in an archive, and uploaded a .txt file, and I was able to view the contents of both. Looks good!

PR #363, opened 2024-02-21
PER-8502 Notify user when upload fails due to role permission

crisnicandrei:
- Add a new message type to the message service to handle the permission error
- Made it possible to pass a html string to the message service, which can be used as an external link (ex:'Read More in case of error')

Steps to test:
1. Navigate to any archive where your role is viewer
2. Try uploading a file by drag and drop
3. It should display the following error message: You do not have permission to upload to this archive. [Read more.](https://desk.zoho.com/portal/permanent/en/kb/articles/roles-for-collaboration-and-sharing)

### DIFF
--- a/src/app/core/services/upload/upload.service.spec.ts
+++ b/src/app/core/services/upload/upload.service.spec.ts
@@ -102,4 +102,20 @@ describe('UploadService', () => {
 
     expect(emittedSessionStatus).toBe(UploadSessionStatus.FileNoBytesError);
   });
+
+  it('should handle NoAccessToUpload correctly', async () => {
+    uploadSessionMock.progress.subscribe((event) => {
+      emittedSessionStatus = event.sessionStatus;
+    });
+
+    const mockEvent: UploadProgressEvent = {
+      item: null,
+      sessionStatus: UploadSessionStatus.NoAccessToUpload,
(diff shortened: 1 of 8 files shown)


meisekimiu (approved):
I'm not sure how I feel about the `externalUrl` parameter taking in a pure HTML string. If possible I think it'd be best to store the actual link href and display text as their own parameters (or in a single data structure) rather than allowing for arbitrary HTML, but the MessageComponent/Service methods already take in a lot of parameters as is. I think this is a good opportunity to see if we should refactor the MessageComponent/Service methods to take in some kind of data structure with all of the options inside instead of having functions with 5+ parameters [(a style choice we have made in other TypeScript codebases here)](https://github.com/PermanentOrg/permanent-sdk/pull/217) but I will ultimately leave that for you to decide.

crisnicandrei:
@meisekimiu I think refactoring is a good idea, but might I suggest we do it in another ticket so we don't make this pr longer than it should be? The MessageComponent is used in quite a few places that's why I am saying this.

k8lyn6 (approved):
Looks good to me! Thanks!

cecilia-donnelly:
@crisnicandrei did you happen to make another ticket for refactoring the Message Service? That could be a good thing to pick up in the next few days if you're looking for tickets while I'm out.

PR #366, opened 2024-03-05
Per 9491 new sign in

crisnicandrei:
This pull request serves as the new log-in/signup/forgotten password pages. This has been checkout out of the new checkbox component because we did not have it, but the design was making use of it.

Some components have slight design changes because when I integrated them into the new design, I realised they were a bit off.


meisekimiu (changes requested):
This looks really good! Most of my feedback comes in the form of some functionality-level accessibility issues. Technically I should have caught some of these a bit sooner in the individual component PRs, but I was a bit more tolerant about some implementation details of these components since they weren't being used in the UI yet. Now that this is their first official usage I have some more notes. Sorry for not testing or noticing all of this earlier.

The main accessibility issue is just that these components are not very keyboard accessible. Here are some of my observations:
1. The checkbox component is not accessible by keyboard at all. Ideally it should be just as functional as a native checkbox. (If I were implementing the component, it would wrap around an actual textbox element so that all native functionality is still intact, even if the Angular component has additional fancy effects. There are other ways of doing this as well, though.)
2. The `secondary` type button component does not have a visible selection outline when using a keyboard, making it hard to tell if it's even focused with the keyboard.
3. The button components cannot be clicked with just the keyboard. This seems to be a conflict with having both a `(click)` event on the button as well as a `[routerLink]`? Maybe?

There are also some more UI-level accessibility issues that the experience team has spotted (things like color contrast, etc), although I think we can ignore those until the designs themselves are updated. (...Maybe we could underline text links though, that's not going to need real design work. Up to you if you want to implement it or wait for specific changes to be made in the design.)

One final question I have: Does this ticket include mobile designs at all or just the main tablet design? This implementation seems to be at least somewhat responsive, but there are a few weird spacing issues in the mobile view. Ideally I'd like to fix those even if mobile isn't a priority.

meisekimiu:
*Bonus tip I'm leaving in an extra comment!* This is a big pull request that's probably going to have some back and forth since it's a big new UI change and I'm sure even after code review passes there will be additional QA comments. Unfortunately, that means this code is not going to be integrated into main until many people at Permanent have signed off on the releasability of this new screen.

In the future for big redesigns like this, we should try experimenting with a strategy that allows us to merge code without releasing that new screen to users yet. What I specifically mean by this is that instead of replacing the old components, we can create new components that make up this new UI and for a time both the old UI and new UI coexist in the codebase. In production the new UI is hidden away from the user, but the code is merged. Once we decide the new screen is actually releasable, we fully replace the old screen with the new one.

This isn't advisable in *all* situations, but I think it could speed up merging PRs on some future big tickets, which through this technique can hopefully be broken into smaller more iterative tickets that can be reviewed faster.

crisnicandrei:
@meisekimiu could you please comment here the spacing issues that you. have encountered? I might overlook something, so having someone else look could help me. Otherwise all your comments are valid and I will take care of them. Also thank you for the tip! Will keep in mind when working on bigger tasks!

meisekimiu:
On mobile view, the margin between the main heading and the actual text inputs is a bit too big. It requires me to actually scroll to access the inputs to log in:


Also on this screen there seems to be no padding on the body text which is inconsistent with other screens (like the terms) and makes everything a bit harder to read:

crisnicandrei:
@meisekimiu Thanks! Great inputs! I left the margin that big to be compliant with the desktop view, but I think you are right. I have pushed the changes.

src/app/component-library/components/checkbox/checkbox.component.html
```diff
@@ -6,6 +6,11 @@
   [class.checkbox-container-primary]="variant === 'primary'"
   [class.checkbox-container-secondary]="variant === 'secondary'"
   (click)="toggleCheck()"
+  (keydown)="handleKeyDown($event)"
+  role="checkbox"
+  [attr.tabindex]="disabled ? -1 : 0"
+  [attr.aria-checked]="isChecked"
+  [attr.aria-disabled]="disabled"
```
meisekimiu, line 13: If we use the `checkbox` role attribute here, browsers will mark the child elements as presentational (see: [ARIA checkbox role descendants](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role#all_descendants_are_presentational)). Since the label is a child element, we should really only apply this role on the actual specific element that's representing the checkbox.

src/app/component-library/components/checkbox/checkbox.component.ts
```diff
@@ -20,7 +20,16 @@ export class CheckboxComponent {
   toggleCheck() {
     if (!this.disabled) {
       this.isChecked = !this.isChecked;
-      this.isCheckedChange.emit(this.value);
+      this.isCheckedChange.emit(this.isChecked);
+    }
+  }
+
+  handleKeyDown(event: KeyboardEvent) {
```
meisekimiu, line 27: Suggestion: While this does make keyboard functionality testable in automated tests, I do want to mention that if we use a [native checkbox element](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role#best_practices) that is simply restyled (or hidden with CSS), we do not have to write the checkbox logic ourselves and we don't have to write tests.

meisekimiu (commented):
Some additional comments from a second round a review.

The checkboxes appear to be a bit off center in my default zoom level in Chrome; this seems to happen around the 1340px screen width:


This seems to be a small bug that happens only at specific screen widths.

crisnicandrei:
@meisekimiu this is how it looks for me at 1340 
Screenshot 2024-03-28 at 11 23 46

crisnicandrei:
@meisekimiu @cecilia-donnelly I have addressed the comments, sorry forgot to respond here

cecilia-donnelly:
Steps for this one:
- @meisekimiu review
- QA
- Deploy (which will change the look of the signin page only)

src/app/auth/components/auth/auth.component.html
```diff
@@ -1,12 +1,28 @@
...
+        gallery and archive spotlights. Explore our public archive spotlights
+        and, when you’re ready, publish or share your own public archive!
+      </p>
+      <a
+        class="link"
+        href="https://permanent.org/blog/tag/public-archive-spotlight/"
+        >Start Exploring Now.</a
+      >
+    </div>
+    <img src="assets/img/sign_in_image.png" />
```
meisekimiu, line 16: This should probably have an `alt` attribute. Either with a blank alt if we decide this is a purely decorative image, or proper alt text if we want to preserve that.

src/app/auth/components/auth/auth.component.scss
```diff
@@ -0,0 +1,97 @@
...
+.image {
+  position: relative;
+  width: 80%;
+  height: 100%;
+
+  & > img {
+    width: 100%;
+    height: 100%;
+    object-fit: contain;
+  }
```
meisekimiu, line 54: Should we have a more specific maximum/minimum height for this image? I feel like it has potential to look weird at different zoom levels, on big monitors, or close to the mobile breakpoint.

This is more a question for the QA process though.

meisekimiu (commented):
I left some comments on here. I think this will probably require a round of (blocking) QA, and for a PR this big I think I would like to re-review code after QA comments so I'm not going to explicitly approve it until after that, if that's okay.

src/app/auth/components/auth/auth.component.html
crisnicandrei: @k8lyn6 is there a specific alt text you might consider using?

emlesienk (changes requested):
**Design:** I think the "At Permanent" block needs some shadow. It looks too flat right now. If you need me to make any changes to the background image, let me know on slack because I have the file in canva. Do we want to round out the corners on the image so it flows better with the rest of the page? It looks too boxy with the squared-off corners. It might be worth me making a new graphic altogether - maybe a grid of the different archive profile pictures so we can remove the purple text boxes on the image. Can someone let me know the ideal size for this image?

Once we figure that out, I can get the alt-text to you. For the current image, the alt-text could be "Photo with a bust of a woman in striking hat and glasses. Public Archive Spotlight: The Unknown Faces Archive."

Also, we'll be making category / tag improvements to the blog soon so that link to the blog is likely to be invalid pretty soon. I'll get that new link once we have it.

**Functionality:** It lets me log in as expected, 2FA is working.

k8lyn6:
@crisnicandrei Can we try out this image instead?
Sign in page

crisnicandrei:
@emlesienk that would be great if you could update the design. Thanks!

crisnicandrei:
@k8lyn6 i have changed the picture. Let me know what you think!

k8lyn6 (changes requested):
Looking good! I've left a few changes below, but want to say -- many of these changes are because we didn't have a final design when this was first given to you. I apologize for that! We've got more of a design to development pipeline, so hopefully this is less of an issue in the future. 

- It looks like the spacing for the right side components (the icon, title, fields, etc) aren't in the correct location on the page. They should all be centered in the middle of the page. There's a big gap between the "Sign in" and the email address field.
- The order of the right side components got adjusted in the final design, it looks like. Specifically, the forgot password/sign in buttons are a little different. I just noticed that the final design does not have a "Remember me" checkbox -- I think that should stay directly below the Password field, but the rest of the components should look the same as the design.
- The text box overlay got changed to be less transparent, to make it more accessible. 
- The text box overlay in the design is also more wide than tall. Is there a way for us to adjust that?
- I think we are going to swap out the photo again (I'm so sorry!) I'll get it to you next week.  

I have a question for @meisekimiu but I'll leave that for next week as well, so this might not be a great ticket to jump on first thing Monday morning because I'll likely request a few more changes (at the very least, the photo swap).

Thanks for all your hard work on this Andrei! I apologize that the process wasn't entirely smooth for this.

k8lyn6:
Sign in page:

- Center text on Sign in button
- Icon on sign in button is an arrow
- Icon on Register button is the person outline with a plus

Sign up page:
- Center text and icon on Register button
- Sign in button should say "Back to sign in" with the arrow

Photo and language:
- The text box overlaid on the photo should be less transparent ([see design](https://www.figma.com/design/lLHzIJcmkwkvdl3ipcjyKs/%5BApp%5D-Permanent?node-id=1973-33744&t=AAt12JQZn2ZUQCOG-0))
- The text box should be wider
- New photo attached
- The text should not be italicized
- Is this the correct font?
- Alt text for photo:

> Black and white photo of a child with a bicycle stopped in front of a black milk truck with a man standing by the door. The milk truck reads "Haadsma's Dairy." The photo appears to be from the 1950s.

- Text for photo:

> At Permanent, we celebrate our members’ hard work through our public gallery and archive spotlights. Explore our gallery and, when you’re ready, publish or share your own public archive!
> 
> [Start Exploring Now.](https://permanent.org/gallery)
> 
> Haadsma Dairy Truck | The Haadsma Dairy Archive [this text should be the same size as the smallest text in the design]

Sign in page (2)

k8lyn6 (changes requested):
Hey Andrei, I think we're getting really close on this! A few things:

Photo and text box:

- the text should be blue `#131B4A ` according to the design
- Can you ensure that the text box is responsive when the zoom or browser window is resized?
- The text box also appears to be much larger than it is in the design (it looks larger on top of the photo than it shows in the design). Ideally, we want to be able to see what's happening in the image behind the text box. I know the design has the text box placed towards the left side and centered vertically. Can we actually move it towards the lower left hand corner so you can see the milk truck behind it? 
- Resize the last line of text (the file name and archive name) to be the same size as the smallest text in the design (the part that says "Shenzhen Shi, China")

Terms link:
- The terms link on the register page should go to this link instead of a new screen: https://www.permanent.org/terms/

crisnicandrei:
@k8lyn6 thanks! Will do!

k8lyn6 (changes requested):
I think we're really really close to finalized here! A few small things (and I think @meisekimiu had some feedback she was going to relay as well):

- Can you capitalize "Public Gallery" in the text overlay? Small tweak I forgot in the last review.
- When the browser is resized smaller, the text box extends past the photo. Is there a way for that to be avoided? I included a screenshot below.
Screen Shot 2024-08-07 at 11 19 03 AM
- When I click on the Terms link, some weird effect is happening over the link (screenshot below). Is that on my end?
Screen Shot 2024-08-07 at 11 21 17 AM

Thanks Andrei!

meisekimiu (changes requested):
I have some CSS feedback based on some of the experience team's comments on the textbox size. These changes have already been approved by @k8lyn6, so I recommend implementing them. Note that these changes only deal with the size of the text box, and once it is resized we might want to reposition it as well.

In addition, I added a comment on an HTML change as well and would like to bring up some other things for discussion:
- A vertical scrollbar is visible on the auth pages when I am at a normal desktop resolution. In theory this page shouldn't need a scrollbar until you reach an absurd viewport size, right?
- In the full desktop view, the big image and textbox take up most of the viewport while the sign in/sign up controls are small and off to the side. This is fine on a full screen width, but as you approach the mobile breakpoint the width ratios between these two sections of the layout continue to remain constant. I feel like the sign in/sign up controls should have some minimum width that they stay while the image just gets smaller as you approach the breakpoint, but that's just my opinion and it's not explicitly specified in the designs as far as I know.

crisnicandrei:
@k8lyn6 If i resize the textbox with the image, the text will be unreadabla. @meisekimiu do you have any suggestions? Maybe i am unaware of something

meisekimiu:
> @k8lyn6 If i resize the textbox with the image, the text will be unreadabla. @meisekimiu do you have any suggestions? Maybe i am unaware of something

I don't think the text size should change at all. What's more important is to eliminate empty space in the textbox (which seems to be done) and to make sure that at smaller widths the position of the textbox doesn't looking weird. At full width, having the textbox in the bottom corner makes sense, but when there's not much room it probably should just fit in the middle or something like that so it isn't jutting out weirdly.

crisnicandrei:
@k8lyn6 @meisekimiu i made some changes to the text box. Let me know how you feel about it now!

k8lyn6 (changes requested):
Ahhh so close! A few things:

- Text box: move down slightly, so more of the two people are showing behind it
- Text box: add a line break before the link, so there's more space between the paragraph and the link
- Text box: I agree with what @meisekimiu said about not making the text smaller when resizing if possible.
- Alt text: is there a way for the alt text to show up in a different color (white?) other than black? You can't really read it because of the dark background.
- Photo: when I log in and then log back out, I'm taken back to the login page, but the photo does not show up and instead the alt text shows up until I refresh the page. This happened multiple times -- I'm not sure why. Can we investigate and fix this?

crisnicandrei:
@k8lyn6 I can't change the alt text colour, but we could use a fallback text if the image does not load?. Also i could not reproduce the last bug with the logout ..

meisekimiu:
> @k8lyn6 I can't change the alt text colour, but we could use a fallback text if the image does not load?. Also i could not reproduce the last bug with the logout ..

The alt text color can be changed just with the CSS `color` attribute applied to the `` tag or parent container(s).

crisnicandrei:
@meisekimiu thanks! Very helpful! I was not aware and what I found mentioned only a fallback. I made the change now.

k8lyn6 (changes requested):
Looking good! That bug I found isn't happening to me anymore so I'm not sure what the problem was.

When you're done making these changes, would you mind tagging @yeslikesolo since I will be out? I added her as a reviewer to this PR. I think we're nearly there! 

Changes:
- Text box: Can you move it down a little bit more, so the bottom of the text box is closer to the edge of the photo?  CC @meisekimiu in case you can add info here that might help make this adjustment.
- Terms link: when the browser gets resized smaller, the line breaks on the "I agree to the terms" link are weird. It should wrap like normal paragraph text instead of wrapping the "I agree to" and the link separately. Photo attached
Screen Shot 2024-08-09 at 10 04 48 AM
- Terms link: can you make the "o" in ".org" lowercase?
- Spacing on the right side: ideally, all the information on the right side should fit on one page on a normal sized screen without scrolling. Can we maybe reduce the amount of space under "Create your new account" to get it all on one page without a scroll? If not, I think changing the link text to just be "Terms of Service" instead of "Permanent.org Terms of Service" will keep it all on one line and make the spacing a little better. @meisekimiu  if you have any ideas here, feel free to share. Screenshot below.
Screenshot 2024-08-09 at 10 11 23 AM

crisnicandrei:
@k8lyn6 pushed the changes! What resolution are you testing on?

k8lyn6:
@crisnicandrei Han's testing on a screen resolution of 3024 x 1964 pixels. That's the screenshot I sent. I have a large monitor so I've been having her take a quick look on an average sized screen to make sure it looks good there too.

k8lyn6 (changes requested):
- Right side spacing: It still scrolls vertically on a smaller screen. Let's drop the "Permanent.org" from the ToS link text to see if that helps.
- Horizontal scrolling: As you resize the screen smaller, at some point, the fields for entering your email and password start to go off the page on the right and a horizontal scroll bar appears. Is there a way to avoid both of those things? @meisekimiu if you have any thoughts here, feel free to share.
- Text box: Now, when the browser screen is very large, the text box is spilling off the left side of the photo. Is there a way to fix that?
- Sign Up Text: Can we make "account" in the heading on the sign up page lowercase?

Almost there!

crisnicandrei:
@k8lyn6 pushed the changes! I think I also fixed the horizontal scrollbar! Let me know!

k8lyn6 (changes requested):
The horizontal scrolling is gone now, thanks! A few things:

- It looks like the Permanent logo is getting cut off at the top from some blue margin. Is there a way to fix that? @meisekimiu mentioned bumping up the minimum width.
Screenshot 2024-08-09 at 4 29 34 PM
- Can we move the text box down the photo a little, so it's closer to the bottom?
- The text box still hangs over on the left side of the image a little when expanding the browser to a large size. Is it possible to address that?

Thanks for all your hard work Andrei! @yeslikesolo will work with you next week for any more changes while I'm out.

crisnicandrei:
@k8lyn6 thank you! I will address all your comments! Enjoy your time away!😃

yeslikesolo (approved):

k8lyn6 (changes requested):
Small small comment and I think we're done!

At smaller screen sizes, the photo is flush to the left side of the screen. In the design, there's supposed to be a little bit of space between the edge of the browser and the photo, so the background shows a little. Is it possible to adjust this?

Great work, @crisnicandrei !

crisnicandrei:
@k8lyn6 margin added

yeslikesolo:
Hi @crisnicandrei ! Kaitlyn has requested to change the "Register" button to say "Sign Up". Could you make that update please? Thanks!

crisnicandrei:
@yeslikesolo done

k8lyn6 (changes requested):
I think this looks really good design wise. Thanks for making those changes Andrei.

I did one last test and noticed that if I put in an email address that isn't registered in the database, I get an error. The user should be able to submit the email address, and it just won't send an email because it's not in the database.

I think we can ship this when the above bug is fixed. I also created a [new ticket](https://permanent.atlassian.net/browse/PER-9657) to add the email opt in checkbox to the sign up page (this was an update to the design after you received it). Could you work on that ASAP after this PR is merged? If you have any questions about the design, let Tibi know!

crisnicandrei:
@k8lyn6 can you explain the email part again please? I don't really get it

k8lyn6:
@crisnicandrei absolutely, sorry for the confusion! Realized I left out a crucial part (that this is the forgot password page)!

Current behavior:
1. On the forgot password page, enter an email address that is not in the database (i.e. one that does not have a Permanent account)
2. Observe that an error is thrown.

Expected behavior:
1. Enter an email address not in the database
2. The form submits normally with no errors. No "Forgot password" email should be sent because the email is not in the database.

k8lyn6 (approved):
Perfect!

PR #367, opened 2024-03-05
Implement folder caching in PermanentFilesystem

meisekimiu:
When calling the "navigate" method from PermanentFilesystem, save the result in a cache and try to fetch from cache when looking up the same folder later.

This cache is not meant to actually reduce the number of network calls, but rather to provide a somewhat accurate temporary cached value while the browser is waiting for the API to respond back with the true value.

As a result, the code is designed in such a way that any objects returned from the `getFolder` method may automatically update after the function is called. Question: Is this good? I like how convenient this is since it could enable a "live update in UI" when the fetched object updates, but I think a lot of our code may have been written to assume that fetched objects won't just suddenly change once they're fetched.

### DIFF
--- a/src/app/filesystem/permanent-filesystem.ts
+++ b/src/app/filesystem/permanent-filesystem.ts
@@ -5,16 +5,38 @@ import {
   FolderIdentifier,
   RecordIdentifier,
 } from './types/filesystem-identifier';
+import { FolderCache } from './folder-cache';
 
 export class PermanentFilesystem {
-  constructor(private api: FilesystemApi) {}
+  private folderCache: FolderCache;
+
+  constructor(private api: FilesystemApi, private cache?: FolderCache) {
+    if (cache) {
+      this.folderCache = cache;
+    } else {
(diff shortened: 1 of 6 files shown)


src/app/filesystem/permanent-filesystem.ts
```diff
@@ -5,16 +5,38 @@ import {
...
     return await this.api.getRoot(archive);
   }
 
   public async getFolder(folder: FolderIdentifier): Promise<FolderVO> {
-    return await this.api.navigate(folder);
+    const cachedFolder = this.folderCache.getFolder(folder);
+    if (cachedFolder) {
+      setTimeout(() => {
+        this.fetchFolderFromApi(folder);
+      }, 0);
```
liam-lloyd, line 30: How does this change the value returned by `getFolder` from the cached value returned originally to the result of the API call? (I can see that it does, because the test passes, but I don't understand it).

liam-lloyd (approved):
Looks good, only comment is an effort to improve my own understanding

PR #370, opened 2024-03-05
Fix failing MainComponent test

meisekimiu:
This test fails sometimes because the test module is provided with an undefined MockDragService. Initialize this mock before setting up providers so that this test passes regardless of which order tests are run in.

**Steps to test:**
1. Start the test runner with `npm run test`
2. Open the Karma debugger in your browser by navigating to http://localhost:9876/debug.html
3. Set the random seed to 83937 by adding `?seed=83937` at the end of the URL
4. On the current mainline branch, a test should fail in `MainComponent`
5. On this branch the test should pass.


liam-lloyd (approved):
I couldn't get the test to fail on `main`, but I do see it passing on this branch

PR #372, opened 2024-03-14
Per 9494 convert mixpanel calls

crisnicandrei:
This pull request replaces the previous mixpanel tracking from the web-app with the back-end tracking using api's

Steps to test:
1. Open the network tab
2. Complete any action which should be tracked by mixpanel
3. In the network tab there should be an event with the name **event** which should send a good result


meisekimiu (changes requested):
Instead of creating a new method in the `ApiService` through a new `MixpanelRepo`, let's instead edit the existing `MixpanelService` to make these calls from its `track` method. Any new tracking calls can be replaced with a call to a `track` method instead. RIght now the `MixpanelService` has basically been gutted of functionality and now all calls to an external endpoint are scattered across the application. The intention behind the `MixpanelService` is to centralize the code that tracks things so that there is only one place to change it in situations such as this.

meisekimiu:
Also here's a thought about the project architecture I've had since #277 has been merged. This is not a blocking change request, just a suggestion you can feel free to try or not.

In hindsight I regret the design of the `MixpanelService` a little bit. I think it should have been an `AnalyticsService` that was more abstract and would act as a [Subject](https://en.wikipedia.org/wiki/Observer_pattern) that dispatched calls to Observers such as a `MixpanelService`. This way, no systems in the web-app would explicitly know they were talking to Mixpanel in particular. If we ever had to change it out for something else (Google Analytics, etc) no calling code would change. This would also make it easier to disable analytics calls in unit tests (the `AnalyticsService` would just have no registered observers during tests).

Changing this here would be nice since the `/event` endpoint has since been designed to have that abstraction in mind (by using the `analytics` property instead of `mixpanel`).

crisnicandrei:
@meisekimiu thank you for your input! I will try implementing your suggestion instead of only moving the code from the repo to the service.

src/app/shared/services/analytics/analytics.service.ts
```diff
@@ -0,0 +1,37 @@
...
+
+@Injectable({
+  providedIn: 'root',
+})
+export class AnalyticsService {
+  private observers: AnalyticsObserver[] = [];
+
+  constructor(private mixpanel: MixpanelService) {
+    this.addObserver(mixpanel);
+  }
```
meisekimiu, line 19: Thanks for using the observer architecture here! One of the benefits we get from using an observer architecture like this is that it allows us to invert dependencies and set up boundaries in our architecture. So in theory, the `AnalyricsService` should not depend on the `MixpanelService` or really know about its existence at all. This way in the future if we add new analytics observers or remove mixpanel, we don't have to update the AnalyticsService file. Of course, ultimately we do need to actually add the MixpanelService to the AnalyticsService in some way, but it would create a nicer architectural boundary if this configuration was outside of the AnalyticsService class itself.

In testing, I found that I could do this in the `AppModule`'s constructor and it worked perfectly. There might be other places to do that, though.

meisekimiu (approved):
Thanks for making the requested changes! I left some notes on some ways to improve the Observer architecture you built. The only other thing I might suggest is that I think we can remove mixpanel as a dependency and the mixpanel token as a secret now that we're no longer using them from the web-app. Otherwise, I think this looks good and I can see the `event` endpoint being called.

crisnicandrei:
@meisekimiu thank you! I have added the service to the app module and adding mixpanel there too

k8lyn6 (approved):
I think this is all working how we intended it to!

PR #373, opened 2024-03-15
PER-9482: left side archive info not matching the file list items displayed

crisnicandrei:
Get the account saved in the storage then use the keepLoggedIn flag from the storage account instead of the account service variable, because the variable was refreshing on each refresh, making keepLoggedIn null

### DIFF
--- a/src/app/shared/services/account/account.service.ts
+++ b/src/app/shared/services/account/account.service.ts
@@ -207,7 +207,7 @@ export class AccountService {
           const newArchive = response.getArchiveVO();
           const newAccount = response.getAccountVO();
           const account = this.getStorage('account');
-          newAccount.keepLoggedIn = account.keepLoggedIn;
+          newAccount.keepLoggedIn = account?.keepLoggedIn;
           this.account.update(newAccount);
           this.archive.update(newArchive);
           this.setStorage(newAccount.keepLoggedIn, ARCHIVE_KEY, this.archive);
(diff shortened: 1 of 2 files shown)


meisekimiu (commented):
I can still recreate the bug by manually adjusting the `archive` field in developer tools. Does this fix prevent the bug from triggering outside of that? I'm not entirely sure how to recreate that. Can you explain how this change fixes the bug?

crisnicandrei:
@meisekimiu the issue that I found was that on page refresh the archive refresh was being saved in the session storage instead of the local storage. When checking if there is an archive either in the local storage or session storage, the session storage always comes first (this is how I implemented it a while ago), so that's why I retrieve the account from to storage, so we save correctly the archive on refresh

Wouldn't overwriting the storage using devtools always mess it up?

meisekimiu:
Can you add steps to reproduce the specific bug being fixed here?

crisnicandrei:
@meisekimiu 
-Have an account with 2 archives
- Open the local storage
- Switch the archive
- Refresh page
- Notice that the archive has been added to the session storage upon refresh

This is what this fixes

meisekimiu (approved):
I can see how this code addresses an issue and approve the code changes, although I'm a bit apprehensive to close the actual bug ticket we have open for this since I'm not sure that this change completely addresses the bug. However I think the experience team might have better judgement for that, so I'll pass it on to them.

k8lyn6 (approved):
I can't reproduce the bug reliably, but I'm going to go ahead and approve this PR. The change hasn't affected any major functionality negatively, and we might just have to wait and see if this addresses the bug. I think we should keep the ticket open as @meisekimiu said, until we know for sure whether or not the bug is fixed.

PR #375, opened 2024-03-20
PER-9503- removed facebook integration

crisnicandrei:
Removed all the facebook related code from the web-app. I thought about refactoring the app, but I think it makes more sense to keep the already implemented switch statements for the future if there will be more connectors. Removing them now will only make us have to re-add them in the future.

### DIFF
--- a/src/app/apps/components/connector/connector.component.html
+++ b/src/app/apps/components/connector/connector.component.html
@@ -12,11 +12,7 @@
     <div class="name">
       {{ connector.type | prConstants }}
     </div>
-    <div class="blurb" [ngSwitch]="connector.type">
-      <ng-container *ngSwitchCase="'type.connector.familysearch'"
-        >Connect your family memories</ng-container
-      >
-    </div>
+    <div class="blurb">Connect your family memories</div>
   </div>
 </div>
 <div class="connector-actions">
@@ -34,7 +30,6 @@
     [hidden]="!folder"
     (click)="goToFolder()"
     [disabled]="waiting"
-    *ngIf="connector.type.includes('familysearch')"
   >
     View Memories
   </button>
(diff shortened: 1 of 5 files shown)


meisekimiu (changes requested):
Thank you for picking this one up. Normally I wouldn't block on refactoring changes, but this is a rare case where I feel strongly enough that we should refactor things more. The ConnectorComponent in its previous form violated a few design principles; specifically I saw it violating both the [Single Responsibility Principle](https://en.wikipedia.org/wiki/Single_responsibility_principle) and the [Open-Closed Prinicple](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle). Removing Facebook from the component does inadvertently take care of the Single Responsiblity Priniciple, but the Open-Closed Principle is still violated because we're expecting to add new connector behavior using modification instead of extension. Ideally if we added a new app connector we'd want to make a new component and not put all the code in the same ConnectorComponent that handles FamilySearch.

While it may seem nice to leave hooks in so we don't have to re-add them later, I think we should be following the advice of ["You Aren't Going to Need It" ](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it). We should try to keep designs as simple as they can be in the present context.

I don't think we need to rename the component or anything right now (that can be done whenever a new connector actually gets made), but can we remove all of the switch/if statements asking for what kind of connector the component is and just display FamilySearch by default?

meisekimiu (approved):
Thanks for the changes!

omnignorant:
I am assuming that the bulk of the work here is in the code architecture and not in the UI, but as far as I can tell, this is what I would have expected. When I navigate to App Files, there is no option to connect Facebook. However, I can still access the Facebook photos I downloaded a long time ago in the Facebook subfolder, accessible from the LHS navigation. I think this is good to go!

omnignorant (approved):
see my previous comment in the conversation thread. looks AOK to me.

PR #379, opened 2024-04-01
Remove unused LoadingArchiveComponent

meisekimiu:
The LoadingArchiveComponent was created to be an interstitial component displayed when switching archives in commit 76883bd. However, as time has gone on, the `/app/switching` path became unused. Since this specific LoadingArchiveComponent was only used in this one route, the component itself became unused. Presumably it stayed around because it *sounds* important!

This component has problematic tests. They are (relatively) slow since they wait 2.5 seconds just for one test, and also sometimes fail due to random asynchronous errors. While we could optimize the component to fix these errors, since the component is unused we should just delete it and the problematic tests.

Steps to test:
1. Run tests/build/etc.
2. Verify that nothing breaks in the web-app when loading things, especially when switching archives.

Resolves PER-9520.


liam-lloyd (approved):

meisekimiu:
Requesting QA on this one just to be safe. In theory there are no user-facing changes but perhaps this breaks something I didn't see? For QA one should mainly check that you can still switch archives.

k8lyn6 (approved):
Switching and creating archives on both new and old accounts seems to work just fine!

PR #382, opened 2024-04-05
Refactoring the messge service

crisnicandrei:
- Made sure that the service uses now objects as args so we don't clutter the function calls. This way we can easy use only the arguments we need instead of using falsy values just to call a function
- Added tests to the service's spec file.


meisekimiu (approved):
Thank you for this change! I appreciate this style of refactoring.

Note for QA, I think there should just be some basic tests to verify that anything that usually has a message pop up at the top of the screen "Logged in successfully" / "There was an error" / etc. still works. This doesn't really need to be too comprehensive since any actual issues that appear would be caught by test/build failures.

yeslikesolo (approved):

Issue #385, opened 2024-04-12
Do we want to use lodash?


meisekimiu:
Lodash used to be a quite useful library, but advancements in JavaScript's standard library features have made it less and less useful. I would argue that it makes some code hard to understand (for example, do you know what [this function does to an array of ShareVOs just by reading it?](https://github.com/PermanentOrg/web-app/blob/main/src/app/models/share-vo.ts#L9) I wasn't able to without reading through some Lodash documentation.) since it relies on knowing how to use Lodash. Do we want to eventually get rid of it and replace it with more modern and native solutions?

meisekimiu:
At least for current "v2" development, lodash is actually core to some functionality. Maybe it can be avoided in "v3" development, but for now it's not going anywhere.

PR #388, opened 2024-04-23
Integrate folder caching into web-app

meisekimiu:
This PR actually integrates the previously hidden PermanentFilesystem class into the application. This is done by creating the `PermanentFilesystemService`, a wrapper service that allows this functionality to be imported into other modules across the application. The service also provides the proper implementation of the `FilesystemApi` interface, which wraps around the `ApiService` and processes the `FolderResponse` automatically.

Steps to test:
1. Open the web-app and make sure you have multiple folders with their own contents
2. Navigate into each folder once
3. Without refreshing, open the developer console and enable network throttling
4. Navigate into the folders you went into before. They should load immediately from the folder cache even if the `navigateLean` response has not finished.
5. You should not notice any extraneous `navigate` API calls, or large shifts in UI after the API calls finish.
6. Make sure that public / shared folders still work as expected as well

### DIFF
--- a/src/app/core/core.module.ts
+++ b/src/app/core/core.module.ts
@@ -33,6 +33,7 @@ import { PledgeModule } from '../pledge/pledge.module';
 import { AnnouncementModule } from '../announcement/announcement.module';
 import { ManageMetadataModule } from '../archive-settings/manage-metadata/manage-metadata.module';
 import { DirectiveModule } from '../directive/directive.module';
+import { FilesystemModule } from '../filesystem/filesystem.module';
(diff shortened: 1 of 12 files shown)


meisekimiu:
Random note about some code changes. I opened #385 pretty recently, but I actually ended up using lodash here since it was available and I wanted to check deep equality between FolderVOs to prevent rerenders. So maybe we should just close that issue for now?

liam-lloyd (approved):

k8lyn6 (approved):
Followed your instructions @meisekimiu and everything looked good to me! And share and public folders are working as expected too. Thanks!

PR #390, opened 2024-04-24
Per 9191 front end 2fa

crisnicandrei:
VERY IMPORTANT: This is only the UI part, using placeholder data until the API is done, to test this, navigate to https://www.dev.permanent.org/app/(private//dialog:security)?display2fa=dev (if you are testing on dev, if you are on staging use the required url with the correct variable at the end, otherwise the component will not be visible. 

How to use: There is already a placeholder method added just as an example, if you want to add another 1 click the Turn on button, fill the input and then click send code. For now any code with the length of 4 will work, but with the api this will be changed. 

If you want to delete a method, click the delete button next to any method, the again, add a code, any code, then click delete. This too will be changed once the api is complete


src/app/core/components/two-factor-auth/two-factor-auth.component.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+    shallow = new Shallow(TwoFactorAuthComponent, CoreModule).mock(
+      MessageService,
+      {
+        showError: () => {
+          messageShown = true;
+        },
+      }
+    );
+
+    it('should create', async () => {
```
meisekimiu, line 24: It seems like you didn't close the `beforeEach` block properly. Can you fix this?

cecilia-donnelly:
@meisekimiu , @crisnicandrei, does this need QA or are we assuming that it'll be QA'd once it has a backend connection? (Really asking if it can be merged now that it's approved.)

cecilia-donnelly:
If it needs QA then please tag @k8lyn6 for a review.

meisekimiu:
I think it could go without QA since it isn't user facing and a bit hard to test at the moment. But doing some QA now would provide feedback on the UI more immediately instead of having to wait for the backend work to be finished.

k8lyn6:
@meisekimiu Happy to do QA now before the backend work is done for the UI (though I probably won't get to it until Monday or Tuesday). Up to you.

cecilia-donnelly:
@k8lyn6 I'd like to get this merged so we can get it off our sprint list - is it on your list for QA this week?

k8lyn6:
Hey @cecilia-donnelly -- my understanding was that the issues on Dev from last week prevents me from successfully deploying this to dev until we've done our deploy this week and the fix goes out. Is that correct? If so, I have it on my list to QA after the deployment this week. If I'm misunderstanding, then I can prioritize this ASAP.

meisekimiu:
You should be able to deploy this branch without any problems, @k8lyn6

k8lyn6:
@meisekimiu Ahh, okay, sorry about that! I'll work on that this afternoon.

k8lyn6 (changes requested):
This is looking good so far! A change and a few comments/questions:

- "At this time, we can only accept North American cell phone numbers for sms." Can we capitalize all instances of "SMS"?
- I was never able to add a SMS method because it wouldn't let me click the "Send code" button. Is this expected?
- (this might be a @cecilia-donnelly question) For the deletion of a method, do we have to confirm deletion with a code? If we do, I guess it does make sense to have the user put in what email address/phone number they'd like to use in case they've lost access to the previous one, but it doesn't seem like a necessary step since users have to authenticate to get to this step (aka log in), unless FusionAuth makes us do it or it is best practice. Just wanted to check since this wasn't in the original design! I can provide language there if we need it.

crisnicandrei:
@k8lyn6 it allowed me to add a sms method, but perhaps i got the validation wrong for us numbers; this is the value i have tested with: 12341242142

k8lyn6:
> @k8lyn6 it allowed me to add a sms method, but perhaps i got the validation wrong for us numbers; this is the value i have tested with: 12341242142

@crisnicandrei The format for US numbers is generally (XXX)-XXX-XXXX (ten numbers) unless we need the country code in front for FusionAuth, in which case we should change the help text for the field to include the country code.

crisnicandrei:
@k8lyn6 i see, thanks!

k8lyn6 (changes requested):
The phone number works now! Thanks for changing that validation.

One small thing I just noticed -- sorry! The icon in the message that appears saying the expiration time is broken. I put a screenshot below. I think everything else is working/looking good!
Screen Shot 2024-05-31 at 10 53 52 AM

crisnicandrei:
@k8lyn6 
image
that is really strange, it works for me. I will try to reproduce it. @meisekimiu was the image working for you?

cecilia-donnelly (changes requested):
Thank you for putting this together, @crisnicandrei ! I looked at it to answer @k8lyn6 's question and found a few things:
1. To answer Kaitlyn's question, it is necessary to send a code on deletion. It's great to auto-fill that. I don't think the field with email address or phone number should be editable, though.
2. A phone number should be labeled "Phone number" instead of email address:
Screenshot 2024-06-05 at 10-54-54 Account Permanent org
3. When I click delete, the input for the 4-digit code appears before I click "send code." This should show up after "Send code."
4. Similarly, when I click "Add another method" I get the input for the phone number *and* the input for the code. This might be because I have added and deleted several methods in a row, so the code is confused, but ideally in both cases the input for the code would appear after the member clicks the "Send code" button.

I know the design was missing a bit of detail, so thanks for working through this with us!

crisnicandrei:
@cecilia-donnelly Thanks for pointing that out! I have made the changes requested :)

cecilia-donnelly:
I am confused about all the new commits - looks like you need to rebase instead of merging? I would do some git management @crisnicandrei to make sure you're getting the intended result.

crisnicandrei:
@cecilia-donnelly i think i managed to fix this

cecilia-donnelly (approved):
You addressed all my comments, thank you!

@k8lyn6 I think this is ready for you to look at again.

k8lyn6 (changes requested):
This is looking great! The only thing I'm still seeing is the icon is broken for me on Chrome. 

Also, I have language for the placeholder text below:
"To disable two-step verification, click the button below to send a one-time use code to your device."

I also have a language update (I'm so sorry!). Can we change all instances of "two-factor authentication" or "2FA" to "two-step verification"?

crisnicandrei:
@cecilia-donnelly @meisekimiu @k8lyn6 I have added the api calls. The icon was still working fine for me :(

crisnicandrei:
@cecilia-donnelly @k8lyn6 where does the "Ths code expires in x amount of time come from"?

meisekimiu (commented):
Does this pull request now encompass both PER-9191 (the 2FA management UI) and [PER-9627](https://permanent.atlassian.net/browse/PER-9627) (the code for calling the 2FA API)? I think in the future we should make sure that tickets like this all get their own separate pull requests so they can be reviewed, tested, and accepted independently of each other. As it stands, the API code is part of the same pull request as the UI which means it is blocked from integration until the UI code is fully accepted by QA.

crisnicandrei:
@k8lyn6 this now has the api calls merged into!

emlesienk (commented):
Was able to set up both SMS Text and Email 2FA. After logging out, it prompted me to enter my new code and let me in once I entered it. Seems to be working as expected.

emlesienk (approved):

crisnicandrei:
@cecilia-donnelly @k8lyn6 can you mark the changes requested as done so i can close this please?

k8lyn6 (approved):

PR #391, opened 2024-04-26
Wrap text properly on item description field

meisekimiu:
In #4, the description field was made to break words when wrapping text (although it's a bit unclear what bug this was specifically fixing). This commit undoes that so that words wrap more naturally and also match what is displayed when editing text.

Resolves PER-9510: Description should not cut words in half when saved


liam-lloyd (approved):

emlesienk (approved):
Fixes bug that split words in description fields. Text wraps properly.

PR #392, opened 2024-05-01
Fix upload-sourcemaps action

meisekimiu:
The Sentry CLI tries to associate a specific group of commits with a "release" to keep track of when errors happen. Unfortunately, this broke in any new PRs based on the current main branch. I believe this is because the current release, `mdot@2.13.5`, has over 1500 commits in it. Fix this by:
* Fetching the whole commit history when using the upload-sourcemaps action
* Using commit hash to track version information instead of the package.json version

### DIFF
--- a/.github/workflows/build-dev.yml
+++ b/.github/workflows/build-dev.yml
@@ -47,6 +47,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


liam-lloyd (approved):

cecilia-donnelly (approved):
Thanks for figuring this out, @meisekimiu !

Issue #269, opened 2023-07-17
Use `@angular/cdk/dialog` instead of custom dialog service


meisekimiu:
Currently the dialog system used to display modals throughout the app is custom built. This implementation has a few issues. One issue is discussed in #19, but also a few of the dependencies used are currently deprecated and will be removed in future library versions which is something to worry about as we continue to upgrade our version of Angular. The Dialog system also uses various cyclical imports that cause the `import/no-cycle` linting rule to fail when enabled. These cyclical imports cannot be automatically fixed, and are hard to detangle manually without tearing apart the current system.

Fortunately, since we have been upgrading Angular regularly now, a new solution has become available to us: [The Component Dev Kit Dialog Service](https://material.angular.io/cdk/dialog/overview)

We already import and use `@angular/cdk` in our project so this tool is already available to us right now. Because our custom implementation was built to mimic an older Angular dialog library, the CDK Dialog API happens to match the existing API pretty well (with some improvements that would allow us to get rid of things like the `DialogComponentToken` and `DialogChildComponentData` types). Let's use this library instead so we don't have to worry about maintaining the Dialog system ourselves and we can clean up code formatting and linting issues as well.

meisekimiu:
This has been done for a while.

PR #395, opened 2024-05-17
Create new component to select archive type

meisekimiu:
This PR adds some new components to be used in the newly redesigned "GLAM" onboarding flow. These components are currently unused since integrating them will be done later, but the components can be tested with Storybook.

One pretty significant change here is that the new archive type selection component uses the [Angular CDK Dialog](https://material.angular.io/cdk/dialog/overview) service instead of our custom-rolled `DialogService`. As discussed in #269, the current `DialogService` is a bit problematic and we want to eventually move away from it. One big problem with it I noticed while working on this ticket is that it violates some design principles (the [Open Closed Principle](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle) and probably others) as the DialogService needs to know about the names of possible components that it can host ([see this type definition](https://github.com/PermanentOrg/web-app/blob/bbd7fa91e9bbc9cc7dc0bde5e032cbf664df97ea/src/app/dialog/dialog.service.ts#L23)) and also needs module-level configuration to work properly.

I did not want to continue going along with this design for these glam components, especially since these components are probably more volatile than normal (they may be moved around the codebase, they currently exist as standalone components, they may be renamed to remove "glam" from their name in the future). So I decided to use the CDK DialogService for these components instead. To do so, I ended up creating a wrapper service so that we could potentially end up using a form of [branch by abstraction](https://martinfowler.com/bliki/BranchByAbstraction.html) to migrate older code over. Also I wanted it to be clear when components were using the new CDK DialogService. In the future when we migrate over fully, we shouldn't really need the wrapper service anymore and we should import the `AngularCdkDialogModule` directly instead.

Steps to test:
1. Run `npm run storybook`
2. Test the "Glam Onboarding: Archive Type Select" story

Incomplete design questions:
- We currently do not have access to all of the FontAwesome icons used in the design. Do we want to use different icons for archive types? Or should we get FontAwesome pro?
- The design does not show any new descriptions for "Other" or "I"m not sure yet". I've left them empty for now, but what should these say and what icons should they use?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,6 +22,7 @@
         "@angular/router": "17.0.8",
         "@fortawesome/angular-fontawesome": "^0.14.1",
         "@fortawesome/fontawesome-svg-core": "^1.2.28",
+        "@fortawesome/free-regular-svg-icons": "^6.5.2",
         "@fortawesome/free-solid-svg-icons": "^5.13.0",
         "@ng-bootstrap/ng-bootstrap": "16.0.0",
         "@popperjs/core": "^2.11.8",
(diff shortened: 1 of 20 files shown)


crisnicandrei (approved):

crisnicandrei:
Personally, I don't really like the border that appears when you click an archive type, but that is subjective. Besides that, this looks great!

meisekimiu:
I'm going to remove some of the unnecessary comments from the CSS, and I will also add the `type` attribute to buttons.

PR #396, opened 2024-05-17
Call v1 endpoints with the Authorization header

meisekimiu:
We want to pass an authorization header to v1 endpoints now, since some of them require it now. This PR generates the proper headers if an auth token exists and passes them to v1 requests.

**Steps to test:**
1. Log in to the web-app
2. Do something that calls a v1 endpoint
3. Verify that the Authorization header is included in the request
4. Verify that Authorization headers are not passed to unauthenticated requests, such as viewing a public archives when logged out.


cecilia-donnelly (approved):
Looks good to me! I could see the Authorization header for calls when I was logged in and not when I was logged out. I did see it passed for public calls (e.g. searching the public gallery) when I was logged in, but based on your PR message that is expected.

PR #398, opened 2024-05-24
Pass Auth header only to specific v1 endpoints instead of all

meisekimiu:
It turns out that #396 introduced some undetected breaking changes, as some v1 endpoints do not behave well if passed an authorization header. While we should fix the known problematic endpoints, I think it's best that we only pass the auth header to the v1 endpoints need it and that we know work properly with it.

### DIFF
--- a/src/app/shared/services/api/account.repo.ts
+++ b/src/app/shared/services/api/account.repo.ts
@@ -12,7 +12,7 @@ export class AccountRepo extends BaseRepo {
     return this.http.sendRequestPromise<AccountResponse>(
       '/account/get',
       [account],
-      AccountResponse
+      { responseClass: AccountResponse }
     );
   }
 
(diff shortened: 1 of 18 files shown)


k8lyn6 (approved):
I did a pretty thorough test of the platform and everything _seemed_ to work pretty well. I had a few blips (public materials weren’t showing up in fullscreen view on the public archive for certain items, but it seemed to resolve itself. Also had a blip trying to switch archives and had to refresh and try again) but I don’t know if they were related

meisekimiu:
That's reassuring! I think in general, the bug this is fixing shouldn't manifest in "blips" but rather permanent things that don't fully work. On the current mainline branch, signups are completely broken without multiple cookie/localstorage resets.

liam-lloyd (approved):

PR #400, opened 2024-06-04
Create UI for New Member Checklist

meisekimiu:
This PR creates all of the UI components for the New Member Checklist. This PR does not include any of the logic for calling API requests, keeping the checklist hidden forever, or integrating it into the actual web-app UI.

This PR also includes an upgrade to Prettier to support Angular's new "control flow" template syntax, which was used heavily enough in my template HTML to warrant proper formatting of it. This does make this pull request seem unreasonably huge since there are both dependency upgrades and *many* small fixes across the codebase to make sure that formatting checks don't fail with upgraded Prettier rulesets. These are separated into their own commits so they should be easy to ignore while reviewing.

Instead of using SVG files in the `assets` directory, I wanted these components to be somewhat portable so they could work in Storybook. As a result, I designed these components to utilize embedded SVG instead of using the `img` tag to point to an SVG asset.

**Steps to Test:**
1. Run `npm run storybook`
2. View the "New Member Checklist" story

This project will be a test run for a "non-blocking QA" experiment, so it will be merged after a code review without any blocking QA step.


meisekimiu:
omg codecov bot is back!!!

... anyways, if you check the full report you should see that the `user-checklist` part of the codebase has 0 lines of missing coverage. The big Prettier rule upgrade caused a lot of files that don't have full coverage to be reformatted ever so slightly, and as a result we should probably ignore the codecov report here, unless we want to spend the rest of the week covering all of these additional trailing commas.

crisnicandrei (approved):

cecilia-donnelly (approved):
The careful commit messages made this really easy to review even though it's so big! Thanks, @meisekimiu. It looks great in Storybook and makes sense. I'm surprised/interested that we need components for the icons, but it's definitely very readable this way.

PR #405, opened 2024-06-10
Update and delete VS Code config files

meisekimiu:
This project contains specific configuration files for VS Code. They have been hanging around in this repo for several years, and in this time have become less and less relevant over time. This PR either deletes the files outright or updates them to a more modern format. For the record, I'm not sure we actually need `.vscode/settings.json` anymore. It's a "nice to have" for recommended settings in the editor, but it could be deleted without any issues if we want to get rid of it. The main value is setting up the auto-fix config on this repo.

This fixes an annoyance I've had with this repo, which is that `.vscode/settings.json` is often auto-updated by VS Code and I have to make sure I don't stage it with other changes.

**Steps to test:**
1. If you have VS Code, open this repo in it
2. Verify that if you hit "save" on a file there are no weird errors that pop up


liam-lloyd (approved):
I don't have VSCode, but this looks good to me.

PR #406, opened 2024-06-11
Hide new member checklist if marked hidden in backend

meisekimiu:
This PR combines some related "hide the checklist" functionality. It includes: checking for the `hideChecklist` property on accounts, checking for the current archive's access role, and setting the `hideChecklist` value when clicking the "Don't show me this again" button.

I decided to keep things consistent and use an interface to control how much the component can do on its own; as a result it is not passed the `AccountService` directly, but instead controls it through the `ChecklistApi` interface. I think all of the usage of the AccountService could have been included directly in the component, but considering I didn't want to import the whole ApiService earlier, I thought it was only fair to not import the whole AccountService (does this component really need access to most of the AccountService methods?).

Since the component is still not fully integrated into the UI, manual testing is still a bit limited. You can view the automated tests for some testing input, and do some limited manual testing in Storybook (mainly verifying that the "Don't show me this again" button actually hides the component).

### DIFF
--- a/src/app/models/account-vo.ts
+++ b/src/app/models/account-vo.ts
@@ -1,6 +1,6 @@
+/* @format */
 import { BaseVO, BaseVOData } from '@models/base-vo';
 import { AccessRoleType } from './access-role';
-
 
 interface NotificationTypes {
   alerts?: boolean;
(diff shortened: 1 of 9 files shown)


meisekimiu:
Actions are [currently having some issues](https://www.githubstatus.com/incidents/lfrlwdg67fn8) on Github which is causing the apparent linting + test failures. Will rerun checks later.

liam-lloyd (approved):

PR #407, opened 2024-06-14
Integrate checklist into UI

meisekimiu:
This PR adds the checklist to the main private workspace UI. The commits in this PR are *not* based on #406, so when testing this PR some of that functionality will not be included until it is merged into main (this includes the functionality in the "Don't show me this again" button as well as the logic behind actually hiding the checklist). Because of that, we probably shouldn't merge *this* PR until those changes are merged as well, though they can be reviewed in any order.

Additionally, this PR includes changes to the Checklist API since there was a small mismatch with the actual Stela endpoint.


liam-lloyd (approved):

meisekimiu:
I tested that this integrates with #406, so the checklist is hidden for new members unless they have the hideChecklist value manually toggled

PR #412, opened 2024-06-17
Per 9569 redesign user journey

crisnicandrei:
This pr adds the new user journey to the application.

Steps to test: 
Same as before, make sure that isGlam is set to true in your localStorage, then create a new account

Additionally, the new square component can be tested in storybook.


meisekimiu (approved):
Thank you for making Storybook stories for some of this work.

This pull request is a bit difficult to review because of all of the branching + rebasing that has happened. The relevant changes seem good to me for now.

I've created [a ticket](https://permanent.atlassian.net/browse/PER-9659) for next quarter about cleaning up some of these branches by setting up these components to be merged into main without necessarily releasing the new GLAM work yet (similar to what we've done with the New Member Checklist tickets). I think more in-depth QA can happen once this and all other work is able to be integrated into the mainline branch.

meisekimiu (commented):
Actually I do have a followup question: Why are some of the commits in here from the year 2078?

PR #415, opened 2024-06-24
Update New Member Checklist box-shadow

meisekimiu:
Update the shadow to allow it to contrast more against the rest of the user interface. The adjustments are:
- Set the shadow offset to (0,0) so that the shadow is visible on all edges of the checklist
- Update shadow length to be 1em instead of 0.75em, and express it in em units as well
- Multiply the opacity value of the shadow by 2.

Screenshot:
Screenshot from 2024-06-24 13-44-05

### DIFF
--- a/src/app/user-checklist/components/user-checklist/user-checklist.component.scss
+++ b/src/app/user-checklist/components/user-checklist/user-checklist.component.scss
@@ -17,7 +17,7 @@
   width: 22.5em;
   border-radius: 5px;
   background-color: #fff;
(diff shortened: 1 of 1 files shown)


meisekimiu:
Added @k8lyn6 onto this one for feedback, since we probably *should* block this one from being merged until everyone's happy with it. We can adjust the parameters mentioned above if we want a different look to it.

liam-lloyd (approved):

emlesienk (approved):
Looks good to me, showing up correctly in dev

PR #417, opened 2024-06-24
Only show new member checklist on default archives

meisekimiu:
Previously the checklist showed up on any archives the user owned. Now only show on the user's default archive. This PR also includes a light refactoring by renaming the `isArchiveOwnedbyAccount` method to `isDefaultArchiveOwnedByAccount` to match the new criteria.

### DIFF
--- a/src/app/user-checklist/components/user-checklist/shared-mocks.ts
+++ b/src/app/user-checklist/components/user-checklist/shared-mocks.ts
@@ -36,7 +36,7 @@ export class DummyChecklistApi implements ChecklistApi {
     return DummyChecklistApi.accountHidden;
   }
 
-  public isArchiveOwnedByAccount(): boolean {
+  public isDefaultArchiveOwnedByAccount(): boolean {
     return DummyChecklistApi.archiveAccess === 'access.role.owner';
   }
 }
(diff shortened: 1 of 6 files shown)


liam-lloyd (commented):
I'm observing a rather strange bug when I test this: when in a default archive, the checklist disappears upon refresh, and then reappears upon another refresh. It will keep cycling this way indefinitely if I keep refreshing.

meisekimiu:
> I'm observing a rather strange bug when I test this: when in a default archive, the checklist disappears upon refresh, and then reappears upon another refresh. It will keep cycling this way indefinitely if I keep refreshing.

Interesting... I can't recreate this. Can you tell if this is an issue with account/archive information not being loaded when the check is occurring? Or if it is actually just an issue with the default archive check?

liam-lloyd:
> > I'm observing a rather strange bug when I test this: when in a default archive, the checklist disappears upon refresh, and then reappears upon another refresh. It will keep cycling this way indefinitely if I keep refreshing.
> 
> Interesting... I can't recreate this. Can you tell if this is an issue with account/archive information not being loaded when the check is occurring? Or if it is actually just an issue with the default archive check?

Looks like I must have had things in a weird state locally yesterday, because I can't reproduce this behavior anymore either.

liam-lloyd (approved):

meisekimiu:
@liam-lloyd sorry to re-request a review, but in trying to recreate your bug I decided that it would be good for the checklist to subscribe to archive changes so that it appears/disappears without a hard refresh, so I added that to this PR.

liam-lloyd (approved):

PR #418, opened 2024-06-24
Add codecov threshold configuration

meisekimiu:
The codecov/project action is failing since there seems to be some fluctuations in the coverage calculations either on Karma or CodeCov's side. Fix this by setting a threshold for codecov changes. I think 1% is good since the codebase is pretty big; a 1% decrease would be a serious oversight that should be addressed. But the threshold also allows the small coverage fluctuations to not block any pull requests. In general, the patch changes are more important than full project coverage anyway.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,5 @@
+coverage:
+  status:
+    project:
+      default:
+        threshold: 1


meisekimiu:
Note the mysterious -0.02% drop reported in the codecov/project action. This is what I'm talking about, and as we can see from this PR, the action now passes despite reporting a "drop in coverage".

liam-lloyd (approved):

PR #397, opened 2024-05-23
Per 9567 redesign create archive

crisnicandrei:
This is my  pull request for the new create archive step, This adds the new design from firgma to the onboarding flow.

To test this, because I have added a way to disable or enable it, as per @meisekimiu request, go to your browser's console and run the following command: `localStorage.setItem('isGlam',true)`. The functionality is the same as previously implemented

PR #421, opened 2024-07-04
Separate the archive creation Pr

crisnicandrei:
This PR is a follow-up to the PR #1. It separates the archive creation feature into a separate PR.

This is #397 without rebranching from the new sign in page.


meisekimiu (approved):
Thank you for doing this. I don't see any accidental changes that would effect production if this was merged :+1: . I don't think this PR needs QA, since any feedback can be done once each PR is merged and can take the form of additional tickets.

PR #423, opened 2024-07-04
PER-9569: Redesign user journey only

crisnicandrei:
This pr removes the user journey from the previously cramped pr and puts it in its own pr. This pr is only for the user journey redesign.

Same as before, this can be tested setting isGlam to true in your local storage


meisekimiu (approved):
This closes #412.

PR #425, opened 2024-07-17
Remove CSS url imports

meisekimiu:
CSS imports create blocking HTTP requests. Styles in the same sheet are not applied until after the external resource loads, and if the resource does not load it seems like browsers do not apply the rules of the whole stylesheet. This can cause issues if a user has a spotty connection, is blocking one of the external CDNs we are using, or is located somewhere where they are blocked by the CDN.

An easy alternative to this is just to include the imported CSS as a normal stylesheet in the HTML. This is fetched asynchronously, and if it fails it does not prevent other stylesheets from loading or applying.

Resolves PER-9726.

**Steps 2 Test:**
1. Go to the "network" tab of the browser dev console and block requests to `fonts.cdnfonts.com`. Also clear the browser cache in case the external styles are cached.
2. Open Permanent and verify that the UI is not broken.
3. Compare to an environment where this branch is not deployed and verify that it is broken.

For some reason I couldn't get the bug fully reproducible on the local environment... maybe it has something to do with the dev proxy server? Regardless, I have this branch currently deployed to dev for testing and can compare it to staging where the branch is not deployed.


meisekimiu:
Added @k8lyn6 to this already just to get some verification from QA as soon as possible.

k8lyn6 (approved):
Tested on Safari, Firefox, Microsoft Edge, and Chrome and everything was successful. (As a bonus, when I tested on Safari, it was on an account with the new member checklist. I don't know if that additional variable would've changed anything, but it worked fine!)

PR #426, opened 2024-07-19
Per 9572 use dialog cdk

crisnicandrei:
This pull request replaces the previous dialog module from the web app with angular's cdk module.


crisnicandrei:
@meisekimiu i think i have fixed the routed wrapper component, but I would still like your input on the code before I can mark this as ready to review, because I still need to change some designs and add some tests. Thanks!

crisnicandrei:
@meisekimiu my main issues was: `    this.dialogRef = this.dialog.createDialog(
      this.dialogToken,
      dialogData,
      this.dialogOptions,
      this.outletTemplate,
      this.route,
    );

    this.dialogRef.dialogComponent.show();

    this.dialogRef.closePromise.finally(() => {
      this.dialogRef = null;
      if (!this.closedByNavigate) {
        const targetRoute = this.routeHistory.previousRoute || '/app/';
        this.router.navigateByUrl(targetRoute);
      }
    });` because i was trying to initialize the ref using the old method whilst calling dialog.open from the new method, but then I switched it all to use the cdk's methods and it worked

src/app/app.routes.ts
```diff
@@ -26,6 +26,7 @@ export interface RouteData {
   isPublic?: boolean;
 
   dialogToken?: DialogComponentToken;
+  component?: any;
```
meisekimiu, line 29: Does this need to be `any` type? I feel like this should match the type that is allowed by the CDK Dialog `open` method. Also the `Component` type included at the top is currently unused in this file.

src/app/app.routes.ts
```diff
@@ -26,6 +26,7 @@ export interface RouteData {
   isPublic?: boolean;
 
   dialogToken?: DialogComponentToken;
```
meisekimiu, line 28: I believe if we fully switch over to the CDK Dialog system, this property should be unused. We should aim to get this property in a removable (and then removed) state.

src/app/core/components/public-settings/public-settings.component.spec.ts
```diff
@@ -32,14 +32,15 @@ const mockApiService = {
 
 describe('PublicSettingsComponent', () => {
   let shallow: Shallow<PublicSettingsComponent>;
+  let messageShown = false;
```
meisekimiu, line 35: Is this value used in tests at all? Its value doesn't seem to be reset in between test runs if it is used.

src/app/file-browser/components/edit-tags/edit-tags.component.spec.ts
```diff
@@ -198,21 +198,6 @@ describe('EditTagsComponent', () => {
...
-    fixture.detectChanges();
-
-    find('.manage-tags-message .manage-tags-link').triggerEventHandler(
-      'click',
-      {},
-    );
-
-    expect(dialogOpenSpy.open).toHaveBeenCalled();
-  });
-
```
meisekimiu, line 215: I think this is an example of a deleted test that didn't need to be fully deleted. If I comment out or remove the dialog call completely, tests will still pass. Do we really want this behavior to not be covered by tests?

src/app/shared/components/account-dropdown/account-dropdown.component.ts
```diff
@@ -24,11 +24,19 @@ import {
...
   animateChild,
 } from '@angular/animations';
-import { Dialog } from '@root/app/dialog/dialog.module';
-import { SettingsTab } from '@core/components/account-settings-dialog/account-settings-dialog.component';
 import { GuidedTourService } from '@shared/services/guided-tour/guided-tour.service';
 import { GuidedTourEvent } from '@shared/services/guided-tour/events';
 import { EventService } from '@shared/services/event/event.service';
+import { ApiService } from '@shared/services/api/api.service';
+import { DeviceService } from '@shared/services/device/device.service';
+import { AnalyticsService } from '@shared/services/analytics/analytics.service';
```
meisekimiu, line 32: Are these new imports used?

src/app/shared/components/routed-dialog-wrapper/routed-dialog-wrapper.component.ts
```diff
@@ -1,96 +1,74 @@
...
+    const dialogToken = (this.route.snapshot.data as RouteData).dialogToken;
+    const component = (this.route.snapshot.data as RouteData).component;
 
-    if (!this.dialogToken) {
+    if (!dialogToken) {
       throw new Error(
-        "RoutedDialogWrapperComponent - missing dialog token on route data, can't open dialog",
+        "RoutedDialogWrapperComponent - missing dialog token on route data, can't open dialog"
       );
     }
```
meisekimiu, line 61: As far as I understand it, the DialogToken type should be completely unused at this point. This means we should be able to remove these lines checking for DialogToken completely, including this throw statement.

meisekimiu (changes requested):
I think this still needs a bit of work. I've found some components aren't displaying properly such as the Welcome dialog that pops up when a user finishes onboarding, and the Archive Profile screen as well.

The old Dialog code is also still intact. While it isn't being used anymore, it is still imported in a few places. It should be completely removed at this point, since the whole point of this work is to replace it entirely. Once we've completely replaced an old piece of code there is no point in keeping it around.

meisekimiu:
Can you try rebasing this branch onto the current version of main?

crisnicandrei:
@meisekimiu thank you! It seems codecov/patch is sending some warning to some lines that were here before

meisekimiu (commented):
Can you improve the test coverage a bit? The patch coverage change is down about a half of a percent and that is going to block merging this. I pointed out an example of a test that didn't have to be fully deleted in my previous review comments.

crisnicandrei:
@meisekimiu thanks! I will!

src/app/views/components/timeline-view/timeline-view.component.ts
```diff
@@ -152,23 +153,21 @@ export class TimelineViewComponent implements OnInit, AfterViewInit, OnDestroy {
   };
   constructor(
     @Inject(DIALOG_DATA) public dialogData: any,
-    @Inject(OUTLET_TEMPLATE) public outletTemplate: any,
+    // @Inject(OUTLET_TEMPLATE) public outletTemplate: any,
```
meisekimiu, line 156:
```suggestion
```
Let's not commit commented out code.

src/app/shared/components/routed-dialog-wrapper/routed-dialog-wrapper.component.ts
```diff
@@ -97,4 +66,24 @@ export class RoutedDialogWrapperComponent
...
+    };
+
+    this.dialogRef = this.dialogService.open(component, config);
+
+    this.dialogRef.closed.subscribe(() => {
+      this.dialogRef = null;
+      if (!this.closedByNavigate) {
+        const targetRoute = this.routeHistory.previousRoute || '/app/';
+        this.router.navigateByUrl(targetRoute);
+      }
```
meisekimiu, line 86: While sometimes these codecov warnings can be ignored, this is a good example of a place where test coverage would have been useful as part of the refactoring process. If I were restructuring this specific bit of code, I would have written tests to cover the original routed dialog wrapper *before* doing any CDK Dialog conversions. Once these were written and passed, *then* I would begin refactoring. In an ideal world, once we restructure the RoutedDialogWrapper the tests would still pass proving that our refactoring didn't break anything. In practice, we'd probably have to edit the tests to match the new API of the CDK Dialog. But I do recommend writing some tests up front before doing actual restructuring, so that they [characterize  the preexisting behavior](https://en.wikipedia.org/wiki/Characterization_test) and we can verify that behavior hasn't changed.

meisekimiu (approved):
This looks mostly good, though I do have some advice for future refactoring PRs. Ideally a refactoring should have no effect on functionality (and no negative effect on test coverage). One of the main benefits of unit tests is that they enable [refactoring](https://refactoring.com/)  by allowing us to make modifications to the code and verify that we have not altered any behavior or introduced any bugs in the process.

I highly recommend for future refactoring work to work in small steps (converting over a component at a time, for example) and running tests after each of these small adjustments so you can confirm nothing is broken. To me it seems like all component conversions were done first and then tests were adjusted to pass again after the fact. The advantage to running tests frequently after each change is that you can quickly identify *what* broke tests. It's a lot harder when you have many failing tests.

I will approve this, though I recommend having a manual QA pass at this as well.

meisekimiu:
Adding @k8lyn6 onto this. The PR can be manually tested by verifying that none of the modals in the app are broken.

k8lyn6:
@emlesienk to QA this, go through and open all the modals across the app (the new windows that open on top of workspaces). These pop open after clicking most buttons and menu items. Let us know if there are any that appear broken.

emlesienk (commented):
TLDR; Notifications and "add connection" aren't opening properly. The rest of the issues seem to be formatting issues that cause the designs to look too tight (see screenshots).
- The notifications module isn't opening properly
- Items in the account menu show a box around the x in the corner 
- Won't open the "add connection" in the Connections module
- The "pending archive" box is off, design is cut off / not enough space. Same deal with the "share" function.
- Publish option screen looks different than what's on prod


menu_screenshot
publish_screenshot
share_screenshot
add_connection_screenshot
pending_connections_screenshot
storage_screenshot
notifications_screenshot

k8lyn6:
@emlesienk can you re-review?

emlesienk (approved):
Looks good, thanks Andrei!

PR #427, opened 2024-07-19
Fix error when destroying hidden checklist

meisekimiu:
The checklist is designed to skip subscribing to refresh events if it is completely hidden in the current view since there's no point to refreshing the checklist if it is completely disabled. This causes an error to be thrown when we later unsubscribe from this undefined subscription. Do a check for undefined when unsubscribing and make the refresh subscription be defined even for hidden checklists, since the checklist could reappear if the user changes archives.

PER-9738: Navigating to Public Gallery from an archive is broken


liam-lloyd (approved):
I can't actually reproduce the bug locally for some reason, but this looks good and I can see that the bug doesn't occur on this branch.

PR #431, opened 2024-07-29
Per 9566 glam invited as archive member

crisnicandrei:
This pr adds the new glam design to the invitations screen. 

To test, same as before, please set isGlam to true in your local storage


meisekimiu (changes requested):
Why are there two glam-pending-archive components? I also do not have a visible access role when testing.

Screenshot from 2024-07-30 13-36-03

crisnicandrei:
@meisekimiu what do you mean about the last part of your comment? I am confused

crisnicandrei:
@meisekimiu which accessRole were you using? I might have missed one

crisnicandrei:
@meisekimiu one component (pending-archives) display all the archives, and pending-archive is the archive itself with the button and the image

meisekimiu:
> @meisekimiu which accessRole were you using? I might have missed one

I was using the owner access role

crisnicandrei:
@meisekimiu can you invite someone else as owner?

meisekimiu:
Screenshot from 2024-08-05 09-03-34
The owner role still does not show up properly.

To test this locally I rebased it onto main since I did a bit of refactoring in #433, and also wanted to use the new Storybook flow instead of having to set up invitations manually. Want me to commit the resolved conflicts to this branch for you?

crisnicandrei:
@meisekimiu yes please!

It seems I got a bit confused there tho, I thought you are not able to invite others as owners to your archive. But it seems I was wrong.

src/app/component-library/components/button/button.component.ts
```diff
@@ -24,6 +24,7 @@ export class ButtonComponent {
   @Input() orientation: ORIENTATION = 'left';
   @Input() faIcon: string = '';
   @Input() buttonType: TYPE = 'button';
+  @Input() textColor: { color: string };
```
meisekimiu, line 27: Instead of having the user pass in a a pure CSS style object like this, I think it would be more convenient for the user just to pass in a color directly and the component should wrap that up into a `{color: ...}` object internally.

meisekimiu (approved):

k8lyn6 (approved):
Looking good so far! Assuming this can be merged without deploying it, I'll approve this and leave comments in a new ticket like we did for the rest of onboarding. You can find the [ticket here.](https://permanent.atlassian.net/browse/PER-9807)

Thanks Andrei!

PR #433, opened 2024-07-31
Update navigation bar for GLAM onboarding

meisekimiu:
This pull request features a style adjustment to GLAM onboarding. Specifically, the header or navigation bar at the top of the page now matches the one seen in the GLAM designs. The original header is preserved for the current onboarding flow while GLAM is not enabled.

In addition, this PR features some small refactoring work. Right now it mainly involves just moving some components around the filesystem. The Onboarding Module's components folder has a "/glam" subdirectory that I created to hold all of the new "GLAM" components so it was easy to tell which components were new and which ones were old. There has been some inconsistency in whether or not new components land in this folder. I have moved each GLAM component over to the `glam` directory so that it's easy to identify the new and old subcomponents. Ideally there'd also be some kind of "classic" or "legacy" folder for the old component, but since the GLAM components are currently used in the create-new-archive component, it felt misleading to fully separate them. That will probably come in a future refactoring. Before doing that I wanted to improve the onboarding testing workflow a bit, which brings me to...

The other big addition I tacked onto this PR is the addition of **a full "Onboarding Demo" suite of stories in Storybook!** This allows us to test the onboarding flow without having to create a million accounts each time, or even be connected to the API at all. It also allows for fast switching between the various onboarding flows (invitations, glam, etc). If you recall when onboarding was first developed, we had a debugger component to help test with things while it was still under construction.

<details>
<summary>Finally after a few years that debugging functionality lives on in Storybook!</summary>

onboarding-debug-component

</details>

Steps to test:
1. Create a new account and activate GLAM onboarding (`localStorage.setItem('isGlam', 'true');`)
2. Make sure the header looks good and works

You can also view the header in the new Onboarding Demo on Storybook, though the "sign out" button doesn't fully work in Storybook since there are no user sessions and you cannot be redirected to the sign in screen.


meisekimiu:
Screenshot from 2024-07-31 16-32-40

A screenshot of Onboarding running in Storybook. You can swap between GLAM/original and norma/invitation flows with the controls on the bottom!

liam-lloyd (approved):

meisekimiu:
@k8lyn6 This branch has been merged already, and probably needs non-blocking QA. It probably makes sense to do some blocking QA for the other onboarding feedback branches, but I wanted to get the Storybook demo merged as soon as possible to assist in testing the other feedback branches.

One visual compromise that had to be made on this header was that the FontAwesome icon we were using for the sign out button was unavailable, so I swapped it with a similar one.

PR #434, opened 2024-08-01
Per 9755 update welcome to permanent onboarding screen

crisnicandrei:
This pull request solves some of the issues found by the qa team on the first page of the new glam onboarding


meisekimiu:
I've also rebased this branch onto main so that it can be tested via Storybook.

meisekimiu (changes requested):
I think these changes look good, though I do have some small things I've noticed.

In the original design, the greeting text (*"Hello, (name). Welcome to Permanent!"*) is still slightly different:
* The font size is actually slightly bigger, at 56px. I would probably convert that to rem units.
* The line-height of these elements is set to be 72px. Again, I would personally convert this to be rem units
* The letter spacing should be set to `letter-spacing: -0.02em;`
* In this implementation it's broken up into two `<p>` tags. Somewhere in our stylesheets `<p>` tags are set to have additional bottom margin by default so it creates some additional space that isn't there in the design. This can be fixed by removing that margin (or using only one `<p>` tag and a `<br />` tag between the two lines).

Something for additional discussion:
This ticket requests that the buttons should be in the middle of the screen rather than flush with the bottom, and I think this does look good although I do wonder what the implications of this are for viewports that happens to be really tall (in portrait mode, but otherwise wide enough to not trigger mobile view). Maybe that's something for another ticket, but I thought I'd bring it up here in case the experience team wants to have that discussion. Right now with the currently defined styles the buttons just kind of float in the middle of the screen and there is a lot of empty space at the bottom of the page.

src/app/onboarding/components/archive-creation-start-screen/archive-creation-start-screen.component.scss
```diff
@@ -31,9 +31,16 @@ p {
   @media screen and (max-width: 600px) {
     width: 100%;
   }
-  & > p {
-    font-size: 50px;
-    font-family: 'UsualRegular', sans-serif;
+  & > .greeting {
+    font-size: 4rem;
+    font-family: 'UsualLight', sans-serif;
+    line-height: 4rem;
```
meisekimiu, line 37:
```suggestion
    font-size: 3.5rem;
    font-family: 'UsualLight', sans-serif;
    line-height: 4.5rem;
```
If we were converting the units from the Figma design to rem it would be these values, right? (I'm pretty sure 1rem = 16px)

meisekimiu:
It also seems like my rebasing work got force-pushed over. I'm not entirely sure this could have been avoided if you pushed with the `--force-with-lease` option but I do recommend using that over the standard `-f` flag in the future.

meisekimiu (approved):
I rebased this onto main again. I'll let the experience team handle feedback from here.

yeslikesolo (commented):
Things are looking good! Andrei and I met on Wednesday 8/14 to fix the responsiveness of text flow for certain breakpoints (screen sizes) which looked great. But, it doesn't look like those changes are showing up as I QA'd today? Just wanted to note that incase the new changes were not pushed through yet!

One other note: error state icons are showing up inside the "Name archive" input field, and it should be outside. Please see the attached image.
Screenshot 2024-08-16 at 11 44 33 AM

crisnicandrei:
@yeslikesolo those changes are on the latest open PR, not this one 😃

yeslikesolo (approved):

k8lyn6:
Oh, actually @emlesienk it looks like this might already be taken care of. I'll remove you from the request.

@meisekimiu @crisnicandrei it looks like this might be approved, since @yeslikesolo approved it last week?

meisekimiu:
Yes this is done, I just hadn't merged it since tests needed to run again. Merging now!

PR #437, opened 2024-08-09
Per unable to click items timeline view

crisnicandrei:
This Pr fixes only the background issue of the timeline view, as @meisekimiu mentioned, we will have to changed the timeline view to not use the dialog anymore and that needs a bit of investigating


meisekimiu (approved):

emlesienk (commented):
Background is showing up now as expected. As written in the documentation, still awaiting the fix for selecting and viewing a file from a timeline.

PR #438, opened 2024-08-12
Per 9743 replace vagrant with docker

valentinvsp:
Add Dockerfile, in order to follow the pattern observed in other repos. This Dockerfile is used by the docker-compose.yml from the **devenv** repo.

A separate start command & config was added for starting the angular dev server, in order to configure it for working with  docker.


cecilia-donnelly (commented):
Going to wait for @meisekimiu to get back to give this a full review.

meisekimiu (approved):

PR #439, opened 2024-08-12
Per 9756 update archive type onboarding screen

crisnicandrei:
Design changes for the archive type select screen. 

@yeslikesolo I am missing some texts for the "Other" and "I'm not sure yet" so we keep the header text format please!

### DIFF
--- a/src/app/onboarding/components/glam/archive-type-select-dialog/archive-type-select-dialog.component.scss
+++ b/src/app/onboarding/components/glam/archive-type-select-dialog/archive-type-select-dialog.component.scss
@@ -1,4 +1,5 @@
 @import 'variables';
+@import 'fonts';
 
 .archive-type-select-dialog {
   font-family: 'UsualRegular', sans-serif;
@@ -76,6 +77,7 @@
       color: $darkBlue;
       font-weight: 600;
       line-height: 24px;
+      font-family: 'UsualExtraBold', sans-serif;
     }
 
     .type-description {
--- a/src/app/onboarding/components/glam/archive-type-select-dialog/archive-type-select-dialog.component.scss
+++ b/src/app/onboarding/components/glam/archive-type-select-dialog/archive-type-select-dialog.component.scss
@@ -3,7 +3,7 @@
 .archive-type-select-dialog {
   font-family: 'UsualRegular', sans-serif;
   background: #fff;
-  border-radius: 12px 12px 0px 0px;
+  border-radius: 12px;
   display: flex;
   width: auto;
   height: 80vh;
@@ -32,7 +32,7 @@
   align-items: center;
   flex-shrink: 0;
   align-self: stretch;
-  border-bottom: 1px solid #E7E8ED;
+  border-bottom: 1px solid #e7e8ed;
 
   .dialog-label {
     color: $darkBlue;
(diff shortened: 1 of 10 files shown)


src/app/onboarding/components/glam/archive-type-select/archive-type-select.component.scss
```diff
@@ -9,7 +10,7 @@
 
   border-radius: 12px;
   border: 1px solid rgba(255, 255, 255, 0.16);
```
meisekimiu, line 12:
```suggestion
  border: none;
```
I think one of the things requested in the ticket is to remove a weird orange line that appears near the component. It seems like this change gets rid of that.

src/app/onboarding/components/glam/archive-type-select-dialog/archive-type-select-dialog.component.scss
```diff
@@ -5,8 +5,8 @@
   background: #fff;
   border-radius: 12px 12px 0px 0px;
```
meisekimiu, line 6:
```suggestion
  border-radius: 12px;
```
The design for this dialog actually changed at some point after it was initially implemented. It is now not flush with the bottom of the screen, so I believe we have a consistent border radius now.

src/app/onboarding/components/glam/select-archive-type-screen/select-archive-type-screen.component.ts
```diff
@@ -12,29 +12,33 @@ import {
...
-        tag: this.tag,
-        headerText: this.headerText,
-      });
+  ngOnInit(): void {
+    if (this.tag) {
+      this.buttonText = generateElementText(
+        this.tag,
+        archiveOptionsWithArticle,
+      );
     }
```
meisekimiu, line 30: Codecov is pointing out that this is missing coverage. It would be nice to write a test that renders the component with the `tag` input provided.

meisekimiu (changes requested):
I've added some small bits of initial feedback before we pass it to QA.

src/app/onboarding/components/glam/archive-type-select/archive-type-select.component.scss
```diff
@@ -1,3 +1,4 @@
+/*N @format */
```
meisekimiu, line 1:
```suggestion
/* @format */
```
The prettier pragma got a bit messed up here.

meisekimiu (approved):

yeslikesolo (changes requested):
Screenshot 2024-09-05 at 5 16 07 PM

- Please decrease the width of H1 component Create your first archive, to half its size. It should stack or be wrapped text, aligned left. 
- Increase bold for "Archive" in h1 text for create new archive component

Screenshot 2024-09-05 at 5 33 57 PM

- bold archive type names: Personal, An Individual, Family, etc.

crisnicandrei:
@yeslikesolo can't do much bolder, this is tthe only bold font that i have :(

k8lyn6:
@crisnicandrei could you confirm what font you're using for onboarding? The bold/not bold has been hard to see across onboarding and I just want to check we're using the right font. If the font is correct, could you coordinate with Tibi about why it's not looking like the design? 

@yeslikesolo once Andrei has checked that, could you re-review this?

crisnicandrei:
@k8lyn6 @yeslikesolo I think I got it right this time. Can you please recheck? It was Usual, but I had to use the extra bold instead

yeslikesolo (approved):

PR #441, opened 2024-08-20
Design changes to the name archive screen

crisnicandrei:
This pull request adds the changes requested in the ticket to the new name archive screen

### DIFF
--- a/src/app/component-library/components/button/button.component.html
+++ b/src/app/component-library/components/button/button.component.html
@@ -12,6 +12,7 @@
   ]"
   [disabled]="disabled"
   (click)="onClick($event)"
+  [style]="style"
 >
   <img *ngIf="icon" class="icon" src="{{ 'assets/svg/' + icon + '.svg' }}" />
   <fa-icon class="icon" *ngIf="faIcon" [icon]="['fas', faIcon]"></fa-icon>
--- a/src/app/component-library/components/button/button.component.ts
+++ b/src/app/component-library/components/button/button.component.ts
@@ -24,7 +24,7 @@ export class ButtonComponent {
   @Input() orientation: ORIENTATION = 'left';
   @Input() faIcon: string = '';
   @Input() buttonType: TYPE = 'button';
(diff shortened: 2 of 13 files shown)


meisekimiu (approved):

k8lyn6 (changes requested):
Looking good! A few notes:

- The error symbol is still hiding behind the word "Archive"
- Only "personal" should be bolded in the left side text
- All the items on the left side (Permanent logo, status bar, and text) should align when screen is resized. Currently, they don't look quite aligned
- I can't see the Permanent logo image -- it's broken for me (not sure if this is a me problem or something else)
- The back and next buttons should be the same size
- When the viewport switches to a mobile size, can we remove some of the space between the archive name field and the buttons, so you don't have to scroll as much?

k8lyn6 (changes requested):
Looks like most things are fixed! Two questions:

1. It's really hard to tell, but is only the archive type bolded on the left side? "Create your" and "Archive" should be normal font, and "Personal" or "Family History" or whatever should be bolded.
2. I still can't see the Permanent logo image. I think we're okay to merge this once you address the first issue, but we'll want to double check it's working before release.

crisnicandrei:
@k8lyn6 can you check now? It should be bolder, this is the boldest I can do. And regarding the image, it works for me so I really don't know what is going on

k8lyn6:
@crisnicandrei Compared to the design, the boldness just looks different. Are you using the font from the design (I think it's called Usual).

crisnicandrei:
@k8lyn6 did some changes

k8lyn6 (changes requested):
Looks good! Just one minor change -- the buttons are the right size, but the text and icon should be centered.

crisnicandrei:
@k8lyn6 pushed the changes!

k8lyn6 (approved):
I think this looks good! Also checked old onboarding to make sure that nothing here changes anything about old onboarding, and it looks fine.

PR #443, opened 2024-08-21
Handle new thumbnail sizes

meisekimiu:
We will eventually be switching over to having Archivematica-generated thumbnails on records. Ideally we want this switchover to be pretty seamless from the web-app's perspective (in other words, we shouldn't need to make sure we merge code in parallel or have some kind of feature flag). This pull request creates a helper function called `GetThumbnail` that adds a layer of abstraction to fetching thumbnails on an item. This function takes in a minimum size, and it will try to pick the best thumbnail to match that size.

This functionality allows us to:
- Stop generating `thumbURL200`, `thumbURL500`, etc properties on objects without any web-app consequences.
- Introduce Archivematica 256x256 thumbnails and have them be used in the web-app
- Add new Archivematica-generated thumbnail sizes later with ease
- Centralize the logic originally contained in the `pr-thumbnail` component

This pull request also features some refactoring: the `pr-thumbnail` component has been refactored significantly and all OpenSeaDragon functionality has been extracted into its own component. Both components should have robust test coverage now.

**Testing Information**
Right now, we can't fully test this in a deployed environment until real Archivematica thumbnails are generated. In theory this could be simulated by modifying a record in the database to only have a 256x256 thumbnail. We do have unit tests in place to test the `GetThumbnail` functionality, but any manual testing should probably just check to see if thumbnails still load properly throughout the app and that the image viewer still works.

### DIFF
--- a/src/app/file-browser/components/file-viewer/file-viewer.component.html
+++ b/src/app/file-browser/components/file-viewer/file-viewer.component.html
@@ -42,8 +42,6 @@
           "
         ></pr-zooming-image-viewer>
         <pr-thumbnail
-          (disableSwipe)="toggleSwipe($event)"
-          (isFullScreen)="toggleFullscreen($event)"
           [item]="currentRecord"
           *ngIf="
             showThumbnail &&
--- a/src/app/file-browser/components/file-viewer/file-viewer.component.html
+++ b/src/app/file-browser/components/file-viewer/file-viewer.component.html
@@ -29,12 +29,29 @@
         *ngIf="isQueued(i)"
       ></div>
       <div *ngIf="currentRecord === record" class="full" [ngSwitch]="">
+        <pr-zooming-image-viewer
+          [item]="currentRecord"
+          (disableSwipe)="toggleSwipe($event)"
+          (isFullScreen)="toggleFullscreen($event)"
+          *ngIf="
+            showThumbnail &&
+            isZoomableImage &&
+            !isDocument &&
+            !isVideo &&
+            !isAudio
+          "
+        ></pr-zooming-image-viewer>
         <pr-thumbnail
           (disableSwipe)="toggleSwipe($event)"
           (isFullScreen)="toggleFullscreen($event)"
           [item]="currentRecord"
-          *ngIf="showThumbnail && !isDocument && !isVideo && !isAudio"
-          [hideResizableImage]="false"
+          *ngIf="
+            showThumbnail &&
+            !isZoomableImage &&
+            !isDocument &&
+            !isVideo &&
+            !isAudio
+          "
         ></pr-thumbnail>
         <pr-video [item]="currentRecord" *ngIf="isVideo"></pr-video>
         <pr-audio [item]="currentRecord" *ngIf="isAudio"></pr-audio>
--- a/src/app/models/folder-vo.ts
+++ b/src/app/models/folder-vo.ts
@@ -10,6 +10,7 @@ import { TimezoneVOData } from './timezone-vo';
 import { FolderType, SortType, FolderLinkType } from './vo-types';
 import { LocnVOData } from './locn-vo';
 import { TagVOData } from './tag-vo';
+import { HasThumbnails } from './get-thumbnail';
 import { ItemVO, ArchiveVO } from '.';
 
 export interface HasParentFolder {
@@ -30,7 +31,7 @@ export interface ChildItemData {
 
 export class FolderVO
   extends BaseVO
-  implements ChildItemData, HasParentFolder, DynamicListChild
+  implements ChildItemData, HasParentFolder, DynamicListChild, HasThumbnails
 {
   public isFolder = true;
   public isRecord = false;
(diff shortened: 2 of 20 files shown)


src/app/models/folder-vo.ts
```diff
@@ -66,12 +67,16 @@ export class FolderVO
...
-  public thumbURL2000;
+  public thumbURL200: string;
+  public thumbURL500: string;
+  public thumbURL1000: string;
+  public thumbURL2000: string;
   public thumbDT;
 
+  // New thumbnails
+  public thumbnail256: string;
+  public thumbnail256CloudPath: string;
```
meisekimiu, line 78: I'm actually not sure what the `CloudPath` property should be used for, if it should be used at all.

src/app/shared/components/thumbnail/thumbnail.component.ts
```diff
@@ -5,124 +5,60 @@ import {
...
-  @Output() isFullScreen = new EventEmitter<boolean>(false);
-
-  viewer: OpenSeaDragon.Viewer;
+  private debouncedResize: () => void | undefined;
 
   constructor(
     elementRef: ElementRef,
-    private renderer: Renderer2,
-    private zone: NgZone,
+    @Inject('Image') private imageClass: typeof Image,
```
meisekimiu, line 50: This allows us to mock image loading in unit tests. Previously we were actually sending out HTTP requests in tests to a defunct placeholder image generation website. This could still be useful for the HTML test runner and Storybook, but we should be able to run all of our unit tests without relying on the reliability of external hosts or even being connected to the internet at all.

src/app/shared/components/zooming-image-viewer/zooming-image-viewer.component.ts
```diff
@@ -0,0 +1,100 @@
...
+  @Output() public isFullScreen = new EventEmitter<boolean>(false);
+
+  public viewer: OpenSeaDragon.Viewer;
+
+  private initialZoom: number;
+
+  @ViewChild('viewer') element!: ElementRef<HTMLElement>;
+
+  constructor(
+    @Inject('openseadragon') private imageViewerFn: typeof OpenSeaDragon,
```
meisekimiu, line 35: This follows a similar pattern to mocking the `Image` type earlier. For some reason, OpenSeaDragon takes a long time to initialize in unit tests (to the point where it was reliably causing test failures since the tests were timing out!). I traced the performance issues specifically to the `OpenSeaDragon()` initialization call, so the best course of action was just to mock the OpenSeaDragon object. This is unfortunate because although we generally don't want to test 3rd party libraries, OpenSeaDragon gives us the `raiseEvent` method that would have allowed us to cover all of our custom event handlers easily without any mocking... if only it weren't so slow!

<details>
<summary>Additional mysterious details which are not relevant anymore</summary>
For some reason changing image URLs seemed to have a noticeably significant impact on test run times. Choosing a valid image URL would make tests time out every time, while choosing an invalid one would allow them to at least pass within the 5000ms window most of the time. This seems to indicate that somehow OpenSeaDragon is fetching and initializing images as part of its setup process. But the call to initialize OpenSeaDragon is not in the form of an async/await/promise call. And the tests themselves also do not have any async behavior besides the shallow rendering. The only way to load image assets in JavaScript is through asynchronous calls though, so somehow in unit tests OpenSeaDragon was causing blocking issues with async calls. I don't even know how that is remotely possible. If event handlers weren't called until the images were loaded, this would just cause the tests to fail instantly since we don't have any asynchronous expects in the tests. My only theory is that there was some kind of weird interaction between OpenSeaDragon and shallow-render? I still don't know how changing image URLs would affect that though. This issue gave me a headache.</details>

src/app/file-browser/components/file-viewer/file-viewer.component.html
```diff
@@ -29,12 +29,29 @@
...
-          *ngIf="showThumbnail && !isDocument && !isVideo && !isAudio"
-          [hideResizableImage]="false"
+          *ngIf="
+            showThumbnail &&
+            !isZoomableImage &&
+            !isDocument &&
+            !isVideo &&
+            !isAudio
+          "
         ></pr-thumbnail>
```
meisekimiu, line 55: In theory I think we could get rid of this `pr-thumbnail` element completely, but I think we still want to use this for things like ZIP files and other archive file types.

meisekimiu (commented):
Some comments I have to help with review/discussion

crisnicandrei (approved):

meisekimiu:
Added @k8lyn6 onto this. I wrote testing information in the original pull request comment, but to summarize, this should mainly just be checking that existing thumbnails all continue to work and that the zooming image viewer also works. Edge cases we might want to test in particular:
- Public archive folder thumbnails
- ZIP files do not open up an image viewer
- Thumbnails appear correct in the right side metadata viewer
- Archive thumbnails

emlesienk (commented):
Thumbnails are still loading and image viewer is working

k8lyn6:
@emlesienk Is this good after QA review? If there are no changes needed, can you approve the pull request?

emlesienk:
@k8lyn6 It's not letting me. You might have to reassign.

meisekimiu:
This PR has been merged already so it doesn't need additional approval

k8lyn6:
ah ok, sorry about that! Thanks @meisekimiu

PR #447, opened 2024-08-30
Add gift code result message to modal

meisekimiu:
The MessageService currently cannot show messages above modals. The current form to redeem a gift code is hosted in a modal, so the error and success messages are not visible to users. Show the error and success messages in the modal instead.

**Steps to test:**
1. Open the storage dialog and open the redeem gift tab
2. Test that an error message appears with an invalid promo code
3. Test that a success message appears with a valid one

### DIFF
--- a/src/app/core/components/redeem-gift/redeem-gift.component.ts
+++ b/src/app/core/components/redeem-gift/redeem-gift.component.ts
@@ -1,0 +1,91 @@
+/* @format */
+import { Component, Input, OnInit } from '@angular/core';
+import {
+  UntypedFormGroup,
+  UntypedFormBuilder,
+  Validators,
+} from '@angular/forms';
+import { PromoVOData } from '@models';
+import { ApiService } from '@shared/services/api/api.service';
+import {
+  BillingResponse,
+  AccountResponse,
+} from '@shared/services/api/index.repo';
+import { FileSizePipe } from '@shared/pipes/filesize.pipe';
+import { AccountService } from '@shared/services/account/account.service';
+import { EventService } from '@shared/services/event/event.service';
+
+interface StorageRedemptionMessage {
+  successful: boolean;
+  message: string;
+}
+
+@Component({
+  selector: 'pr-redeem-gift',
+  templateUrl: './redeem-gift.component.html',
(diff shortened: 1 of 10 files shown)


src/app/core/components/redeem-gift/redeem-gift.component.ts
```diff
@@ -0,0 +1,91 @@
...
+    });
+    this.promoForm.reset();
+  }
+
+  private async updateAccountStorageBytes(response: BillingResponse) {
+    await this.account.refreshAccount();
+    const promo = response.getPromoVO();
+    const bytes = promo.sizeInMB * (1024 * 1024);
+    this.account.addStorageBytes(bytes);
+    this.showPromoCodeSuccess(bytes);
```
liam-lloyd, line 81: It feels like this call is more part of "handling a valid promo code" than "updating account storage"; should it be in `handleValidPromoCode` instead?

src/app/core/components/redeem-gift/redeem-gift.component.ts
```diff
@@ -0,0 +1,91 @@
...
+  styleUrl: './redeem-gift.component.scss',
+})
+export class RedeemGiftComponent implements OnInit {
+  @Input() public promoCode: string = '';
+  public resultMessage: StorageRedemptionMessage | undefined;
+  public promoForm: UntypedFormGroup;
+  public waiting: boolean;
+
+  constructor(
+    private fb: UntypedFormBuilder,
```
liam-lloyd, line 35: I know this is a pre-existing variable name, but this is probably a good opportunity to change it to something more informative.

liam-lloyd (approved):
Two quite minor comments, otherwise looks good (and works correctly upon testing).

src/app/core/components/redeem-gift/redeem-gift.component.ts
meisekimiu: The main reason it's called here is that we want to show the user how much storage space was redeemed. I didn't want to make the redeemed bytes a property of the class and it also felt weird to return the bytes from this method as well.

k8lyn6 (approved):
Thanks for fitting this in, @meisekimiu ! It looks good to me.

PR #450, opened 2024-09-09
Design changes

crisnicandrei:
More changes to the what's important to you screen

### DIFF
--- a/src/app/component-library/components/button/button.component.html
+++ b/src/app/component-library/components/button/button.component.html
@@ -11,6 +11,7 @@
     orientation === 'right' ? 'button-reverse' : ''
   ]"
   [disabled]="disabled"
+  [style]="style"
   (click)="onClick($event)"
   [style]="style"
 >
--- a/src/app/component-library/components/checkbox/checkbox.component.html
+++ b/src/app/component-library/components/checkbox/checkbox.component.html
@@ -5,6 +5,7 @@
   [class.checkbox-container-enabled]="!disabled"
   [class.checkbox-container-primary]="variant === 'primary'"
   [class.checkbox-container-secondary]="variant === 'secondary'"
+  [class.checkbox-container-secondary-onboarding]="variant === 'secondary' && onboarding"
   (click)="toggleCheck()"
   (keydown)="handleKeyDown($event)"
 >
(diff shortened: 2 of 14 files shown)


meisekimiu (approved):

k8lyn6 (changes requested):
I think most of these changes look good! 

One small thing: When the screen expands, the progress bar should probably stop expanding at some point and stay aligned with the chips and text below, and any extra space should just be filled with the blue background on the right and left sides.

crisnicandrei:
@k8lyn6 thanks! I have added a max-width to the progress bar, let me know if this works or it should be less?

yeslikesolo:
QA Notes:
The goal is to move "Let's set some goals" component to the right and fill the spacing for screens larger than 1440px. Two possible solutions:

- Increase spacing between "Chart your path" and "Let's set some goals" div boxes.
- If possible, adjust width of "let's set some goals" component to auto flex based on the width of the screen

Let me know if I can clarify anything!

Screenshot 2024-09-19 at 1 38 35 PM

yeslikesolo:
Another note:
You can see the blue background isn't moving dynamically with the screen. The white shows up on the right side behind the selection choices when I open and close the screen. The left side removes the blue margin and "Chart your path" stays flushed against the side. This issue is occurring at ~960px? (I can't truly tell my screen size, so this is a guess.) 

https://github.com/user-attachments/assets/c515f2b8-ef20-42c5-8709-bdef54a513e5

yeslikesolo:
"Tell us what's important to you" font seems to be bolded differently than previous header text. Please change to match "chat your path to success" font settings! Thank you (:

Screenshot 2024-09-19 at 2 09 07 PM

crisnicandrei:
@yeslikesolo pushed the changes, but could you explain the auto flex part to me please?

k8lyn6:
@yeslikesolo do you mind continuing to work with Andrei on this?

k8lyn6 (changes requested):
Small change -- I think the text and icons on the buttons are supposed to be centered according to the design. Once you've changed that, let's merge it!

crisnicandrei:
@k8lyn6 done!

k8lyn6 (approved):
Looks good!

PR #456, opened 2024-09-12
PER-9807-Design changes to archive invite

crisnicandrei:
Design changes requested to the invite to archive screen

### DIFF
--- a/src/app/onboarding/components/create-new-archive/create-new-archive.component.ts
+++ b/src/app/onboarding/components/create-new-archive/create-new-archive.component.ts
@@ -102,6 +102,10 @@ export class CreateNewArchiveComponent implements OnInit {
   }
 
   ngOnInit(): void {
+    if (this.pendingArchives.length && !this.pendingArchive) {
+      this.screen = 'create';
+    }
+
     if (this.pendingArchive) {
       this.screen = 'goals';
       this.progress.emit(1);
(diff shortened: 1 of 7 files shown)


k8lyn6 (changes requested):
Looking good!

- When someone clicks "Accept" the text should change to say "Accepted" not "Accept"
- The last page of onboarding should list all new archives that the member accepted the invitation to. Right now, it looks like it only lists one, even though multiple were accepted (I checked in my account afterwards, so it does look like it's working as expected. They're just not listed on that final page). @crisnicandrei is this addressed in another ticket already? If not, I can create a ticket.
- Do you also mind checking with @tiberiuVSP  on the size of the help text? It looks small and a little hard to read. If the size of the text is correct with the design, Tibi might want to think about adjusting it to a larger size so it's more readable.

crisnicandrei:
@k8lyn6 that is not addressed in another ticket. I can make one tomorrow of you don't have time today😀

k8lyn6:
@crisnicandrei [here's the new ticket!](https://permanent.atlassian.net/browse/PER-9891) Thanks! Taking a look at your changes to this PR now.

k8lyn6 (approved):
Changes look good! @crisnicandrei before you merge this, do you mind checking on the font size with @tiberiuVSP ?

crisnicandrei:
@k8lyn6 I have checked the fonts with him, and made a few changes

PR #457, opened 2024-09-13
Profile settings dialog fixes

meisekimiu:
This PR features two fixes to the profile settings dialog:
- The "view on web" button has now been fixed
- The change profile / banner buttons now work as expected

This can be tested locally or by deploying to the `dev` environment.

**Steps to test:**
1. Verify that an archive's profile and banner can be updated
2. Verify that the "view on web" button now works
    - When on the archive's public profile, hit the back button in the browser, and then verify that the close button (the back arrow at the top of the screen) properly redirects the user back to the archive manager.

### DIFF
--- a/src/app/route-history/route-history.service.ts
+++ b/src/app/route-history/route-history.service.ts
@@ -1,6 +1,6 @@
 /* @format */
 import { Injectable, OnDestroy } from '@angular/core';
-import { Event, NavigationEnd, Router } from '@angular/router';
+import { Event, NavigationEnd, NavigationStart, Router } from '@angular/router';
 import { Subscription } from 'rxjs';
 
 @Injectable({
@@ -11,12 +11,19 @@ export class RouteHistoryService implements OnDestroy {
   public previousRoute: string;
 
   private subscription: Subscription;
+  private popstate: boolean = false;
 
   constructor(router: Router) {
     this.subscription = router.events.subscribe((event: Event) => {
+      if (event instanceof NavigationStart) {
+        if (event.navigationTrigger === 'popstate') {
+          this.popstate = true;
+        }
+      }
       if (event instanceof NavigationEnd) {
-        this.previousRoute = this.currentRoute;
+        this.previousRoute = this.popstate ? undefined : this.currentRoute;
         this.currentRoute = event.url;
+        this.popstate = false;
       }
     });
   }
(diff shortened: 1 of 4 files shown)


meisekimiu:
I could not for the life of me figure out how to cover the "popstate" navigation trigger in tests, if anyone has suggestions I'd be happy to cover those lines.

crisnicandrei (approved):

src/app/route-history/route-history.service.ts
```diff
@@ -11,12 +11,19 @@ export class RouteHistoryService implements OnDestroy {
   public previousRoute: string;
 
   private subscription: Subscription;
+  private popstate: boolean = false;
 
   constructor(router: Router) {
     this.subscription = router.events.subscribe((event: Event) => {
+      if (event instanceof NavigationStart) {
+        if (event.navigationTrigger === 'popstate') {
```
slifty, line 19: Would it make sense to combine this into a single if?

On the flip side, if the intention of the nested if is to clearly signal that this function is serving a *variety* of distinct event handlers for different event types (and you're nesting inside), that might be even clearer by creating named handlers (e.g. `handleNavigationStart(event)` and `handleNavigationEnd(end)`) and this handler serves as only a router that calls the more specific event handlers.

All of this is purely for your consideration of course

src/app/route-history/route-history.service.ts
```diff
@@ -11,12 +11,19 @@ export class RouteHistoryService implements OnDestroy {
...
+          this.popstate = true;
```
slifty, line 20: codecov is pointing out this isn't covered by a test -- could it be or is that a big lift?

src/app/shared/components/routed-dialog-wrapper/routed-dialog-wrapper.component.ts
```diff
@@ -63,10 +63,10 @@ export class RoutedDialogWrapperComponent
   }
 
   ngOnDestroy(): void {
+    this.closedByNavigate = true;
```
slifty, line 66: Similarly: can we have this covered by a test?

slifty (approved):
This was *just* a code review as I don't have a working dev env for the front end right now.

I left some thoughts for consideration but they are not blocking / please feel free to merge without making changes!

src/app/route-history/route-history.service.ts
meisekimiu: I tried to figure out how to cover this with a test and from what I saw it seemed impossible to actually simulate a `popstate` event with Angular's test router. I can take a look at it again though.

slifty: If you looked already that's good enough for me!

omnignorant:
Confirmed that this is working as expected! Great job :)

omnignorant (approved):
I followed the test instructions. works as expected. great job!

PR #461, opened 2024-09-17
Load archives/tags properly on login

meisekimiu:
The default (non-MFA) login process does not load up to date archive information as part of its flow. As a result, all logins currently redirect the user to onboarding since the login code doesn't see any archives that belong to the user. Thankfully this doesn't actually create any major issues, since the onboarding page immediately runs a check for archives and redirects the user back to the private workspace.

This does create a few subtle issues though:
  - The default archive is not being set in a way that the TagsService can notice
  - The "Logged in as $user" message is not being displayed

The former issue in particular means that tags are not loaded properly in a standard login. Add a call to "refreshArchives" in the login process so we can properly run the "non-onboarding non-MFA" login routine, and then add a call to `switchToDefaultArchive` in that routine since we know the user has at least one non-pending archive that must be a default. The latter function call also triggers an event that causes the TagsService to load all archive tags correctly.

**Steps to test:**
1. On an account **without** MFA enabled, add some records to the root of the private workspace and add some tags to them.
2. Create a subdirectory in the private workspace, upload a new record into it, and then add a brand new tag onto that record that is not used elsewhere in the archive.
3. Log out and log back in. Try editing the tags of any specific items in the root of the private workspace. Under this branch, the new tag you just created should be visible in the tags list of the metadata editor. Under the main branch, it should not be visible.
- Bonus check: Verify that `getTagsByArchive` is not being called twice in rapid succession in the developer network logs on login

### DIFF
--- a/src/app/auth/components/login/login.component.spec.ts
+++ b/src/app/auth/components/login/login.component.spec.ts
@@ -1,39 +1,178 @@
 /* @format */
 import { NgModule } from '@angular/core';
-import { ActivatedRoute } from '@angular/router';
+import { ActivatedRoute, Router } from '@angular/router';
 import { CookieService } from 'ngx-cookie-service';
 import { Shallow } from 'shallow-render';
 import { LoginComponent } from '@auth/components/login/login.component';
 import { MessageService } from '@shared/services/message/message.service';
 import { TEST_DATA } from '@core/core.module.spec';
 import { AccountService } from '@shared/services/account/account.service';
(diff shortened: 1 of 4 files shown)


meisekimiu:
I will work on addressing logincomponent test coverage (I was trying to avoid it but I must make CodeCov bot happy)

crisnicandrei (approved):

yeslikesolo (approved):

PR #463, opened 2024-09-19
Per 9660 shares from invitation

crisnicandrei:
This PR adds the new view when the user without an account receives an invitation to view someone else's file.

Steps to test:
- Log in into an already existing account and share an item with an an email that is not tied to any Permanent accounts.
- Access the link from the email received on the non-existing user.
- Create an account
- You should see the new view when starting the onboarding process


src/app/onboarding/components/glam/archive-creation-start-screen/archive-creation-start-screen.component.html
```diff
@@ -6,18 +6,48 @@
...
+        that's for you or for your friends, family, interests or organizations.
+      </p>
+      <p>
+        We know that starting this journey can sometimes be overwhelming, but
+        don’t worry. We’re here to help you every step of the way.
+      </p>
+    </div>
+  }
+  @if (hasShareToken) {
+    <div class="text text-share">
```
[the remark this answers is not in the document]
crisnicandrei: @meisekimiu I will create a separate component for it

crisnicandrei:
@meisekimiu do you have other ideas besides the local storage? I was conflicted between that sending the token to the url, but t he url changes from invite to signup to onboarding, so that's why I chose the storage. Should I go with the other option?

crisnicandrei:
@meisekimiu I have added the flow to the onboarding storybook

meisekimiu:
> @meisekimiu do you have other ideas besides the local storage? I was conflicted between that sending the token to the url, but t he url changes from invite to signup to onboarding, so that's why I chose the storage. Should I go with the other option?

I think that the URL makes sense. We would just have to make sure to preserve it between page loads. This is how I envisioned this to work when initially going over the designs with the experience team. That being said, I think maybe we should keep the localStorage (or potentially use sessionStorage) solution in place and see how the experience team wants this to work.

crisnicandrei:
@k8lyn6 what do you think?

k8lyn6:
@crisnicandrei @meisekimiu good question. If I'm following correctly, I think you're asking how long we should remember that someone came from an invitation when signing up. I think Natalie is right -- it might be confusing for a new user to see something about an invitation if it's been weeks since they clicked on that invitation (assuming they use a different method this time to try to sign up). They should probably only see share-related information in onboarding if they came from a share preview page.

This is probably a pretty edge case, but good to consider anyway. The additional share information Tibi designed (badging in the Shares workspace) will be implemented eventually, so even if a new user signs up much later, they should still hopefully see anything that was directly shared with them. 

Let me know if you need further clarification!

src/app/onboarding/components/onboarding/onboarding.component.stories.ts
```diff
@@ -230,3 +261,10 @@ Glam.args = {
   hasInvitations: false,
   isGlam: true,
 };
+
+export const GlamShareInvite: Story = StoryTemplate.bind({});
+GlamShareInvite.args = {
+  accountName: 'Test User',
+  isGlam: true,
+  hasShareToken: true,
```
meisekimiu, line 269:
```suggestion
  hasInvitations: false,
  shareToken: true,
```
This argument name isn't correct and we also probably want to be explicit about turning off invitations in this flow. We also probably want to be explicit about disabling the share token in all other stories.

src/app/onboarding/components/glam/archive-creation-start-screen/archive-creation-start-screen.component.html
```diff
@@ -6,18 +6,23 @@
...
+        store, preserve, and share the digital legacy of all people, whether
+        that's for you or for your friends, family, interests or organizations.
+      </p>
+      <p>
+        We know that starting this journey can sometimes be overwhelming, but
+        don’t worry. We’re here to help you every step of the way.
+      </p>
+    </div>
+  }
+  @if (hasShareToken) {
```
meisekimiu, line 23: I believe we can just use `@else` here. (I know, it's weird to be able to use an `else` statement in an Angular template!)

meisekimiu (commented):
Thank you for extracting out that component. Can you move it into the `glam` directory so it lives alongside all other glam-specific components?

k8lyn6 (changes requested):
Hi Andrei, looking good! I'm going to have @tiberiuVSP review the design on this to make sure that everything looks how we expect it. Tibi, specifically, the font weights look off to me, and we did not account for what the share information should look like when only one item is shared (i.e. what should the icon be, because a folder doesn't make sense). Do you mind checking those two things out as soon as you can? I've added you as a reviewer on this pull request.

Functionally, there are a few changes that need to be made:
- the title of the folder or file should show, not the name of the archive (to the right of the folder name)
- The icon shouldn't be a folder when just one item is shared (Tibi will provide a solution here)
- Once the member has created the archive, the share should be in their Shared workspace. This seems to work for individual files shared, but not for folders. Can you make this change in this ticket? If it becomes too complicated, we can spin it off into another ticket (cc @meisekimiu)
- There is currently a bug that's making QA testing difficult. Specifically, the modal for sign up is broken on the share preview screen, and clicking the "sign up" button in the top right corner is broken. I think for now, the rest of this seems to work okay, so let's address the above feedback and circle back around to those after the bugs are fixed. That being said, Andrei, the expected behavior is that clicking "Sign Up" anywhere on the share preview page should go through this onboarding from a share flow. If that's not something you considered when you built this, could you make that change?

crisnicandrei:
@k8lyn6 what should happen when the user clicks the Sign Up on the top banner? Should the user be redirected to the register page? The same as when I click the "Create my Account" button?

k8lyn6:
@crisnicandrei Yep, if you click the Sign Up button in the top right corner, it should take you to sign up page and you should get the Shares version of the first page of onboarding.

crisnicandrei:
@k8lyn6 it fixed the dialog background, it isnt transparent anymore

k8lyn6:
@crisnicandrei thanks for being proactive on that! I think it would be better for that fix to be in a separate PR, because it's unrelated to this glam onboarding work and we will want to release it quicker than I think this ticket will get merged. [I just made a ticket for that bug.](https://permanent.atlassian.net/browse/PER-9885) Are you able to walk back that change and move it to a new branch?

crisnicandrei:
@k8lyn6 yes. Do you remember how wide the dialog was before? So i can make the change in the new branch?

k8lyn6:
@crisnicandrei I do not. @meisekimiu do you know?

k8lyn6 (changes requested):
The change to use the file/folder name in the title is working. Thanks!

@crisnicandrei have you looked into ensuring that the shared file ends up in the new member's shared workspace? I think that's the only functional change that's left to be made.

crisnicandrei:
@k8lyn6 I did not look because I thought we are waiting for Tibi's design but I will look into it tomorrow.

crisnicandrei:
@k8lyn6 I am having a bit of trouble reproducing the share functionality from before glam(I want to see how it works so I can use it here), do you have any tips? I am not getting any mails back.

k8lyn6 (approved):
Hey, the update to land the new user in their shares workspace with the shared file is perfect! Ahh, such a small change but such a big difference for the user experience!

I did a little shortcut to test this by grabbing a share link rather than inviting an email, and I noticed that the file/folder name and the "shared by" name are not filled in on the first page of onboarding. I think this is a case we didn't think about when we first built this, but since new members could come from just a link rather than be invited, we should resolve this bug. I'm going to make a [new ticket](https://permanent.atlassian.net/browse/PER-9889) for that though and approve this PR. I'll also make a ticket to [change out the folder icon with a file icon](https://permanent.atlassian.net/browse/PER-9890) when @tiberiuVSP has that for you. Great job Andrei!

crisnicandrei:
@meisekimiu can you take a look at this and  if it's alright approve it?

meisekimiu (changes requested):
It looks like there are some failing tests on this, and they're not random test failures but they're associated with the changes. Can you fix these tests?

crisnicandrei:
@meisekimiu fixed the test

meisekimiu (approved):

PR #466, opened 2024-09-23
replaced all tags with keywords

crisnicandrei:
Replaced all tag instances with keywords


meisekimiu (approved):

k8lyn6:
@yeslikesolo the original ticket for this can be found here. 

To test: look for all instances across the platform of the word "tag" or "tags" and make sure it's been changed to "keyword" or "keywords." I would check:
- info tab (metadata tab)
- full screen metadata view
- public archive metadata view
- archive settings that have to do with tags

I may be missing some, so if you could do a quick pass of the app and see if there are any other places where it needs to be changed, that would be great. Thanks!

yeslikesolo (changes requested):
Screenshot 2024-10-10 at 1 45 12 PM

One last place to change: Info Tab

crisnicandrei:
Thanks @yeslikesolo

yeslikesolo (approved):
Looks great!

PR #467, opened 2024-09-26
PER-9611: Preserve line breaks in file viewer

crisnicandrei:
Changed the input type to textarea


meisekimiu (approved):

k8lyn6 (approved):
Hey! This is exciting! Thanks for working on this @crisnicandrei. 

I think this ticket only covers the description. Is that correct? Assuming that is correct, I'll go ahead and mark this PR as approved.

When I was testing, I also tested the other text fields in the archive profile (the "About this archive" field and the description fields on milestones. The milestones preserved line breaks, but the About section on the right side of the public profile did not maintain line breaks. I can make a separate ticket for that!

crisnicandrei:
@k8lyn6 thank you!

PR #468, opened 2024-10-03
PER-9863: Create archive for me - 2 archives

crisnicandrei:
Disable the buttons after submitting the new archive, this way we prevent the user from clicking the buttons twice until the api calls are done


meisekimiu (approved):

k8lyn6 (approved):
This appears to fix the issue. Thanks Andrei!

PR #473, opened 2024-10-10
PER-9865

crisnicandrei:
Added inner html instead of displaying the description as a string

### DIFF
--- a/src/app/public/components/public-archive-web-links/public-archive-web-links.component.html
+++ b/src/app/public/components/public-archive-web-links/public-archive-web-links.component.html
@@ -13,15 +13,18 @@
     >
       {{ item }}
     </a>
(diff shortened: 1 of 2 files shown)


meisekimiu (approved):

k8lyn6:
@omnignorant QA instructions [(see original ticket here)](https://permanent.atlassian.net/browse/PER-9865):

-  add a description to the long archive summary in the profile that includes multiple paragraphs using the enter key
- navigate to the public archive profile
- check that the description on the right side of the archive profile includes line breaks

omnignorant (approved):
Looks good to me! I followed @k8lyn6 suggested testing steps and successfully saw line breaks in the profile text.
Screenshot 2024-10-11 at 12 19 00 PM

omnignorant:
Screenshot 2024-10-11 at 12 38 44 PM
this page too

PR #476, opened 2024-10-17
PER-9881: The location picker never closes

crisnicandrei:
Change the focus event to keydown event so the location picker closes when the user presses done.

Steps to test:

Open the location picker and select a location, then click done.
The location picker should close


src/app/file-browser/components/sidebar/sidebar.component.ts
```diff
@@ -168,6 +168,12 @@ export class SidebarComponent implements OnDestroy, HasSubscriptions {
     }
   }
 
+  onLocationEnterPress(e: KeyboardEvent): void {
+    if (this.canEdit && e.key === 'Enter') {
+      this.editService.openLocationDialog(this.selectedItem);
+    }
+  }
```
meisekimiu, line 175: Can you cover this function with some basic tests? You do not need to cover the rest of the component for now.

meisekimiu (approved):

crisnicandrei:
@meisekimiu it seems codecov still isn't happy, but i will add the other tests myself. Thank you for your help on that one!😄

PR #477, opened 2024-10-21
Fix select input component

meisekimiu:
This component was not dispatching events properly when a select option was picked, which was preventing the country field from saving and working properly in billing settings.

**Steps to test:**
1. Try saving the country in billing info in account settings
2. Verify that it saves properly

### DIFF
--- a/src/app/core/components/billing-settings/billing-settings.component.html
+++ b/src/app/core/components/billing-settings/billing-settings.component.html
@@ -1,44 +1,72 @@
+<!-- @format -->
 <div class="settings-item">
   <label>Full Name</label>
   <div class="settings-item-content">
-    <pr-inline-value-edit [displayValue]="account.fullName" [required]="true" [noScroll]="true" [minLength]="1"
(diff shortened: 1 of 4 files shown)


crisnicandrei (approved):

k8lyn6 (approved):
Looks like this fixes the issue!

As a side note, @meisekimiu , it looks like every time I make a change to a field on this page, it pops up a "Account information saved" toast message in the background. I do think that's a little clunky. I changed every field and it just kept popping up and popping up. I'm not sure if there's a good solution here since every field is saved individually, but wanted to bring it up in case you had any ideas.

meisekimiu:
> As a side note, @meisekimiu , it looks like every time I make a change to a field on this page, it pops up a "Account information saved" toast message in the background. I do think that's a little clunky. I changed every field and it just kept popping up and popping up. I'm not sure if there's a good solution here since every field is saved individually, but wanted to bring it up in case you had any ideas.

I think we can make a separate ticket for this.

PR #481, opened 2024-10-29
Fix CustomOverlayContainer test failures

meisekimiu:
These tests were utilizing the Document object in ways that were not perfectly isolated. While they had some cleanup logic, other tests may manipulate the Document in such a way that make these tests fail. Fix this by allowing tests to inject a node to search for instead of defaulting to `document.body`. The `document.body` node is used in the actual `_createContainer()` method.

Since this is mostly a testing adjustment, this probably does not need manual QA.

### DIFF
--- a/src/app/dialog-cdk/custom-overlay-container.spec.ts
+++ b/src/app/dialog-cdk/custom-overlay-container.spec.ts
@@ -26,31 +26,26 @@ describe('CustomOverlayContainer', () => {
     expect(customOverlayContainer).toBeTruthy();
   });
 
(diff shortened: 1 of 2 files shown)


meisekimiu:
I will be putting up a few other test fix PRs since I identified a few other broken tests while trying to figure out the failures behind this one.

meisekimiu:
One line is missing coverage here but the class has been redesigned so any actual logic is present in the function that has coverage.

liam-lloyd (approved):

PR #485, opened 2024-10-31
PER-9762-update-glam-onboarding

crisnicandrei:
Design changes to the mobile onboarding flw


meisekimiu (approved):

k8lyn6 (approved):
I think this looks good! I also tested non-glam onboarding to make sure that nothing changed there.

PR #486, opened 2024-10-31
PER-9892 Add opt in option to the sign up page

crisnicandrei:
Add checkbox for email opt in option to the signup page

@k8lyn6 I kept the checkboxes in the front, because it was never mentioned to use the iOS looking ones. Should I change them to those?

### DIFF
--- a/src/app/auth/components/signup/signup.component.html
+++ b/src/app/auth/components/signup/signup.component.html
@@ -80,21 +80,15 @@
   </div>
   <div class="auth-options">
     <div class="checkbox-container">
-      <pr-checkbox
-        [isChecked]="receiveUpdatesViaEmail"
-        (isCheckedChange)="receiveUpdatesViaEmail = $event"
-        [variant]="'secondary'"
-      ></pr-checkbox>
       <label class="form-check-label">
         I agree to receive updates via email
       </label>
+      <pr-toggle
+        [isChecked]="receiveUpdatesViaEmail"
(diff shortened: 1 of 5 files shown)


meisekimiu (approved):

k8lyn6:
@crisnicandrei I'd say to do whatever is in the design unless it poses a problem. Do we have those already in the app somewhere? Will it be a lot of effort to use the sliders?

crisnicandrei:
@k8lyn6  we do have them! I just wanted to check with you about this firat. I will change them tomorrow! Thanks!

crisnicandrei:
@k8lyn6 I have changed the checkboxes to toggles!

k8lyn6 (changes requested):
Looking good so far! I noticed that the toggles look a little different when turned off. From the design, it looks like the bubble of the toggle should be more transparent, so you can see it's a toggle. I've attached two screenshots to show the difference. If you have any questions about it, I'd ask Tibi or Natalie.
Screenshot 2024-11-01 at 9 48 41 AM
Screenshot 2024-11-01 at 9 48 31 AM

The API calls look like they're working correctly to Mailchimp and getting tagged correctly. Thanks Andrei!

crisnicandrei:
@k8lyn6 pushed the changes. Thank you!

k8lyn6 (approved):
Yep, that's better. Thanks for the quick change!

PR #488, opened 2024-11-12
Prefer Archivematica access copies over older ones

meisekimiu:
Prefer archivematica access copies over older converted access copies in the file viewer. This is accomplished with a centralized helper function that picks the proper access copy in our specific priority order. This function is then used in various file viewer components.

This also includes the change to the download options, since it depends on the new FileVO type definitions.

Resolves PER-9873 + PER-9902.

**Steps to test:**
1. If the environment you're testing on includes archivematica-generated access copies, then you can actually test this functionality. Make sure to verify new access copies are downloadable as well.
2. If not, verify that unit tests pass and that file viewing for all types (image, document, audio, video, are there others?) is otherwise not affected. Verify that downloading them still functions as well.

### DIFF
--- a/src/app/models/get-access-file.ts
+++ b/src/app/models/get-access-file.ts
@@ -1,3 +1,4 @@
+import { prioritizeIf } from '@root/utils/prioritize-if';
 import { FileFormat, PermanentFile } from './file-vo';
 
 export interface HasFiles {
@@ -9,15 +10,10 @@ function getArchivematicaAccess(files: PermanentFile[]) {
 }
 
 function getPrioritizedConvertedFile(files: PermanentFile[]) {
-  return files
-    .filter((file) => file.format === FileFormat.Converted)
-    .sort((a, b) => {
-      const preferredType = 'pdf';
-      return (
-        Number(b.type.includes(preferredType)) -
-        Number(a.type.includes(preferredType))
-      );
-    })[0];
+  return prioritizeIf(
+    files.filter((file) => file.format === FileFormat.Converted),
+    (file) => file.type.includes('pdf'),
+  )[0];
 }
 
 export function GetAccessFile(record: HasFiles): PermanentFile | undefined {
--- a/src/app/models/get-access-file.ts
+++ b/src/app/models/get-access-file.ts
@@ -1,0 +1,30 @@
+import { FileFormat, PermanentFile } from './file-vo';
+
+export interface HasFiles {
+  FileVOs: PermanentFile[];
+}
+
+function getArchivematicaAccess(files: PermanentFile[]) {
+  return files.find((file) => file.format === FileFormat.ArchivematicaAccess);
+}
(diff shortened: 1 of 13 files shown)


src/app/models/get-access-file.ts
```diff
@@ -0,0 +1,30 @@
...
+}
+
+function getPrioritizedConvertedFile(files: PermanentFile[]) {
+  return files
+    .filter((file) => file.format === FileFormat.Converted)
+    .sort((a, b) => {
+      const preferredType = 'pdf';
+      return (
+        Number(b.type.includes(preferredType)) -
+        Number(a.type.includes(preferredType))
```
meisekimiu, line 18: Is this cursed? I can change this to be more readable, but the potential for this sorting function to be a single subtraction was irresistible. I'm using the `Number` constructor here because `+x - +y` seemed even more confusing.

liam-lloyd: I think this is reasonable

liam-lloyd (commented):
In testing this, it looks like the Archivematica access copy is available via the download button in the record info panel on the right of the screen, but not via the download button at the top of the file list.

meisekimiu:
> In testing this, it looks like the Archivematica access copy is available via the download button in the record info panel on the right of the screen, but not via the download button at the top of the file list.

I believe this should be fixed now

src/app/models/record-vo.ts
```diff
@@ -155,6 +156,20 @@ export class RecordVO
...
     }
   }
+
+  public getDownloadOptionsList() {
+    const files = [...(this.FileVOs ?? [])];
+    return files
+      .sort(
+        (a, b) =>
+          Number(b.format === FileFormat.Original) -
+          Number(a.format === FileFormat.Original),
```
meisekimiu, line 166: I ended up using this same pattern in this method. Again, despite its cursed appearance I think this is more flexible than previous implementations method for getting the original file, which tries to find it manually and copy it to the top of the array.

meisekimiu:
I decided to refactor the "cursed" pattern I was using earlier since it ended up being used in two places now. This should make those calls easier to read while allowing us to continue to use this same prioritization behavior in each case, and hopefully future cases if they arise.

liam-lloyd (approved):

k8lyn6:
@yeslikesolo to test this, please check all formats listed as accepted file types in [this KB article](https://permanent.zohodesk.com/portal/en/kb/articles/file-types-that-we-accept-and-support) to make sure they are 1. viewable (in both the preview info tab and full screen!) and 2. downloadable (from the button at the top and the button in the info tab). If you need test files, [you can find some here.](https://drive.google.com/drive/folders/14V5FrH3uF855u41DBuGiK__DD8us6WXk?usp=drive_link)

yeslikesolo (approved):

PR #490, opened 2024-11-18
Per 9261 search by tag public archive

crisnicandrei:
This pull request adds the public tag search functionality to the app.

Steps to test:
1. Go to a public archive
2. Type a tag's name in the search bar
3. A dropdown will appear
4. Click on one of the tags
5. The search should be working

### DIFF
--- a/src/app/shared/services/api/search.repo.ts
+++ b/src/app/shared/services/api/search.repo.ts
@@ -70,14 +70,23 @@ export class SearchRepo extends BaseRepo {
   ) {
     const endpoint = '/search/folderAndRecord';
 
-    const tagsQuery = tags
-      .map((tag, index) => `tags[${index}][tagId]=${tag.tagId}`)
-      .join('&');
+    const tagsDict = tags.reduce((obj, tag, index) => {
+      obj[`tags[${index}][tagId]`] = tag.tagId;
+      return obj;
+    }, {});
+
+    console.log(tagsDict);
+
+    let tagsQuery = '';
+
+    Object.keys(tagsDict).forEach((key) => {
+      tagsQuery += `&${key}=${tagsDict[key]}`;
+    });
 
     const queryString =
       `query=${query}&archiveId=${archiveId}&publicOnly=true` +
       (limit ? `&numberOfResults=${limit}` : '') +
-      (tagsQuery ? `&${tagsQuery}` : '');
+      (tagsQuery ? `${tagsQuery}` : '');
 
     return getFirst(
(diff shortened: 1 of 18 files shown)


meisekimiu (changes requested):
Left some comments, but not all of them need to be explicitly addressed (some of them are more pieces of advice in general and do not need to be specifically applied if they aren't applicable).

crisnicandrei:
@meisekimiu did the suggested changes. Thanks a lot for your suggestions!

src/app/shared/services/api/search.repo.ts
```diff
@@ -68,19 +68,27 @@ export class SearchRepo extends BaseRepo {
...
+      `query=${query}&archiveId=${archiveId}&publicOnly=true` +
+      (limit ? `&numberOfResults=${limit}` : '') +
+      (tagsQuery ? `&${tagsQuery}` : '');
 
     return getFirst(
-      this.httpV2.get<SearchResponse>('/search/folderAndRecord', data, null, {
+      this.httpV2.get<SearchResponse>(`${endpoint}?${queryString}`, {}, null, {
         authToken: false,
       }),
     );
```
meisekimiu, line 86: This is a bit on me for using this code as an example of how we can use `const` in most places in JavaScript, but I didn't actually want my example to be directly used here. As mentioned in [this comment,](https://github.com/PermanentOrg/web-app/pull/490#discussion_r1848600610) we don't actually have to manually build up the query string as the HttpV2Service can take in a data object like a POST request, and generate and append a query string automatically.

My `const` suggestion was more a standard JavaScript tip, though it still had applicability in this function. Instead of building up a query string with `const`s, we should be building up a `const` data object we pass into the `httpV2.get` method. And since the parameters for this method are somewhat complex, we probably have to compose this data object out of multiple `const`s, or at least inlined expressions.

```ts
// I have not tested this code!
// This is just a sketch of what the code might look like.
// It definitely needs adjustments for typescript!

const tagQueries = tags.reduce((obj, tag, index) => {
    obj[`tags[${index}][tagId]`] = tag.tagId;
    return obj;
}, {});

const data = {
query,
archiveId,
publicOnly: 'true'
...tagQueries
};

if (limit) {
data.numberOfResults = limit;
}

return getFirst(
this.httpV2.get<SearchResponse>('/search/folderAndRecord', data, null, {
        authToken: false,
      })
);
```

meisekimiu (changes requested):

crisnicandrei:
@meisekimiu Thanks! Pushed the changes!

yeslikesolo (approved):
Search my tags working great, thanks Andrei!

PR #492, opened 2024-11-20
design changes to the create archive for me screen

crisnicandrei:


meisekimiu (approved):

k8lyn6 (changes requested):
Looks good! Small thing: the archive icon is broken for me on this page. I've attached a screenshot. Could you look into this? I tried on both Chrome and Firefox.
Screenshot 2024-11-26 at 1 29 25 PM

crisnicandrei:
@k8lyn6 the icons work well for me locally! Which is the env you are testing on and how can I access it? It seems there might be a problem with how the image paths are handled on it so I would like to take a look, as this is not the first time it happens to the Experience Team

k8lyn6:
@crisnicandrei I am testing on the `dev` environment, which you can access at [dev.permanent.org/app/auth/login](dev.permanent.org/app/auth/login) (you can just create a test account there).

k8lyn6 (approved):
I think we made the above issue a separate ticket so I'm approving this PR!

PR #493, opened 2024-11-21
Remove the old welcome dialog and invitation dialog components

crisnicandrei:
I have removed the old welcome dialogs because with the new onboarding they are not needed anymore.

Steps to test:
1. Create a new account (with or without invitation)
2. Go through the onboarding process
3. When done, it should redirect you to the private workspace, without any dialog popping up


crisnicandrei:
@meisekimiu Thank you for your explanation! I have reverted the changes

meisekimiu (approved):

k8lyn6:
@yeslikesolo For testing, go through Glam onboarding (using multiple ways, such as archive member, etc) and make sure the final pop up over the workspace doesn't show up. Then, test non-Glam and make sure it _does_ pop up.

yeslikesolo (approved):

PR #494, opened 2024-11-27
Per 9929 show password strength

crisnicandrei:
This pull request displays the strength of a user's password when registering a new account.

Steps to test: 
1. Go to the register page
2. Start inputting a password
3. A message will appear below with the strength of it


src/app/auth/components/signup/signup.component.html
```diff
@@ -112,7 +117,12 @@ <h1 class="page-title" *ngIf="isForShareInvite">
       [icon]="'/auth/sign-up'"
       [orientation]="'right'"
       [style]="{ justifyContent: 'center' }"
-      [disabled]="waiting || signupForm.invalid || !agreedTerms"
+      [disabled]="
+        waiting ||
+        signupForm.invalid ||
+        !agreedTerms ||
+        passwordStrengthMessage !== 'Strong'
```
meisekimiu, line 124: I'm not sure if we want to block people from signing up just based on password strength. The intention is to show an indicator so that people can see their password strength and make their own decision on if they want to pick a stronger password.

src/app/auth/components/signup/signup.component.scss
```diff
@@ -119,3 +119,24 @@ form {
...
+.strength-weak {
+  color: orange;
+}
+
+.strength-medium {
+  color: blue;
+}
+
+.strength-strong {
+  color: green;
```
meisekimiu, line 136: This is not an request for changes immediately but we probably want some design/accessibility review here for color choices that match Permanent's theme and also have proper contrast against the background.

src/app/auth/components/signup/signup.component.ts
```diff
@@ -110,14 +114,55 @@ export class SignupComponent {
...
-
     const confirmPasswordControl = new UntypedFormControl('', [
       Validators.required,
       matchControlValidator(this.signupForm.controls['password']),
     ]);
     this.signupForm.addControl('confirm', confirmPasswordControl);
   }
 
+  ngOnInit(): void {
+    this.signupForm.controls['password'].valueChanges.subscribe((password) => {
```
meisekimiu, line 125: Does this subscription get automatically cleaned up? It might, and I'd love confirmation that it does, but I feel like subscriptions have occasionally caused errors in the application once the things they refer to have been destroyed.

src/app/auth/components/signup/signup.component.ts
```diff
@@ -49,6 +50,9 @@ export class SignupComponent {
   agreedTerms = false;
   receiveUpdatesViaEmail = false;
 
+  passwordStrengthMessage: string = '';
+  passwordStrengthClass: string = '';
```
meisekimiu, line 54: In theory we can make this typing even more specific with a type union. That way we know that `Message` can only ever be "Strong", "Weak", etc.

In this case it might be a bit redundant but it's nice to keep that in mind, especially when testing for string equality like we do in this component.

crisnicandrei:
@meisekimiu thanks! I will do the changes! @k8lyn6 can you please suggest some color values for the messages?

k8lyn6:
@crisnicandrei we can look at it tomorrow during our accessibility review. cc @yeslikesolo

meisekimiu (commented):
We've decided this probably needs more design work before it's ready to be released. We probably don't want to erase the work already done though, so can you hide this behind a feature flag, using the `FeatureFlagService` to test for its presence?

crisnicandrei:
@meisekimiu for sure!

src/app/auth/components/signup/signup.component.ts
```diff
@@ -107,17 +123,87 @@ export class SignupComponent {
...
       matchControlValidator(this.signupForm.controls['password']),
     ]);
     this.signupForm.addControl('confirm', confirmPasswordControl);
   }
 
+  ngOnInit(): void {
+    this.passwordSubscription = this.signupForm.controls[
+      'password'
+    ].valueChanges.subscribe((password) => {
+      if (this.featureFlagService.isEnabled('passwordStrngth')) {
```
meisekimiu, line 146:
```suggestion
      if (this.featureFlagService.isEnabled('password-strength')) {
```

There's a typo in this flag name. We might also want to use "kebab-case" for feature flags. I think it might be the most accessible to people on the experience team, though we don't have a defined standard yet so it might be fine to leave it like this.

As a small note, I like naming a component's instance of the FeatureFlagService just `feature` so that the code can read nicely: `this.feature.isEnabled('feature-name')`, but it doesn't really matter what it's named and it's easy to refactor later.

k8lyn6:
@meisekimiu can you tag me in a comment when this is ready for QA?

meisekimiu (approved):
@k8lyn6 this is ready for QA now. In theory the only step should be to verify it is controlled by the feature flag and doesn't appear by default.

k8lyn6 (approved):
Sorry for the confusion with this ticket Andrei! I shouldn't have put it in the front-end-backlog without a design first. The work you did is great and as soon as Tibi has a design for us, we can pick it back up.

The feature flag seems to be working as designed. When this branch is deployed, I no longer see the password strength information.

crisnicandrei:
@k8lyn6 I have pushed the changes with tibi's new design. I also have found a bug while working on this so I will create a ticket for that! Thanks!

k8lyn6:
@meisekimiu before I do a review, was this branch already merged? Should there be a separate PR for additional work? Just wanted to check before proceeding.

crisnicandrei:
@k8lyn6 it was not merged

meisekimiu (approved):
I have approved this already, with the exception of the linting and prettier changes which don't really need to be checked again.

src/app/shared/components/strength-meter/strength-meter.component.ts
```diff
@@ -0,0 +1,50 @@
...
+  bars: string[];
+}
+
+@Component({
+  selector: 'pr-strength-meter',
+  templateUrl: './strength-meter.component.html',
+  styleUrls: ['./strength-meter.component.scss'],
+})
+export class StrengthMeterComponent implements OnChanges {
+  progressBars: string[] = ['', '', ''];
```
meisekimiu, line 15: **This is not a blocker so feel free not to address it.**

I think this implementation is a bit loose, while also being a bit too coupled to the output HTML I guess? If we break down this component it really has 4 possible states: `null`, `weak`, `medium`, and `strong`. But the way the state is represented internally in the component allows for at least 12* states to be represented in the progress bars alone. While something like `['weak', '', '']` is a normal state, there's nothing in the code that prevents us from having a weird nonsensical state like `['','strong','half']`. It also means that in tests we have to express our assertions with these arrays, even though we're logically only testing for those 4 possible states. These tests have to know about the internal details of our component instead of testing more abstract state behavior.

<sub>*Technically since progressBars is only of type `string[]`, it's possibility of states is near infinite!</sub>

I think ideally the internal state of the application should try to be consistent with how we'd map it out logically. It helps prevent [unintended errors or states from appearing](https://en.wikipedia.org/wiki/Weird_machine) and also just makes these components easier to work with. Instead of having to know what specific array values we need to set, we can just set `this.strength = PasswordStrength.strong` or something like that (it doesn't need to be an enum, but I used an enum since that's typically the tool used to convey a single state of many that an object can be in).

meisekimiu (changes requested):
I didn't realize that this incorporates new design elements and therefore needs a more thorough review. The idea behind hiding the previous changes behind a feature flag is that it allowed us to merge those changes to main and address the new design elements in their own ticket. In general it's preferable to do that so that tickets can associated with their own pull requests whenever possible, changes are visible to other developers sooner, there are less merge conflicts when integrating more frequently, and it just feels good to be able to close pull requests faster (as it stands right now, we now have to do review for a pull request opened in November, which isn't great).

It feels weird at first, but the whole point of feature flags is that they allow us to merge code for a feature that is incomplete without users seeing it. This lets us make smaller pull requests that can be integrated more frequently, since the feature doesn't have to be 100% done to warrant merging it. See https://martinfowler.com/articles/feature-toggles.html for more details. I encourage us all to take advantage of feature flags to merge things earlier and split things into multiple small tickets.

crisnicandrei:
@meisekimiu great points! And thank you for the article! I will address this!

src/app/auth/components/signup/signup.component.ts
```diff
@@ -23,10 +22,13 @@ import {
 } from '@models';
 import { DeviceService } from '@shared/services/device/device.service';
 import { GoogleAnalyticsService } from '@shared/services/google-analytics/google-analytics.service';
+import { passwordStrength } from 'check-password-strength';
```
meisekimiu, line 25:
```suggestion
```
We can get rid of this import.

src/app/auth/components/signup/signup.component.ts
```diff
@@ -23,10 +22,13 @@ import {
...
 
 const MIN_PASSWORD_LENGTH = APP_CONFIG.passwordMinLength;
 const NEW_ONBOARDING_CHANCE = 1;
 
+type PasswordType = '' | 'Too Weak' | 'Weak' | 'Medium' | 'Strong';
+
```
meisekimiu, line 31:
```suggestion
```
This can probably go too, right?

src/app/shared/components/strength-meter/password-strength.ts
```diff
@@ -0,0 +1,70 @@
...
+  styleUrls: ['./password-strength.scss'],
+})
+export class PasswordStrengthComponent implements OnChanges {
+  progressBars: string[] = ['', '', ''];
+
+  public strength: PasswordStrength = PasswordStrength.Null;
+  public message: string = '';
+  public passwordClass: string = '';
+  @Input() password: string = '';
+  public enabledPasswordCheckStrength: boolean;
```
meisekimiu, line 25: There's a bit of style inconsistency here. Generally we like to visually separate the Input() fields from standard class properties. And some of these have access modifiers (something I generally like to see) while some don't.
```suggestion
  @Input() public password: string = '';
  
  public progressBars: [string, string, string] = ['', '', ''];
  public strength: PasswordStrength = PasswordStrength.Null;
  public message: string = '';
  public passwordClass: string = '';
  public enabledPasswordCheckStrength: boolean;
```

src/app/shared/components/strength-meter/password-strength.ts
```diff
@@ -0,0 +1,70 @@
...
+    }
+  }
+
+  private setStrength(strengthId: number): void {
+    switch (strengthId) {
+      case 0:
+        this.strength = PasswordStrength.Weak;
+        this.message = 'weak';
+        this.passwordClass = 'too-weak';
+        this.progressBars = ['weak', '', ''];
```
meisekimiu, line 44: This is just a general piece of advice. This switch statement is rather verbose, isn't it? This is at least partly due to the fact that while we're now using an enum to declare password strength, we still have to set all of those other values alongside it. We could clear up this code quite a bit if we could derive those values from our `this.strength` value with something like a function call in the typescript layer or a pipe in the template layer.

crisnicandrei:
@meisekimiu did the changes

k8lyn6 (commented):
Hey @crisnicandrei I'm not seeing any password strength indications at all when I deploy this branch. Is there something else I need to do? @meisekimiu do I need to activate the feature flag?

crisnicandrei:
@k8lyn6 yes, the feature flag must be active

k8lyn6:
@crisnicandrei what is the name of the feature flag? I don't see it in the retool app (I switched the environment to dev)

crisnicandrei:
@k8lyn6 it is 'password-strength', but I did not set it in retool, do I have to do that @meisekimiu ? This is the first time using the feature flag so I am unsure

crisnicandrei:
@k8lyn6 what do you mean about a copy check? I have looked it up online but found nothing so I am a bit confused

k8lyn6:
My apologies for the confusion @crisnicandrei ! By copy, I mean the words and language we're using in the design (such as "Your password is strong!"). Normally, in the design process, the last thing I do with Tibi is a review of the copy so that I don't have to ask you to re-code any of the language in the design after the fact. I wasn't able to do an accessibility or copy review before the implementation, which is why we're changing up the language on the fly. 

If we could swap out the text that's currently implemented with what I wrote in my review, that would be great. I think that's the only change I have for now, but Han, Natalie, and I will need to do an accessibility review of the color contrast. Because this is behind a feature flag, you should be able to merge the changes after we change the text and it won't go live on production yet. We can make additional tickets for any changes after the accessibility review.

crisnicandrei:
@k8lyn6 I have updated the text

k8lyn6 (approved):
Thanks for the update @crisnicandrei ! I think we're good to merge this PR, but not turn the feature on on prod yet (since it's still hidden by a feature flag). We'll do an accessibility review and add any additional tickets that are necessary afterwards.

cc @meisekimiu

PR #497, opened 2024-12-04
Add background to FamilySearch Import Modal

meisekimiu:
Add a background to the FamilySearch Import Modal, since it only had a transparent background. This was a style breakage caused by our migration to thew new CDK dialog.

Ideally this would be testable in Storybook but these components were not designed to be mocked very easily and efforts to put them in a Storybook story only ended in strange errors.

**Steps To Test:**
1. Go to apps and connect to FamilySearch
2. Import a family tree
3. Verify that all modals, in all steps of the process, have a background

### DIFF
--- a/src/app/apps/components/family-search-import/family-search-import.component.html
+++ b/src/app/apps/components/family-search-import/family-search-import.component.html
@@ -2,72 +2,94 @@
   <div class="dialog-body">
     <div class="page-title">Select Person(s) To Import</div>
     <p>
-      <span>Choose members of your family tree to create Permanent Archives for:</span>
+      <span
+        >Choose members of your family tree to create Permanent Archives
+        for:</span
+      >
     </p>
-    <div class="family-member" *ngFor="let family of familyMembers">  
-      <div class="select" [ngClass]="{'exists': family.permExists}">
-        <label>
-          <input type="checkbox" [(ngModel)]="family.isSelected" [disabled]="family.permExists">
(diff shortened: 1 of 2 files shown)


liam-lloyd (approved):

yeslikesolo (commented):
I've tried connecting a familysearch account and keep running into this screen. I've tried making new permanent accounts on dev and link the new account, but each time I end up at this same screen. 

image

k8lyn6 (changes requested):
Hey @meisekimiu it looks like the modal doesn't scroll and goes off the page right now, at least on Chrome. I'm not able to see the full list, and there's no "submit" button or anything. See screenshot. I'd love for this to go out in tomorrow's deployment if we're able to do it, but it's okay if not!
Screenshot 2024-12-04 at 5 58 13 PM

@yeslikesolo once Natalie has made changes, can you take over QA for this? You have to deploy to staging because we can't test FamilySearch on dev. To deploy to staging, you build the branch just like you would on dev, just in the "Staging Build" tab. Then you'll have to manually deploy it yourself on the infrastructure page. Ping me if you have questions about it!

meisekimiu:
Screenshot from 2024-12-05 09-12-45
I added a scrollbar to the middle of the modal so that the "Select Person(s) To Import" text and the buttons are always visible even if there's a long list of members.

yeslikesolo (changes requested):
Changes approved! Sorry I clicked the wrong button >.<

yeslikesolo (approved):

PR #498, opened 2024-12-04
Call v2 version of registerRecord on all uploads

meisekimiu:
We previously only used this for multipart uploads. Migrate to the V2 version of this endpoint.

**Steps to test:**
1. Verify that uploads of files under 50MB in size still work properly.


k8lyn6 (approved):
Uploaded a few files ranging from a couple of KB to a couple of MB and they all uploaded just fine!

PR #501, opened 2024-12-06
PER-9916-exiting-onboarding-on-refresh

crisnicandrei:
Because of a user having no archive during onboarding, the onboarding process would throw an error each time the page would refresh, loggging the user out.

I have fixed that problem and added the current steps of onboarding to the session storage so that the user can continue where they left off after refreshing, but if the tab is closed all the progress is being erased

Steps to test:
1. Create a new archive
2. Go to the onboarding page
3. Refresh and the user should still be logged in
4. Go through each process, refreshing each itme, and the data should persist


meisekimiu (approved):

k8lyn6 (approved):
Awesome! It works perfectly. Thanks Andrei!

PR #502, opened 2024-12-09
Integrate Glam Onboarding Feature Flag

meisekimiu:
Instead of using `localStorage`, Glam onboarding is now configured with a feature flag defined in the environment's database. Also let the Feature Flag service use environment variables as fallback to make it easier to test them in local or other ephemeral environments.

**Steps to Test:**
1. Sign up in an environment with the feature flag disabled and verify that old onboarding is presented to the user
2. Enable the feature flag and verify that glam onboarding is now shown instead

To test this in local, you can add `glam-onboarding` to `optional-secrets` and `.env`, and restart the development server whenever the flag is changed. An empty string (or absent flag) is considered a disabled feature flag.

### DIFF
--- a/load-secrets.js
+++ b/load-secrets.js
@@ -25,15 +25,15 @@ for (const secret of optionalSecrets) {
 if (missingSecrets.length) {
   console.error(
     `ERROR: Missing the following environment variables: ${missingSecrets.join(
-      ','
-    )}`
+      ',',
+    )}`,
   );
   process.exit(1);
 }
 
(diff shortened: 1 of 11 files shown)


crisnicandrei (approved):

meisekimiu:
This should be ready for QA now. New changes did add some uncovered code, but as I've said before, module constructors are essentially the `main()` part of an Angular application so they don't really need to be tested.

yeslikesolo (approved):

PR #503, opened 2024-12-11
PER-9943-cant-delete-sms-mfa

crisnicandrei:
Handle 11 character phone numbers(with country code) in SMS MFA setup

Steps to test:
1. Have an account with two factor auth using sms with county code
2. Go to settings to disable the method
3. Click the delete button next to the method
4. The form should now work as before

### DIFF
--- a/src/app/core/components/two-factor-auth/two-factor-auth.component.ts
+++ b/src/app/core/components/two-factor-auth/two-factor-auth.component.ts
@@ -55,11 +55,32 @@ export class TwoFactorAuthComponent implements OnInit {
 
   formatPhoneNumber(value: string) {
     let numbers = value.replace(/\D/g, '');
-    let char = { 0: '(', 3: ')  ', 6: ' - ' };
-    let formatted = '';
+    let countryCode = '';
+
+    if (numbers.startsWith('00')) {
+      const match = numbers.match(/^00(\d{1,3})/);
+      if (match) {
+        countryCode = `+${match[1]}`;
+        numbers = numbers.substring(match[0].length);
+      }
+    } else if (numbers.startsWith('1') && value.startsWith('+')) {
+      countryCode = '+1';
+      numbers = numbers.substring(1);
+    } else if (value.startsWith('+')) {
+      const match = numbers.match(/^(\d{1,3})/);
+      if (match) {
+        countryCode = `+${match[1]}`;
+        numbers = numbers.substring(match[1].length);
+      }
+    }
+
+    const char = { 0: '(', 3: ') ', 6: '-' };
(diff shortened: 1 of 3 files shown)


src/app/core/components/two-factor-auth/two-factor-auth.component.ts
```diff
@@ -55,11 +55,32 @@ export class TwoFactorAuthComponent implements OnInit {
...
+      }
+    } else if (numbers.startsWith('1') && value.startsWith('+')) {
+      countryCode = '+1';
+      numbers = numbers.substring(1);
+    } else if (value.startsWith('+')) {
+      const match = numbers.match(/^(\d{1,3})/);
+      if (match) {
+        countryCode = `+${match[1]}`;
+        numbers = numbers.substring(match[1].length);
+      }
```
meisekimiu, line 74: Can you add coverage for these paths of code? The other ones have examples provided in the tests and it would be nice to see that spelled out for this branch as well.

crisnicandrei:
@meisekimiu you are right, test added :)

meisekimiu (approved):

k8lyn6 (commented):
A few things:

1. @meisekimiu I'm not entirely sure how to test the actual bug for this, because I no longer have an account set up with a SMS 2FA number with the country code. I'm also not able to set that up directly in the webapp because we don't allow country codes. Is there a way to do this in FusionAuth, that you're aware of?
2. @crisnicandrei I did notice that when I try to add a new SMS 2FA method, I enter my phone number but the button to submit the form to request a code requires 13 characters before it becomes active. It should require only 10 numerical characters. This appears to be a new change as this is not the case on prod. This might solve the issue to point #1 above, though, if I am able to submit a phone number here that looks like #-###-###-####.

crisnicandrei:
@k8lyn6 I did not understand the last part. Should the button become active after 10 characters?

k8lyn6:
@crisnicandrei yes, US numbers are ten digits long (in the format XXX-XXX-XXXX). I'm not sure if you adjusted the form to allow for a few extra numbers, which I think is okay to solve the bug reported, but it should also be able to accept 10 digits. Sorry for the confusion!

crisnicandrei:
@k8lyn6 It now allows 10 digit long numbers :)

meisekimiu:
> 1. Is there a way to do this in FusionAuth, that you're aware of?

I'm not sure, but you might be able to override that as an admin in FusionAuth dev.

k8lyn6 (approved):
The bug itself is proving challenging to QA test, because I'm not able to get my account into that state anymore and it's difficult to search for an account currently in that state. That being said, 2FA functionality seems to work as expected on this pull request, so I'm going to approve this.

PR #504, opened 2024-12-11
Build and Upload Storybook in Github Workflows

meisekimiu:
To make Storybook easier to access for non-developers, we want to build a static version of it that gets deployed to each environment. This requires building a copy of Storybook alongside the main production build and then uploading that static Storybook directory as well. This is done by creating a new `build-storybook` step. Both Storybook and the main production build are uploaded in the `upload-code` step.

Also update .gitignore to ignore the statically built Storybook directory.

**Steps to test:**
1. Run a build of this branch and deploy it to a test environment (the deploy is done automatically on dev)
2. After the deployment is done, ssh into the backend box for the environment and check that `/data/www/mdot` contains a `storybook` directory alongside its usual `dist` directory.
3. Check that the `storybook` directory has built HTML/JavaScript files in it.

This does not need any QA.

### DIFF
--- a/.github/workflows/build-dev.yml
+++ b/.github/workflows/build-dev.yml
@@ -42,6 +42,34 @@ jobs:
           name: dist
           path: dist
 
+  build-storybook:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "18"
+
+      - uses: actions/cache@v4
+        with:
+          path: node_modules
+          key: ${{ runner.os }}-node_modules-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node_modules-
+
+      - name: Install dependencies
+        run: npm install
+
+      - name: Build Storybook
+        run: npm run build-storybook
+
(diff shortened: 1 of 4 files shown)


.github/workflows/build-dev.yml
```diff
@@ -42,6 +42,34 @@ jobs:
...
+
+      - name: Build Storybook
+        run: npm run build-storybook
+
+      - name: Archive `storybook-static`
+        uses: actions/upload-artifact@v4
+        with:
+          name: storybook
+          path: storybook-static
+
```
liam-lloyd, line 72: I'd recommend factoring this (and the `build` job) into their own files and invoking them from each of the deploy actions via
```
jobs:
  build:
    uses: ./.github/workflows/build.yml
```

It looks to me like the artifact upload/download will just continue to work in that case, though I haven't tested this myself.

liam-lloyd (approved):
Confirmed that this does work, and I think it's up to you whether you want to factor out the build jobs here or leave that for the potholes sprint.

meisekimiu:
I'll merge this now and refactor it during the Potholes sprint since it seems a bit involved (and multiple open PRs have edits to the workflows right now and I'd prefer to merge these changes as soon as possible)

PR #506, opened 2024-12-12
Add secrets to env when building storybook

meisekimiu:
I forgor to actually supply required secrets. Thankfully, these tokens aren't actually used, they just need to be defined at all so we don't even need to supply an actual token for these.

PER-9954: Generate secrets as part of Storybook build

### DIFF
--- a/.github/workflows/build-dev.yml
+++ b/.github/workflows/build-dev.yml
@@ -63,6 +63,10 @@ jobs:
 
       - name: Build Storybook
         run: npm run build-storybook
+        env:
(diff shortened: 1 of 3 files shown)


meisekimiu:
Since I messed with dev's apache config and the dev load balancer, you can actually test that works by verifying no stories have weird errors (especially the onboarding demo or directive stories) on the instance of Storybook deployed to dev under this branch. (At `/storybook` on dev)

liam-lloyd (approved):

PR #508, opened 2024-12-13
Pass client into PledgeData

meisekimiu:
We can specify what client is making a storage transaction now. Send this data to donations-firebase so it can write a nice description.

**Note:** This does nothing and potentially causes errors if [the corresponding donations-firebase changes](https://github.com/PermanentOrg/donations-firebase/pull/81) are not deployed. Do not merge this until it that is merged.

**Steps to test:**
(Ensure that the corresponding donations-firebase changes are deployed first)
1. Under a test environment (local/dev/staging), make a storage transaction with a Stripe test card.
2. Verify the transaction was successful with no errors.
3. Log into Stripe and view the Transactions dashboard with Test Mode enabled. The description field for your transaction should say "Web App Storage Transaction" instead of just listing an ID.


liam-lloyd (approved):

k8lyn6 (approved):
Ahhh! Exciting! A small change that will go a long way. Looks great to me!

PR #511, opened 2025-01-09
PER-9957: Add sharing options to the sharing dialog

crisnicandrei:
Added a text which explains the sharing options in the sharing dialog for the user. Also updated the scss file.


meisekimiu (approved):

k8lyn6 (changes requested):
Looking great! Would it be possible to reduce the font size to the small font size we use in the Legacy Planning modal pictured below. Otherwise I think this is good!

Screenshot 2025-01-09 at 4 54 01 PM

crisnicandrei:
@k8lyn6 of course! I will!

k8lyn6 (approved):
That's perfect, thanks Andrei!

PR #513, opened 2025-01-13
PER-9959: Fix public search

crisnicandrei:
Fix the search by adding the limit only if it is defined

Steps to test:
Search for a substring in the public search bar
Click view all results
it should display all the records and folders accordingly

### DIFF
--- a/src/app/shared/services/api/search.repo.ts
+++ b/src/app/shared/services/api/search.repo.ts
@@ -77,9 +77,12 @@ export class SearchRepo extends BaseRepo {
       query,
       archiveId,
       publicOnly: true,
-      numberOfResults: limit,
       ...tagsDict,
     };
+
+    if (limit) {
+      data['numberOfResults'] = limit;
+    }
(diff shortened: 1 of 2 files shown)


src/app/shared/services/api/search.repo.ts
```diff
@@ -77,10 +77,13 @@ export class SearchRepo extends BaseRepo {
       query,
       archiveId,
       publicOnly: true,
-      numberOfResults: limit,
       ...tagsDict,
     };
 
+    if (limit) {
+      data['numberOfResults'] = limit;
```
meisekimiu, line 84: We can and probably should avoid accessing this property with a string by using an explicit type definition for `data` instead.

meisekimiu (approved):

k8lyn6 (approved):
This appears to solve the issue! I can see search results now if I hit enter in the search bar, and when I click "View all results." Great work, Andrei!

PR #514, opened 2025-01-13
PER-9961: Fix public search thumbnails size

crisnicandrei:


meisekimiu (approved):

k8lyn6 (approved):
I think this looks much better. Thanks for the update!

PR #515, opened 2025-01-16
PER-9960 Clicking on tag should search the result

crisnicandrei:
Created a new route for clicking on tag. This way I can handle the tag scenario in the public search results component

Steps to test: 
1. Go to a public archive
2. Search for a tag
3. Click on the tag results
4. It should display all the search results by that tag

### DIFF
--- a/src/app/public/components/public-archive/public-archive.component.spec.ts
+++ b/src/app/public/components/public-archive/public-archive.component.spec.ts
@@ -4,6 +4,7 @@ import { PublicModule } from '@public/public.module';
 import { Shallow } from 'shallow-render';
 import { of } from 'rxjs';
 import { PublicProfileService } from '@public/services/public-profile/public-profile.service';
+import { Router } from '@angular/router';
 import { PublicArchiveComponent } from './public-archive.component';
 
 const publicProfileServiceMock = {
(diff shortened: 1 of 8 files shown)


meisekimiu:
Can you improve the patch coverage of this pull request so that the codecov check passes?

crisnicandrei:
@meisekimiu fixed the coverage

src/app/public/components/public-archive/public-archive.component.spec.ts
```diff
@@ -54,4 +56,34 @@ describe('PublicArchiveComponent', () => {
...
+    const router = inject(Router);
+    spyOn(router, 'navigate');
+
+    instance.archive = { archiveId: '123' } as any;
+
+    instance.onHandleSearch('test-query');
+
+    expect(router.navigate).toHaveBeenCalledWith(
+      ['search', '123', 'test-query'],
+      { relativeTo: instance.route },
```
meisekimiu, line 71: I'm not sure we should be exposing the `ActivatedRoute` of this component publicly just for the sake of unit testing convenience. Can you try constructing the `router.navigate` test double in such a way where we can ignore that other parameter? I don't think we need to be so granular as to test the existence of the `relativeTo` property.

meisekimiu (approved):

k8lyn6:
@yeslikesolo some QA pointers:
- Ensure that you have an archive with published materials. Ideally, try multiple nested folders with different kinds of keywords added (both tags and custom metadata would be best)
- Navigate to the public gallery for that archive and search for the tags you created
- They should show up in the search results. Click on a tag to test it
- What do the results look like? Are they accurate? 
- Try searching while navigated into folders in the public archive. Does this work as expected?
- Try searching for non-keyword text (such as a description or file name). Does this functionality still work?

yeslikesolo (approved):

PR #519, opened 2025-01-22
Adjust sign in/sign up page to support announcements

meisekimiu:
We want to display a downtime banner soon, but the new "glam" auth page does not support showing the banner. Well, it does, but it doesn't account for it in the layout and some flexbox nightmares manifest and ruin the whole layout of the page.

Adjust the page so that it can handle the announcement box without disrupting the layout of the screen. The specific challenge in this is making the rest of the page responsive to its presence in a way that does not cause scrollbars to appear.

This reworks how the image appears on signup right now. Previously it was an `` tag. Now it is a background image again, this time with an `aria-description` that describes the text instead of an `img` tag with an `alt` attribute.

**Steps to test:**
1. Open the sign in/sign up pages with the alert banner present.
    - If you have access to this branch locally, you can create a test announcement in `src/app/announcement/data/events.ts` with a start time of `0` and end time of `Infinity` so that it's always present while testing.
    - If you are testing this in a deployed environment (dev/staging), then you can run the following script in the developer console once the page loads to simulate announcements:
<details>
<summary>[Code to paste into developer console]</summary>

```js
const n = document.createElement("div");
n.style = "height: 100px; background-color: white; display: flex; justify-content: center; align-items: center; font-weight: bold;"
n.innerText = "Pretend that the alert banner is here it's too hard to actually trigger it only in testing environments right now lol"
document.querySelector("pr-auth").prepend(n);
```
</details>

2. Verify that the page displays correctly, including at various screen widths/heights.

### DIFF
--- a/src/app/auth/components/auth/auth.component.html
+++ b/src/app/auth/components/auth/auth.component.html
@@ -1,33 +1,33 @@
-<!-- @format -->
-<div class="auth-container">
-  <div class="image">
-    <div class="testimony">
-      <p class="text">
-        At Permanent, we celebrate our members’ hard work through our Public
-        Gallery and archive spotlights. Explore our gallery and, when you’re
-        ready, publish or share your own public archive!
-      </p>
-      <br />
-      <a class="link" href="https://permanent.org/gallery"
-        >Start Exploring Now.</a
-      >
-      <p class="text archive-data">
-        Haadsma Dairy Truck | The Haadsma Dairy Archive
-      </p>
-    </div>
-    <img
-      src="assets/img/sign_in_image.png"
-      alt='Black and white photo of a child with a bicycle stopped in front of a black milk truck with a man standing by the door. The milk truck reads "Haadsma&#39;s Dairy." The photo appears to be from the 1950s.'
-    />
-  </div>
-  <div class="auth">
-    <div class="permanent-logo">
-      <img src="assets/svg/app-icon-glam.svg" alt="Permanent.org logo" />
-    </div>
-    <router-outlet></router-outlet>
-  </div>
-</div>
-<pr-mobile-banner></pr-mobile-banner>
 <pr-announcement></pr-announcement>
-<pr-android-app-notify></pr-android-app-notify>
-<pr-prompt></pr-prompt>
+<div class="auth-root">
+  <div class="auth-container">
+    <div
+      class="image"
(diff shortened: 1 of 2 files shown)


crisnicandrei (approved):

k8lyn6 (commented):
@meisekimiu I think this generally looks good, but I have a few questions to check just to make sure everything is working properly.

1. Am I supposed to still be seeing the sign in/sign up photo? It has disappeared, but the text box remains.
2. When I shrink the size of my browser down to mobile-adjacent size, the text is basically right on the edge of the browser on the left side. Is this just because the actual alert can't be triggered in the dev environment, or will this impact the text of the actual alert? Screenshot attached.
3. Similarly, there does seem to be a horizontal scroll bar when the screen is reduced to mobile-adjacent size, and there's extra white space on the right side of the screen. Not sure if this is just with the test banner or if it would also be the same with a real alert. I also think we could live with this issue and the previous issue if we needed to -- it won't be up for long this time around and we're on a time crunch. 

Screenshot 2025-01-24 at 9 59 16 AM

meisekimiu:
> 1. Am I supposed to still be seeing the sign in/sign up photo? It has disappeared, but the text box remains.

This is caused by an Angular discrepancy between local / deployed environments. I am working on a fix.

> 2\. When I shrink the size of my browser down to mobile-adjacent size, the text is basically right on the edge of the browser on the left side. Is this just because the actual alert can't be triggered in the dev environment, or will this impact the text of the actual alert? Screenshot attached.

This placeholder banner does not have the styling that the actual announcement has, it is just a hack to get the spacing of the page adjusted as if the announcement banner is here. The actual banner does have this styling, I just didn't think it was necessary to inject the entire announcement style into a placeholder. We do not have to adjust the styling of the announcement banner for this ticket.

> 3\. Similarly, there does seem to be a horizontal scroll bar when the screen is reduced to mobile-adjacent size, and there's extra white space on the right side of the screen. Not sure if this is just with the test banner or if it would also be the same with a real alert. I also think we could live with this issue and the previous issue if we needed to -- it won't be up for long this time around and we're on a time crunch.

I think the scrollbars are an artifact of the test banner. I can fix the white space.

k8lyn6:
@meisekimiu that makes sense! I wasn't sure which style issues would affect the banner, so I wanted to make sure. The white space isn't a deal breaker but if it's easy to fix quickly, let's do it.

k8lyn6 (approved):
I see the image now! I still see a horizontal scroll bar on a mobile-sized browser, but there's no white space. I think that'll be okay for now. Thanks Natalie!

PR #520, opened 2025-01-23
PER-9962 clicking search result in subfolder

crisnicandrei:
Navigate to the folder when clicking a record, just like the workspace search. I chose this approach because the dataservice only searches through the files in the current folder that the user is in, so searching for a file in a different folder would crash the app.


src/app/file-browser/components/file-viewer/file-viewer.component.ts
```diff
@@ -103,6 +103,7 @@ export class FileViewerComponent implements OnInit, OnDestroy {
       this.currentIndex = findIndex(this.records, {
         folder_linkId: resolvedRecord.folder_linkId,
       });
+
```
meisekimiu, line 106: This looks like a leftover that doesn't need to be committed.

meisekimiu (approved):

yeslikesolo (approved):

PR #521, opened 2025-01-23
Add FusionAuth downtime announcement banner

meisekimiu:
Add the banner as specified by the experience team.

I set it to be active at midnight on January 29th MST (picked arbitrarily since the maintenance window from FusionAuth is given in MST) and it will go away 6 hours after the maintenance window when we're sure that they'll be done.

There's no real way to test this without adjusting your computer's time, so mainly just verify that the input string is valid HTML and contains no errors or typos.


liam-lloyd (approved):

meisekimiu:
Merging this since QA wouldn't really accomplish anything.

PR #522, opened 2025-01-23
Add edge cases for Public Route generation

meisekimiu:
There are a few edge cases in the PublicRoutePipe that are not addressed and so they cause some incorrect URLs to be generated.

First, make it so that public objects in the public root do not need to feature their parent folder's archiveNbr and folder_linkId. This makes URLs match those that are found in the actual public gallery.

Second, address issues when it comes to mismatches in a record/folder's archiveNbr and the archiveNbr of the archive the item is contained in. In theory these numbers should always match, and the original (and new fallback) implementation utilizes this assumption. In practice, there are records in the public gallery in production that have archiveNbr mismatches. This is concerning and we should figure out what's causing it on the back-end, but we also need to adjust the link generation code to handle these existing records.

Use the `archiveArchiveNbr` on the record as the canonical source of the parent archive's ArchiveNbr. That is what this field is for so we should read it instead of trying to reverse-engineer it ourselves. However, in the case of the errant record on production, this alone does not fix URL generation since it's actually null. As a fallback, check the parent folder's `archiveArchiveNbr` as this value is properly set in the production bug we've found. Finally use the old algorithm if everything else fails.

**Steps to test:**
1. Verify that the "Get Link" link on public records that are not nested in folders works
2. Verify that the links work for other records and folders as well

I don't really know how to test the buggiest edge cases in dev/staging/production, so it might just be safe to verify with the automated tests.

### DIFF
--- a/src/app/models/folder-vo.ts
+++ b/src/app/models/folder-vo.ts
@@ -45,6 +45,7 @@ export class FolderVO
 
   public folderId;
   public archiveNbr;
+  public archiveArchiveNbr: string | undefined;
   public archiveId;
   public displayName;
   public displayDT;
--- a/src/app/models/record-vo.ts
+++ b/src/app/models/record-vo.ts
@@ -44,6 +44,7 @@ export class RecordVO
   public recordId;
   public archiveId;
   public archiveNbr;
+  public archiveArchiveNbr: string | undefined;
   public publicDT;
   public note;
   public displayName;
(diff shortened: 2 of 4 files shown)


liam-lloyd (approved):

k8lyn6 (approved):
I encounter no issues with opening individual records or folders using the Get Link and View on Web buttons, regardless of whether they lived in a folder or in the top level of the Public Files workspace.

The only thing I noticed is that the breadcrumbs for a file nested in a folder are missing, which I think will be covered in [PER-9994](https://permanent.atlassian.net/browse/PER-9994). Is that correct @meisekimiu ? If so, I approve this PR! If not, do I need to create a separate ticket for that issue? Should it be addressed here?

I'm going to go ahead and mark this approved, though, because I suspect if the issue isn't covered in the ticket I've already made, we should make a new ticket to address it.

meisekimiu:
The breadcrumbs are definitely a separate issue unrelated to this ticket. I'm not sure if they will be addressed in the linked ticket, but if they are not I will go ahead and make a separate ticket while doing review for that one.

PR #525, opened 2025-01-29
PER-9994- Back button missing

crisnicandrei:
- Removed @Input that was hiding the back button in breadcrumbs component

Steps to test:
1. Navigate to any public archive
2. Navigate to any public folder from that archive
3. The breadcrumbs should appear


meisekimiu (approved):

yeslikesolo (approved):
looks great!

PR #526, opened 2025-02-03
Add links to new user checklist

meisekimiu:
Add links to the currently defined tasks in the new user checklist. This also adds functionality to link to specific sections of the account settings, archive settings, and storage dialogs using the URL fragment. This deletes some of the previous code that handled this with query parameter subscriptions.

**Steps to test:**
1. Create a new account or access an account with the checklist active
2. Verify that the tasks with links defined do have links
3. Verify that the links work as expected
    - In-app links should open the correct dialog without any weird issues
    - External links should open in a new tab/window

**After this is merged and deployed:**
Because of these changes, the link to the "redeem gift code" dialog used in our emails should be changed to the following: `https://www.permanent.org/app/(private//dialog:storage/add)#promo`

### DIFF
--- a/src/app/core/components/account-settings-dialog/account-settings-dialog.component.ts
+++ b/src/app/core/components/account-settings-dialog/account-settings-dialog.component.ts
@@ -1,4 +1,3 @@
-/* @format */
 import { Component, Inject } from '@angular/core';
 import { AccountService } from '@shared/services/account/account.service';
 import { AccountResponse } from '@shared/services/api/index.repo';
@@ -7,16 +6,18 @@ import { ApiService } from '@shared/services/api/api.service';
 import { ActivatedRoute } from '@angular/router';
 import { DIALOG_DATA, DialogRef } from '@angular/cdk/dialog';
 
-export type SettingsTab =
-  | 'storage'
-  | 'account'
-  | 'notification'
-  | 'billing'
-  | 'legacy-contact'
-  | 'delete'
-  | 'advanced-settings'
-  | 'security';
+const settingsTabs = [
+  'storage',
+  'account',
+  'notification',
+  'billing',
+  'legacy-contact',
+  'delete',
+  'advanced-settings',
+  'security',
+] as const;
 
+export type SettingsTab = (typeof settingsTabs)[number];
 @Component({
   selector: 'pr-account-settings-dialog',
   templateUrl: './account-settings-dialog.component.html',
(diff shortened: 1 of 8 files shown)


liam-lloyd (approved):

yeslikesolo (changes requested):
Last two tasks, Upload first file and Publish archive are deploying the same KB link. I think we should have different links for this, @k8lyn6 ? 

Also, just wondering why we aren't using the same action of page redirection for the last two tasks? At least for Upload First File - can we use the same button link as the orange upload button? I think that makes more sense than the KB article because its a task we know (through data) that users are comfortable doing on their own. This would also skew some KB article KPIs and mixpanel tracking data for this particular task. Just a thought!

k8lyn6:
@meisekimiu that's my bad if I didn't give you the right links! The publish one should be [this one](https://permanent.zohodesk.com/portal/en/kb/articles/how-to-publish) and the upload one [this one](https://permanent.zohodesk.com/portal/en/kb/articles/upload-your-first-file).

@yeslikesolo I'm torn on having the upload task on the checklist go straight to the upload picker. If we open it for them, they can't choose where they want to upload their first file, since you can do that in multiple places. I'm not sure how much work that might entail to set up, too (@meisekimiu can chime in here if she wants) but especially if people tend to know how to do that already, I don't think we need to open the picker for them. I think we can just leave it as the KB article for now, and then change it later if we need to.

Thanks Natalie for doing this! We're very excited about it.

meisekimiu:
The links were accidentally duplicated and I've fixed that.

As for the last two tasks, there's not really a good way of making them act as in-app links. For upload we *could* have the link trigger the same logic that the orange upload button does, but this would require additional work compared to the other links. I'm also not sure it'd be the best user experience; users might be startled by the appearance of the file dialog and it might not be clear they'd be uploading into the current folder.

And as far as I can tell there's not any in-app link or function that could really work for publishing files.

k8lyn6:
@yeslikesolo oh yeah, I didn't answer for publish. Yeah, I think this is just a problem that needs to be solved with the work we're doing to consolidate share and publish. Also, a user might have many different files to choose from to publish, and it'd be hard to come up with a way to present their options after clicking on Publish from the checklist. 

But +1 to what Natalie said -- I think we should just keep the KB articles for now and potentially revisit in the future.

Han, can you re-test those KB articles and then I think this will be good to go!

yeslikesolo (approved):
Cool, thanks Natalie! Everything is good to go

PR #527, opened 2025-02-04
Per 9985 entering 2fa incorrect code works

crisnicandrei:
This pull request fixes the bug that occurs when entering an incorrect code when removing a multi factor method

Steps to test:
1. Choose any method you want to remove
2. Send a code
3. Enter any code but NOT the one that you received
4. The method should not be removed
@meisekimiu Do you think this implementation will do? Or should we change the backend as I mentioned on Slack?

### DIFF
--- a/src/app/shared/services/http-v2/http-v2.service.ts
+++ b/src/app/shared/services/http-v2/http-v2.service.ts
@@ -14,8 +14,6 @@ const AUTH_KEY = 'AUTH_TOKEN';
 type HttpMethod = 'post' | 'get' | 'put' | 'delete';
 type ResponseClass<T> = new (data: any) => T;
 
-type ResType = 'json' | 'text';
-
 interface RequestOptions {
   csrf?: boolean;
   authToken?: boolean;
--- a/src/app/shared/services/http-v2/http-v2.service.ts
+++ b/src/app/shared/services/http-v2/http-v2.service.ts
@@ -14,6 +14,8 @@ const AUTH_KEY = 'AUTH_TOKEN';
 type HttpMethod = 'post' | 'get' | 'put' | 'delete';
 type ResponseClass<T> = new (data: any) => T;
 
+type ResType = 'json' | 'text';
+
 interface RequestOptions {
   csrf?: boolean;
   authToken?: boolean;
@@ -61,6 +63,7 @@ export class HttpV2Service {
     data: any = {},
     responseClass?: ResponseClass<T>,
     options: RequestOptions = defaultOptions,
+    responseType?: 'json' | 'text',
   ): Observable<T[]> {
     return this.makeHttpClientRequest(
       endpoint,
@@ -68,6 +71,7 @@ export class HttpV2Service {
       'post',
       responseClass,
       this.getOptions(options),
+      responseType,
     );
   }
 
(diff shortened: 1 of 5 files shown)


src/app/shared/services/http-v2/http-v2.service.ts
```diff
@@ -61,13 +61,15 @@ export class HttpV2Service {
     data: any = {},
     responseClass?: ResponseClass<T>,
     options: RequestOptions = defaultOptions,
+    responseType?: 'json' | 'text',
```
meisekimiu, line 64: Thank you for adding this new feature to the HttpV2Service. I think ideally I'd like to see `responseType` in the `options` object instead of a positional argument. It makes our public methods have up to five arguments and some internal methods have even more. This can be annoying when we want to make a basic request with default options but we need to adjust the response type because we have to fill out every argument that would have been optional. (I regret having options come *after* `resposneClass` since those options are typically set more than the `responseClass` is).

src/app/shared/services/http-v2/http-v2.service.ts
```diff
@@ -76,13 +78,15 @@ export class HttpV2Service {
...
```
meisekimiu, line 81: Since we're using this `'json' | 'text'` typing everywhere, we should probably just define it as a specific type like
```ts
type ResponseType = 'json' | 'text';
```

That way if we ever have to communicate with an external API that uses XML for instance we only have to change this typing on one line instead of many.

src/app/shared/services/http-v2/http-v2.service.ts
```diff
@@ -231,10 +253,22 @@ export class HttpV2Service {
...
+      data,
+      method,
+      options,
+      responseType,
+    );
+
     return observable.pipe(
       map((response: Object | Array<Object>) => {
+        if (responseType === 'text') {
+          return [response as unknown as T];
```
meisekimiu, line 269: This is a legitimate question we should probably think about... if the resposneType is text, we should always be returning a string instead of an object so response should just be able to be cast to string here. Of course, this function is expecting an observable of type `T[]`. I wonder if there's anyway to force `T` to be `string` when `responseType` is set to `text`.

src/app/shared/services/http-v2/http-v2.service.ts
```diff
@@ -231,10 +253,22 @@ export class HttpV2Service {
     method: HttpMethod = 'post',
     responseClass?: new (data: any) => T,
     options: RequestOptions = defaultOptions,
+    responseType: 'json' | 'text' = 'json', // NEW: response type parameter
```
meisekimiu, line 256: We do not need that comment.

src/app/shared/services/http-v2/http-v2.service.ts
crisnicandrei: Good question. Maybe this could be a separate ticket? I have tried doing just that but I got many errors in different files unrelated to the ticket

crisnicandrei:
@meisekimiu addressed your comments! Thank you!

meisekimiu (approved):

k8lyn6 (commented):
This fix does resolve the issue (submitting an incorrect code just takes you back to the beginning). However, I don't think this is the best user experience, as it might still be a little confusing. It would be a better user experience if entering an incorrect code gives you an error and allows you to try entering the code again. 

@meisekimiu would this be doable with this ticket, or does it require backend work? If it requires work outside of the ticket, I'll go ahead and approve this and we can create additional tickets.

k8lyn6 (approved):
Thanks @meisekimiu . I'll create an additional ticket and go ahead and mark this as approved. Thanks @crisnicandrei !

PR #528, opened 2025-02-06
Per 10036 swap out sign in image

crisnicandrei:


src/app/auth/components/auth/auth.component.scss
```diff
@@ -58,7 +58,7 @@
 .image {
   position: relative;
   height: 100%;
-  background: url('../../../../assets/img/sign_in_image.png');
```
meisekimiu, line 61: Can you give this a bit of a clearer name like `legacy_lab_sign_in_image.png`? Just so it's easier to track and understand its purpose without opening the file.

src/app/auth/components/auth/auth.component.html
```diff
@@ -18,6 +18,25 @@
...
+      <div class="legacy-lab-container">
+        <div class="legacy-lab-text">
+          <h1 class="header">Legacy Lab</h1>
+          <h2 class="start-date">Begins March 3rd</h2>
+          <p class="text">
+            At Permanent, we believe in the power of preserving your story. Join
+            Legacy Lab, a free four-week course where you'll create a lasting
+            archive and learn through:
+          </p>
+          <ul class="list">
```
meisekimiu, line 31: In theory we don't need some of these classes, right? There's only one kind of list that will ever show up in `.legacy-lab-container`. I think some of these classes are fine (such as `start-date` which conveys more semantic meaning about what the text is doing) but the `ul` tag is *always* going to represent a list so I feel like representing that in the classname is a bit redundant. Same for `list-item` on the `li` tags.

meisekimiu (approved):
Left some code style comments, though I'll preemptively approve this so that we can get the experience team looking at this.

meisekimiu:
@yeslikesolo and I met to go over some adjustments to this page that we wanted to make and I applied them to this branch! Note that we can now remove the `legacy-lab` class from some of the HTML to help revert styles easier and swap back to this if needed without deleting the CSS.

yeslikesolo (approved):
WOOT WOOT! Looks great

meisekimiu:
I'm going to rebase and merge this so that it can go out in a deployment this week, regardless of when the deployment happens.

PR #530, opened 2025-02-11
PER-9982 hitting enter submit 2fa code

crisnicandrei:
Add new method to submit 2fa code when user hits 

Steps to test:
Generate a 2fa code via email or sms
Add it in the input field and then press eneter
It should add/remove the method

### DIFF
--- a/src/app/core/components/two-factor-auth/two-factor-auth.component.html
+++ b/src/app/core/components/two-factor-auth/two-factor-auth.component.html
@@ -50,7 +50,12 @@
     <option [disabled]="hasMethod('email')" value="email">Email</option>
     <option [disabled]="hasMethod('sms')" value="sms">SMS</option>
   </select>
-  <form class="send-code-form" *ngIf="method" [formGroup]="form" (submit)="sendCode(form.value)">
+  <form
+    class="send-code-form"
+    *ngIf="method"
+    [formGroup]="form"
+    (submit)="sendCode(form.value)"
+  >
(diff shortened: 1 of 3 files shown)


src/app/core/components/two-factor-auth/two-factor-auth.component.html
```diff
@@ -72,6 +76,7 @@
         }}"
         [control]="form.controls['contactInfo']"
         [disabled]="!!selectedMethodToDelete"
+        (keydown.enter)="onEnterPress($event, 'sendCode')"
```
meisekimiu, line 79: In general we should avoid mapping a literal `keydown.enter` event to these inputs. While the ticket specifically calls out hitting the enter key, we should be using more generalized solutions. Specifically: using a form `submit` event instead. Mobile keyboards often have some kind of submit functionality built into them and they usually work by submitting the form itself rather than actually triggering an enter key press. So this solution might not work as smoothly on mobile devices.

Right now we have two different inputs with different purposes under one form, which is causing the existing bug. But I think semantically it makes sense for these inputs to be in *two separate forms* since they have completely different purposes and actions upon submitting. This lets us hit enter/submit/etc on each input and have them do separate things.

meisekimiu (changes requested):

meisekimiu (approved):
This is fine! There are already existing tests that run `sendCode` and `submitData` so we don't really need to test them through the UI.

yeslikesolo (approved):
Everything looks great! :D

PR #531, opened 2025-02-13
PER-10055: Reorder tabs in archive settings dialog

crisnicandrei:
- Reorder tabs in archive settings dialog
- Make sure that the first tab when opened is archive settings

### DIFF
--- a/src/app/core/components/archive-settings-dialog/archive-settings-dialog.component.html
+++ b/src/app/core/components/archive-settings-dialog/archive-settings-dialog.component.html
@@ -10,20 +10,6 @@
     <div class="tabs">
       <div
         class="dialog-tab"
-        (click)="setTab('manage-keywords')"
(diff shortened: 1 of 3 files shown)


meisekimiu (approved):

crisnicandrei:
@yeslikesolo it works for me 
Screenshot from 2025-02-19 10-47-34

k8lyn6:
@yeslikesolo I just deployed and took a look, and the order has updated for me. What browser are you using? Have you tried logging out and logging back in/doing a hard refresh of your browser? 

@crisnicandrei I think this is probably good to go, but I want to confirm with Han that she sees the updated order before we merge in case there's a weird edge case bug we're missing.

yeslikesolo:
It worked for me now! Had to open in a new browser.

k8lyn6 (approved):
I think this is good to go then! Thanks for double checking @yeslikesolo .

PR #533, opened 2025-02-13
Fix Legacy Lab Promo Link

meisekimiu:
The link wasn't clickable due to the fact that it was underneath another layer. Move up its z-index so that it's clickable.

**Steps to test:**
1. Check if the link on the sign up screen works


yeslikesolo (approved):
Thank you, natalie!

PR #534, opened 2025-02-17
PER-10030: Add link to redeem storage

crisnicandrei:


src/app/core/components/storage-dialog/storage-dialog.component.html
```diff
@@ -61,6 +61,9 @@
             >
           </p>
           <pr-new-pledge [inlineFlow]="true"></pr-new-pledge>
+          <button class="btn btn-link redeem-promo" (click)="setTab('promo')">
+            Have a code to redeem for free storage? Click here to redeem it.
+          </button>
```
meisekimiu, line 66: We might want this to be an `a` tag instead since from the user perspective, this is essentially a link. For accessibility purposes we probably want this to be a link to match that expectation rather than it being a button.

meisekimiu (approved):

k8lyn6 (approved):
This looks good to me! Thanks Andrei!

PR #535, opened 2025-02-19
PER-10060-text-does-not-wrap-desc-field

crisnicandrei:
Add text wrap to the description field in the inline value edit component

### DIFF
--- a/src/app/shared/components/inline-value-edit/inline-value-edit.component.scss
+++ b/src/app/shared/components/inline-value-edit/inline-value-edit.component.scss
@@ -120,6 +120,7 @@ select {
 .inline-value-textarea {
   min-height: 6 * 1.5rem;
   overflow-wrap: break-word;
+  white-space: pre-wrap;
 
   &.showing-empty-message {
     color: $text-muted;


src/app/shared/components/inline-value-edit/inline-value-edit.component.scss
```diff
@@ -120,6 +120,7 @@ select {
 .inline-value-textarea {
   min-height: 6 * 1.5rem;
   white-space: pre-wrap;
```
meisekimiu, line 122:
```suggestion
```
I don't think we need the `white-space` attribute anymore since `overflow-wrap` (presumably?) overrides its behavior, and including both of them is a bit confusing.

meisekimiu (approved):

k8lyn6:
@crisnicandrei I'm so sorry! Somehow this PR slipped past me. I'll look at it today!

k8lyn6 (changes requested):
Hey @crisnicandrei , I think this solves the issue from the ticket! However, I did notice that it is no longer preserving line breaks created using the Enter button. Would you be able to fix that so line breaks are preserved while also keeping the fix for the wrapping issue?

PR #538, opened 2025-02-24
Per use v2 for redeem code

crisnicandrei:


src/app/core/components/redeem-gift/redeem-gift.component.ts
```diff
@@ -64,7 +65,7 @@ export class RedeemGiftComponent implements OnInit {
     this.resultMessage = { message, successful: false };
   }
 
-  private async handleValidPromoCode(response: BillingResponse) {
+  private async handleValidPromoCode(response) {
```
meisekimiu, line 68: This `response` parameter should be explicitly typed. As far as I can tell, this should be of type `ClaimingPromoResponse`, right?

src/app/core/components/redeem-gift/shared-mocks.ts
```diff
@@ -1,23 +1,14 @@
...
-  isSuccessful: true,
-};
+const mockPromoResponse = [
+  {
+    promoId: 13,
+    code: 'promo9',
+    sizeInMB: 5000,
+  },
+];
 const failedPromoResponse = {
```
meisekimiu, line 12: This failed response should probably match what we'd expect from an incorrect response from the actual API endpoint. It doesn't really matter, since no logic is reading from fields of a failed response, but it's still weird to see the signature format of the old V1 API response being used in a failed response.

crisnicandrei: @meisekimiu it seems the response is the same for errors as in the mocks.

meisekimiu (approved):

yeslikesolo (approved):

PR #540, opened 2025-03-03
Per 10047 use v2 record update

crisnicandrei:
Use v2 api calls for the record/update endpoint

### DIFF
--- a/src/app/core/services/edit/edit.service.spec.ts
+++ b/src/app/core/services/edit/edit.service.spec.ts
@@ -161,7 +161,7 @@ describe('EditService', () => {
       mockRecords,
       archiveId,
     );
-    
+
     expect(results[0][0].updatedDT).toBe('2024-03-03');
   });
 
--- a/src/app/core/services/edit/edit.service.spec.ts
+++ b/src/app/core/services/edit/edit.service.spec.ts
@@ -55,23 +55,14 @@ describe('EditService', () => {
   }));
 
   it('should call update when there are records', async () => {
-    const mockRecords = [new RecordVO({ recordId: 1, displayName: 'Test' })];
+    const record = new RecordVO({ recordId: 1, displayName: 'Test' });
+    const mockRecords = [record];
     (apiService.record.update as jasmine.Spy).and.returnValue(
       Promise.resolve([{ updatedDT: '2024-03-03' }]),
     );
-
-    const promises = [];
-    if (mockRecords.length) {
-      const archiveId = accountService.getArchive().archiveId;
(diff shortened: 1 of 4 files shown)


meisekimiu (changes requested):
I left some feedback on how to fix the coverage issues, as well as some miscellaneous comments.

src/app/core/services/edit/edit.service.spec.ts
```diff
@@ -1,21 +1,188 @@
...
+
+      const record =
+        (itemsByLinkId[res.folder_linkId] as RecordVO) ||
+        recordsByRecordId.get(res.recordId);
+
+      record.update(newData);
+    }
+
+    expect(recordMock.update).not.toHaveBeenCalled();
+  });
```
meisekimiu, line 187: Can you either edit these to call the `EditService` like in the test I wrote or just delete them since these aren't testing any production code?

meisekimiu (approved):
Thanks for addressing the test coverage issues! Do you think you can address the minor type issues I pointed out as well?

crisnicandrei:
@meisekimiu the backend is returning an array containing an object, not a RecordVO. Should I change the api call request on the front-end or should the back-end be changed?

meisekimiu:
It's an array but shouldn't it be an array of RecordVOs? As far as I know all of the v2 endpoints have simplified bodies and are not of the `-Response` type. Is `record/update` different?

crisnicandrei:
@meisekimiu yeah it still returns an array for some reason. Or what do you mean?

k8lyn6:
@yeslikesolo this will be for testing metadata updates. Please test all different kinds of metadata updates across different workspaces and double check that no actual record metadata info is present in the mixpanel data. I'm not entirely sure what this will look like in mixpanel so feel free to ask questions to me (or save them for Liam on Monday) if you see anything that's confusing!

meisekimiu:
> @meisekimiu yeah it still returns an array for some reason. Or what do you mean?

I mean that it returns an array of RecordVo objects instead of the typical `RecordResponse` type. The latter being a data structure with the typical V1 response data being attached as well as the array of record objects deep into the structure. I think it'd be for the best to change that typing since we're not expecting a singular (or array of!) `RecordResponse` objects.

yeslikesolo (approved):

k8lyn6:
@meisekimiu I wanted to check and see if this was ready to merge since Andrei is out. Is there additional work that's needed, do you know?

meisekimiu:
This is basically ready to go. There's a small change that I can make really quickly that won't effect runtime behavior so we won't need to redo QA.

PR #541, opened 2025-03-04
swap legacy lab image

crisnicandrei:
This pr swaps the legacy lab sign in image to the new one


src/app/auth/components/auth/auth.component.scss
```diff
@@ -58,29 +58,9 @@
...
-    position: absolute;
-    left: 0;
-    top: 0;
-    width: 100%;
-    height: 100%;
-    background: url('../../../../assets/img/legacy_lab_sign_in_image.png');
-    background-size: contain;
-    background-position: center right 1.5rem;
-    background-repeat: no-repeat;
-  }
```
meisekimiu, line 79: It might be useful to ask if we're going to reuse the legacy lab screen. If we are, then we should just leave this code here so we can swap the class back. @k8lyn6 do you know if we're going to reuse this same screen design in the future?

meisekimiu (approved):

src/app/auth/components/auth/auth.component.scss
k8lyn6: @meisekimiu @crisnicandrei Yes, I suspect we'll want to put it back up (with an amended date) at the end of August for the next round of Legacy Lab. Probably no need to change the design since we already have it working!

k8lyn6:
Also, @crisnicandrei is this ready for QA? I want to try to get this change out ASAP this week, so I apologize for moving quickly!

meisekimiu:
This is ready for QA, though if we know we'll reuse the design of the legacy lab screen I want to make sure that code isn't deleted before we merge.

yeslikesolo (approved):
This is minor, but I will comment just in case for improvements. The current screen is a good MVP.

I think the default image is a bit too large. It's overflowing past the fold on my screen. Can we reduce the image size by 1 rem for width and length? 
Screenshot 2025-03-11 at 11 27 54 AM

k8lyn6:
Thanks for the photo @yeslikesolo ! Another minor styling change, but I think the deadline ("Deadline to apply: April 1, 2025") was supposed to be bolded according to the ticket. @meisekimiu would you or @crisnicandrei be able to make those changes today to get this out today? If not, that's okay!

PR #544, opened 2025-03-13
Do not send Mixpanel events if auth token is unset

meisekimiu:
The `event` endpoint expects a valid auth token. Trying to send an event when not logged in causes an error to happen which refreshes the page. On pages like the public gallery, this creates an endless loop of refreshes.

**Steps to test:**
1. View the public gallery page when logged out and verify that the page doesn't auto-refresh endlessly.
2. Run the following line in the developer console and verify that the page refreshes a maximum of one time: `localStorage.setItem('AUTH_TOKEN', 'test');`


crisnicandrei (approved):

k8lyn6 (approved):
Looks like this addresses the issue! I used the command and it didn't do anything, but then I refreshed the page and it refreshed one more time before resolving, so I think that works. Thanks for the quick fix!

PR #546, opened 2025-03-19
PER-v2-invite/share

crisnicandrei:


meisekimiu (approved):

k8lyn6 (commented):
Leaving a comment that I see this is being called (which is good!) but I'm not certain in what circumstances it's supposed to be called. That makes it difficult to test. I'll touch base with @liam-lloyd about it when he returns on Monday and finish testing then.

PR #549, opened 2025-03-31
Update properties in v2 record/update call

meisekimiu:
`displayDT`, `displayEndDT`, and `publicDT` all have case changes in the V2 version of `record/update`. Update the request body to match the new expected values.

**Steps to test:**
1. Verify that changing a record's date works properly.
2. Verify that changing other properties does not erase a record's date.
3. Verify that dates are still fetched properly.

### DIFF
--- a/src/app/shared/services/api/record.repo.ts
+++ b/src/app/shared/services/api/record.repo.ts
@@ -159,8 +159,26 @@ export class RecordRepo extends BaseRepo {
 
   public async update(recordVOs: RecordVO[], archiveId: number) {
     return await firstValueFrom(
-      this.httpV2.post('/record/update', { ...recordVOs[0], archiveId }, null),
+      this.httpV2.post(
+        '/record/update',
+        this.prepareUpdateRecordRequest(recordVOs, archiveId),
+        null,
+      ),
     );
+  }
+
+  private prepareUpdateRecordRequest(recordVOs: RecordVO[], archiveId: number) {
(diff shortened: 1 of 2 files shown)


k8lyn6 (approved):
This resolves the issue with dates. I verified that I can reliably change a date and it persists after refreshing. Editing other metadata fields does not affect the set date.

src/app/shared/services/api/record.repo.ts
```diff
@@ -159,10 +159,28 @@ export class RecordRepo extends BaseRepo {
...
-      this.httpV2.post('/record/update', { ...recordVOs[0], archiveId }, null),
+      this.httpV2.post(
+        '/record/update',
+        this.prepareUpdateRecordRequest(recordVOs, archiveId),
+        null,
+      ),
     );
   }
 
+  private prepareUpdateRecordRequest(recordVOs: RecordVO[], archiveId: number) {
```
liam-lloyd, line 170: What type does this end up returning? (approving regardless)

liam-lloyd (approved):

src/app/shared/services/api/record.repo.ts
meisekimiu: This is returning a thing that is like a `RecordVO` but with the cases swapped for what the endpoint is expecting. This probably could be a type called `UpdateRecordV2Request` or something like that, but since it's only used internally here I have not externally defined it.

PR #551, opened 2025-04-02
PER-9419-call-v2-get

crisnicandrei:
Use the v2 version of the record get api call from stela.

### DIFF
--- a/src/app/shared/services/api/test-utils.ts
+++ b/src/app/shared/services/api/test-utils.ts
@@ -1,0 +1,30 @@
+export function buildUrlWithParams(
+  baseUrl: string,
+  queryParams: Record<string, any>,
+): string {
+  const params = new URLSearchParams();
+
+  const buildParam = (key: string, value: any) => {
+    if (Array.isArray(value)) {
+      value.forEach((v, index) => {
+        if (typeof v === 'object' && v !== null) {
+          Object.entries(v).forEach(([k, val]) =>
+            params.append(`${key}[${index}][${k}]`, String(val)),
+          );
+        } else {
+          params.append(`${key}[]`, String(v));
+        }
+      });
+    } else if (typeof value === 'object' && value !== null) {
+      Object.entries(value).forEach(([k, val]) =>
+        params.append(`${key}[${k}]`, String(val)),
+      );
+    } else {
+      params.append(key, String(value));
+    }
+  };
+
+  Object.entries(queryParams).forEach(([key, value]) => buildParam(key, value));
+
+  return `${baseUrl}?${params.toString()}`;
+}
--- a/src/app/shared/services/api/test-utils.ts
+++ b/src/app/shared/services/api/test-utils.ts
@@ -1,30 +1,39 @@
+function buildParam(
+  params: URLSearchParams,
+  [paramKey, paramValue]: [string, any],
+): URLSearchParams {
+  if (Array.isArray(paramValue)) {
+    paramValue.forEach((arrayItem, index) => {
+      if (typeof arrayItem === 'object' && arrayItem !== null) {
+        Object.entries(arrayItem).forEach(([objectKey, objectValue]) => {
+          params.append(
+            `${paramKey}[${index}][${objectKey}]`,
+            String(objectValue),
+          );
+        });
+      } else {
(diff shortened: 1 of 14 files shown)


src/app/shared/services/http-v2/http-v2-encoder.ts
```diff
@@ -2,7 +2,7 @@ import { HttpParameterCodec } from '@angular/common/http';
 
 export class HttpV2Encoder implements HttpParameterCodec {
   public encodeKey(key: string): string {
-    return encodeURIComponent(key).replace('%5B', '[').replace('%5D', ']');
+    return encodeURIComponent(key).replace('%255B', '[').replace('%255D', ']');
```
meisekimiu, line 5: Does the previous behavior not work in browsers? I added this functionality in the first place so that we can use the `[]` characters in tests just like the calls in Stela test cases.

If we do need this change to call things properly in browser, then I highly recommend deleting the HttpV2Encoder class outright since this change is not doing anything. I also suggest adding some kind of testing utility function so we do not have to write something like `arrayVals%5B%5D=1&arrayVals%5B%5D=2&arrayVals%5B%5D=3` in all of our tests. Tests should be very easy to understand and should read like documentation and I think it's incredibly unpleasant to have to manage URL values like this written in tests.

meisekimiu (changes requested):

src/app/shared/services/http-v2/http-v2-encoder.ts
crisnicandrei: No, sadly it did not work. The character for paranthesis was 255B in the browser, thats why I changed it

src/app/shared/services/api/test-utils.ts
```diff
@@ -0,0 +1,30 @@
...
+    } else if (typeof value === 'object' && value !== null) {
+      Object.entries(value).forEach(([k, val]) =>
+        params.append(`${key}[${k}]`, String(val)),
+      );
+    } else {
+      params.append(key, String(value));
+    }
+  };
+
+  Object.entries(queryParams).forEach(([key, value]) => buildParam(key, value));
```
liam-lloyd, line 27: I think this function would be more readable if you modified `buildParam` to have the signature `(params: URLSearchParams, [key: string, value: any]): URLSearchParams`, and then used a `reduce` operation here. It's easier to grasp what a function does when it doesn't have side effects, and this would also allow you to define `buildParam` outside of `buildUrlWithParams`.

liam-lloyd (changes requested):

src/app/shared/services/api/test-utils.ts
```diff
@@ -1,30 +1,39 @@
...
-      );
-    } else {
-      params.append(key, String(value));
-    }
-  };
-
-  Object.entries(queryParams).forEach(([key, value]) => buildParam(key, value));
+  const params = Object.entries(queryParams).reduce(
+    (accumulatedParams, entry) => buildParam(accumulatedParams, entry),
+    new URLSearchParams(),
```
liam-lloyd, line 35: I think you can make this slightly more concise by passing just `buildParam` in place of `(accumulatedParams, entry) => buildParam(accumulatedParams, entry)`

liam-lloyd (approved):

k8lyn6 (changes requested):
@crisnicandrei thanks for working on this! I am trying to download any file from my archive, and the drop down for selecting which file to download is missing from all files (not just ones in an error state). It seems to be working on prod. I've attached a video showing what I mean.

https://github.com/user-attachments/assets/5f408072-41e5-4bed-8439-06c6be7f8e9c

k8lyn6 (changes requested):
Hey @crisnicandrei I'm able to see the drop down now (yay!) but clicking on any of the available download options does not start a download. This is true for all individual files regardless of whether they processed successfully or not. Downloading a folder still works, though. I've attached a video. I tried on both Chrome and Firefox with no success.

https://github.com/user-attachments/assets/59022461-bf82-4088-9fa8-8ed28816352a

crisnicandrei:
@k8lyn6 fixed the download. Sorry!

k8lyn6 (approved):
This looks good now, @crisnicandrei , thank you!

PR #553, opened 2025-04-04
Unlisted Shares API methods

meisekimiu:
This is where I left off with the unlisted shares API work. Please note that the module should be renamed from `shares-unlisted` to `share-links` and the `unlisted-shares-api` should be `share-links-api`. This was the temporary name I was working on but I realize that this functionality encompasses share links rather than just unlisted shares.

Also if you use this code please include me in a Co-authored-by tag or my spirit will haunt this codebase forever


slifty (approved):
This looks good!

I will say that in future we should seriously consider, as a team, the strategic benefit of adding this type of code to our SDK (and using the SDK) rather than implementing it in web-app.

Writing it in the SDK means that the additions are re-usable across projects / we know things are working for any SDK users as well.  It also helps separate responsibilities of our code bases.  Finally, if we do end up making improvements to the SDK (e.g. using codegen to make type definitions / etc), then we're able to benefit down the line!

cc @cecilia-donnelly and @liam-lloyd just to put this thought on their radars as well!

slifty:
@crisnicandrei since this is just adding the support and not using it I agree.  When we actually *use* it somewhere then QA will be able to verify if everything is as expected.

PR #554, opened 2025-04-08
PER-9933- use v2 folder/getwithchildren

crisnicandrei:
This pull request replaces the current api call for folder/getWithChildren with it's v2 version

### DIFF
--- a/src/app/shared/services/api/folder.repo.spec.ts
+++ b/src/app/shared/services/api/folder.repo.spec.ts
@@ -13,7 +13,11 @@ import {
   AccountPasswordVO,
   AccountVO,
   ArchiveVO,
+  FolderVO,
 } from '@root/app/models';
+import { HttpV2Service } from '../http-v2/http-v2.service';
+
+const apiUrl = (endpoint: string) => `${environment.apiUrl}/${endpoint}`;
 
 describe('FolderRepo', () => {
   let repo: FolderRepo;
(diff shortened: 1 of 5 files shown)


src/app/shared/services/api/folder.repo.spec.ts
```diff
@@ -13,7 +13,11 @@ import {
   AccountPasswordVO,
   AccountVO,
   ArchiveVO,
+  FolderVO,
 } from '@root/app/models';
+import { HttpV2Service } from '../http-v2/http-v2.service';
+
+const apiUrl = (endpoint: string) => `${environment.apiUrl}/${endpoint}`;
```
liam-lloyd, line 20: Where is this new function used?

liam-lloyd (approved):

k8lyn6:
@crisnicandrei could you provide some testing steps here? I'm not entirely sure what I should be testing. What parts of the application does this pull request potentially change? Should there be any different behaviors than our current webapp?

crisnicandrei:
@k8lyn6 Of course, sorry I missed that! Open the private workspace and the network tab. Refresh the page and check to see if the getWithChildren api call has a 200 response status and that everything works as it should :)

k8lyn6 (commented):
@crisnicandrei 

https://github.com/user-attachments/assets/5b5ca6d6-dda9-4464-a60d-b36c21469062

Thank you for the testing instructions! I was able to successfully see the getWithChildren with a 200 status. However, the page took a very long time to load (over 30 seconds, recording attached). Do you think that's related to this change? For reference, it takes maybe 1-3 seconds to refresh the Private Workspace on prod right now.

crisnicandrei:
@k8lyn6 Thanks for the video. I just tested on my machine and it worked fine. @liam-lloyd could there be other issues besides this?

k8lyn6 (approved):
@crisnicandrei okay, I tested again and I guess I didn't refresh the page without having the dev console open last time? When I close the dev console in my browser, the refresh load time is just a second or two. I didn't realize that might cause additional issues when I tested last time! I think this PR is probably fine to merge. Thanks!

PR #555, opened 2025-04-15
Remove "Deadline" text from the sign in/sign up pages

crisnicandrei:


slifty (approved):
Thanks for the quick fix!

crisnicandrei:
@yeslikesolo Are you sure? Maybe it needs to be deployed. This has passed code review by @slifty and the file changes reflect the change so I am unsure

yeslikesolo (approved):
Okay I gave it another try, we are good!

PR #560, opened 2025-04-29
Use query params to check which device needs verif

crisnicandrei:
Steps to test:

Add phone number
Click verify phone
it should redirect you as expected to phone verification

Navigate back
Go to email verification
It should take you to email verification

### DIFF
--- a/src/app/auth/components/verify/verify.component.ts
+++ b/src/app/auth/components/verify/verify.component.ts
@@ -62,8 +62,10 @@ export class VerifyComponent implements OnInit {
     }
     const queryParams = route.snapshot.queryParams;
 
-    this.needsEmail = queryParams.sendEmail;
-    this.needsPhone = queryParams.sendSms;
+    this.needsEmail =
+      (account.emailNeedsVerification() || queryParams.sendEmail) &&
+      !queryParams.sendSms;
+    this.needsPhone = account.phoneNeedsVerification() || queryParams.sendSms;
 
     this.verifyForm = fb.group({
       token: [queryParams.token || ''],
--- a/src/app/auth/components/verify/verify.component.ts
+++ b/src/app/auth/components/verify/verify.component.ts
@@ -14,6 +14,7 @@ import { AccountVO } from '@root/app/models';
 import { SecretsService } from '@shared/services/secrets/secrets.service';
 import { RecaptchaErrorParameters } from 'ng-recaptcha';
 import { EventService } from '@shared/services/event/event.service';
+import { C } from '@angular/cdk/keycodes';
 
 @Component({
   selector: 'pr-verify',
@@ -59,11 +60,10 @@ export class VerifyComponent implements OnInit {
       });
       return;
     }
-
-    this.needsEmail = account.emailNeedsVerification();
-    this.needsPhone = account.phoneNeedsVerification();
-
     const queryParams = route.snapshot.queryParams;
+
+    this.needsEmail = queryParams.sendEmail;
+    this.needsPhone = queryParams.sendSms;
 
     this.verifyForm = fb.group({
       token: [queryParams.token || ''],
--- a/src/app/auth/components/verify/verify.component.ts
+++ b/src/app/auth/components/verify/verify.component.ts
@@ -30,6 +30,8 @@ export class VerifyComponent implements OnInit {
   verifyingEmail = true;
   verifyingPhone = false;
 
+  currentVerifyFlow: 'none' | 'email' | 'phone' = 'none';
+
   needsEmail: boolean;
   needsPhone: boolean;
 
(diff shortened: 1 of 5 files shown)


src/app/auth/components/verify/verify.component.ts
```diff
@@ -59,12 +60,11 @@ export class VerifyComponent implements OnInit {
       });
       return;
     }
-
-    this.needsEmail = account.emailNeedsVerification();
-    this.needsPhone = account.phoneNeedsVerification();
-
     const queryParams = route.snapshot.queryParams;
 
+    this.needsEmail = queryParams.sendEmail;
```
slifty, line 65: What happens if these are not specified in the querystring at all / where is `sendEmail` and `sendSms` actually set?

I understand that we want to be able to explicitly set / override which verification flow is happening, but should we still have a default value based on the account's verification status in the event that no specific flow / override is currently underway?

crisnicandrei: Good question, I think we can discuss this during today's call because there are some edge cases that made me choose this approach.

src/app/auth/components/verify/verify.component.ts
```diff
@@ -59,12 +60,13 @@ export class VerifyComponent implements OnInit {
       });
       return;
     }
-
-    this.needsEmail = account.emailNeedsVerification();
-    this.needsPhone = account.phoneNeedsVerification();
-
     const queryParams = route.snapshot.queryParams;
 
+    this.needsEmail =
```
slifty, line 65: This variable name no longer actually describes what it claims to contain -- specifically, `needsEmail` might contain false even though the current user does, in fact, need email.  We should be careful here, as this represents tech debt / makes the variable un-intuitive.

Instead of changing the purpose of those variables I think it would be cleaner and more direct to populate and use a *new* variable which captures our true intent.  For instance:

```
this.currentVerifyFlow = (
  this.needsEmail && !this.queryParams?.sendSms ? 'email' : 'phone'
)
```

Later in the component we would then branch behavior based on the value of `this.currentVerifyFlow`.

NOTE: I wrote `this.queryParams?...` because later we have an `if (queryParams)` which indicates it's possible for queryParams to be a nullish value, and accessing it directly would risk an error.

slifty (changes requested):
Slight suggested refactor to make sure the code is easy to understand later; also it looks like tests currently fail.

src/app/auth/components/verify/verify.component.ts
```diff
@@ -99,12 +110,15 @@ export class VerifyComponent implements OnInit {
...
-      this.formTitle = 'Verify Phone Number';
-    } else if (!this.needsEmail) {
-      this.router.navigate(['/private'], { queryParamsHandling: 'preserve' });
+    switch (this.currentVerifyFlow) {
+      case 'phone':
+        this.verifyingEmail = false;
+        this.verifyingPhone = true;
+        this.formTitle = 'Verify Phone Number';
+        break;
+      case 'none':
```
slifty, line 119: It feels strange to have this switch case in a way that does not cover all three flow cases.

We can do this is because the "default" state of the component is email verification flow, but that is actually a brittle approach -- we could easily modify the defaults not realizing that the default isn't a generic pre-initialization state but is actually an email-initialization state.

I think this is the appropriate time to fix this.

Suggestions:

1. Remove the `this.verifyingEmail` and `this.verifyingPhone` attributes completely -- there is no valid use case where both can be true.  Instead we would have *ONLY* `currentVerifyFlow` and the renderer would be updated to look at that variable instead.

2. Remove `formTitle` -- instead, `src/app/auth/components/verify/verify.component.html` should render the appropriate title based on the current verify flow (this is already done for other aspects of the form).

3. I think this means you can remove this switch statement entirely.  Instead you would have a "sanity" check of "if none then redirect because that shouldn't be the case"

slifty (changes requested):
Sorry for still more changes here.

Right now this component has just too many potentially conflicting state-representing-attributes.

We should just have one: `currentVerifyFlow`

The value of that single attribute should determine what language is rendered, what flows are followed, etc.

Also I looked at the rest of the component and it looks like there is a bit of redundant logic which hasn't been updated.  (for instance: `onSubmit` has `return verifyPromise` which has a bunch of logic around form titles and other things.).  I couldn't add a comment insline since it's not part of the diff, but we should got through the entire component and clean it up / remove redundant logic and shift to `currentVerifyFlow`

crisnicandrei:
@slifty thanks Dan! This is such a great suggestion! Sorry I didn't catch it earlier

slifty (approved):
I really appreciate the iteration on this!  Code is looking to be in a great place, and this is ready for QA

k8lyn6 (approved):
This works great! I tested:

- Adding my phone number, immediately verifying phone number, then verifying email
- Adding my phone number, verifying my email, then verifying my number
- Verifying my email, adding my phone number, then verifying my phone number
- Changing my phone number and reverifying it

I did notice while testing that the verify phone number button is clickable even if the phone number field is blank, which drops you into phone verification with an error. We should probably disable that field if the phone number is blank. I created a ticket [here.](https://permanent.atlassian.net/browse/PER-10193) I also created a [ticket](https://permanent.atlassian.net/browse/PER-10194) for a small bug that makes the "verified" status persist even though there's no phone number entered. 

If you want to pick up either of these, you're welcome to @crisnicandrei ! They aren't particularly urgent though, so we can wait for the next cross team grooming to discuss them.

PR #562, opened 2025-05-05
hide any tooltip if no value is being selected.

crisnicandrei:
This pull request displays the file controls, but greyes them out if no item is selected.


slifty (approved):

k8lyn6 (approved):
Works perfectly! I tested for files and folders, and tested different access levels to make sure they all work as expected. Everything did. Thanks @crisnicandrei !

PR #564, opened 2025-05-09
Per saving or viewing tags fails

crisnicandrei:
Steps to test: 
Upload a record
Add a tag
It should display the tag


k8lyn6 (approved):
This appears to fix the issue! Checked existing files with existing keywords and custom metadata, as well as existing files that did not have keywords/custom metadata, and uploaded a new file to test with as well. All tests worked. :)

PR #565, opened 2025-05-12
PER-10187-pdfs-and-images-not-displaying-in-viewer

crisnicandrei:
Replace instances of FileVOs to files for pdfs and images. Also changed the instances of fileURL to fileUrl to match the new backend.


src/app/models/record-vo.ts
```diff
@@ -236,4 +236,5 @@ export interface RecordVOData extends BaseVOData {
   ShareVOs?: any;
   AccessVO?: any;
   isFolder?: boolean;
+  files?: any;
```
slifty, line 239: A few things here.

1. I think we are starting to wade into an issue where we have two types of model: Stela model and Permanent model.

To some extent this speaks to the importance of the perm SDK which translates both to a common interface / does not force the client to understand our internally intended mapping.

I'm uncomfortable having a single type here that is being BOTH a RecordVO and a Stela Record.  We now have both `file` and `FileVOs` in one place; every time a recordVo is accessed we have to account for the potential dual state, which results in the kinds of issues we're seeing with tags.

I think we need to (A) Fix these bugs but then (B) step back and more seriously address the deeper problem, which is that we need to be very intentional and clean regarding our type definitions.

2. In the short term though: can we remove `FileVOs?` from this type?  Do those *ever* exist on a record nay longer?

3. This is currently marked as an `any` -- we should never allow `any` (this should be linter enforced, but I understand why we may have needed to allow it for legacy code)

slifty (changes requested):
We should make sure that we're making these changes in a way that *simplifies* our types where possible.

Also we can't use `any`

src/app/models/record-vo.ts
crisnicandrei: @slifty what is the difference between a PErmanent model and a Stela model?

crisnicandrei:
@slifty replaced all occurences of FileVOs to files (even in the json files even though im nt sure it was needed)

slifty (approved):
Code looks good for the purpose of this PR and solving this issue.

That said: there is a broader "red flag" which is that it appears we are fusing Stela's `Record` with the PHP API's `RecordVO` into a single type.  This is a bit dangerous and is, I believe, part of why some of these issues made it to production.

Once we have fixed the user-facing issues I think we need to talk about PHP API types vs Stela types and adopt a consistent implementation.

src/app/models/record-vo.ts
slifty: @crisnicandrei Basically the PHP API returns a RecordVO, but Stela returns a more streamlined "Record"

The shape and of each object is different, and we should treat them as distinct types of entity.

k8lyn6 (changes requested):
Hey @crisnicandrei! It looks like this potentially fixes the issue for new files uploaded to Permanent, but does not resolve it for existing files. When I look at a docx file or PDF (both of which are converted to PDFs) that was previously uploaded to Permanent, I get an error that says "Access Denied." I've attached a screenshot here.
Screenshot 2025-05-14 at 12 03 39 PM

crisnicandrei:
@k8lyn6 this might be tricky to fix because I don't have accounts with pdfs or docs locally. Does the zooming work or not on old records that are images?

k8lyn6:
@crisnicandrei ah whoops, I forgot zooming was in here too. Let me redeploy and check. Would having access to an archive with those docs on dev be helpful at all? If not, I'll bring this conversation out to Slack to ask Dan and Liam for assistance.

k8lyn6:
@crisnicandrei added you to an archive with those files on dev!

k8lyn6:
@crisnicandrei as for the zoom, it looks like I'm getting an error when viewing some (but not all?) images. I'm not entirely sure which images are or are not working, but I figured I'd go ahead and tell you if you have time to look at it while I'm running a webinar. If not, I'll do some more investigation when I return! Screenshot attached.

Screenshot 2025-05-14 at 12 27 21 PM

k8lyn6 (changes requested):
@crisnicandrei okay, I went back through now that we know the image errors are likely caused by expired URLs, and zooming works great! Thanks for addressing that too.

I'm still seeing the "Access denied" error for PDFs that were previously uploaded to Permanent before this change. This error could also be from expired URLs, but I tested on staging too with some newer files (from 2023 rather than like 2022 or 2021 like I have on dev) and got the same error. Now that you have access to an archive with PDFs on dev, could you take a look and see if you can figure out what's causing that error? If you can't figure it out, we'll talk to the rest of the engineering team for more support. Thank you!

crisnicandrei:
@k8lyn6 in which folder do I find these pdfs? And is the branch still deployed to dev? Because a 2025 pdf isn't being displayed as before now

k8lyn6:
@crisnicandrei they should be in Testing_files_one_of_each_type, then under "pdf" and "document" (the document folder has a file that was uploaded as a word doc that should now be a pdf after our processing). Neither of those appear to be working at the moment. Dev has the main branch deployed, but I'll go ahead and start the deployment for this branch so you can take a look! It should be ready to go in a few minutes.

crisnicandrei:
@k8lyn6 let me know when it is deployed please!

k8lyn6:
@crisnicandrei it's up now!

crisnicandrei:
@k8lyn6 it seems to be the same issue! Send the url on Slack!

k8lyn6 (approved):
I'm going to go ahead and merge this PR. I believe this PR has solved the original issue, but it is difficult to fully test because of the discovery of [this issue.](https://permanent.atlassian.net/browse/PER-10192?atlOrigin=eyJpIjoiOTBhNDE3YzcyNjgyNGQwZmFkNDQyZmRiZTg3MmVmMzEiLCJwIjoiamlyYS1zbGFjay1pbnQifQ) It appears that previously uploaded PDFs I'm seeing on dev are still failing because of the issue in the linked ticket. I don't have access to any other PDFs, and newly uploaded PDFs work now as intended. 

I'm making the decision to approve this because at the very least, it did fix newly uploaded PDFs which solves half the problem. If additional work is still needed to fix the issue we're observing on dev, we can make a new ticket.

PR #566, opened 2025-05-13
PER-10188-location-not-displaying

crisnicandrei:
Display locations for records after the v2 change

@slifty heads up I did not remove LocnVOs yet from the codebase because the profile also  has them :)

### DIFF
--- a/src/app/file-browser/components/file-viewer/file-viewer.component.html
+++ b/src/app/file-browser/components/file-viewer/file-viewer.component.html
@@ -166,7 +166,7 @@
           <td class="location-container" (click)="onLocationClick()">
             <pr-static-map
               [location]="currentRecord.LocnVO || currentRecord.location"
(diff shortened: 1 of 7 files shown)


slifty (approved):

k8lyn6 (changes requested):
This works for setting a new location! I'm able to set a location and it remains correct when refreshed. Previously uploaded files that already had an existing location also are showing up correctly now (yay!). 

However, all the previously uploaded files that did not have a location show a google maps error when I look at the metadata (see attached). Could you investigate, Andrei? It seems like this was introduced in this PR as I don't think this issue is happening on prod currently.
Screenshot 2025-05-14 at 3 16 38 PM

k8lyn6 (approved):
Sorry for the delay this morning! This fixed the issue with the map showing up when there was no location. Thanks @crisnicandrei !

PR #570, opened 2025-05-16
Revert "PER-10187-pdfs-and-images-not-displaying-in-viewer"

slifty:
Reverts PermanentOrg/web-app#565


slifty:
We're using this commit in #571

PR #571, opened 2025-05-16
Revert recent v2 changes

slifty:
This PR reverts a series of recent PRs we made related to updating `get` and `getWithChildren` to use v2 endpoints.

Those v2 endpoints were returning different data than the PHP api (v1).  That was resulting in various issues, and in fixing those issues we believe we have introduced problems for v1-styled-data.

This gets us back to the pre-v2-migration state, and will allow us to make those changes in a type-safe way.


liam-lloyd (approved):
I don't see any problems here, though it is possible we're reverting more than we strictly need to. In particular, the commit changing how folder children are retrieved should not have altered the internal structure of the response, because it's not calling `stela`. It is instead making a V2 request to `back-end`'s `/folder/getWithChildren`. V2 requests remove some of the wrappers that `back-end` typically puts around responses, but the underlying `FolderVO` should be unchanged (alternatively, if this change _is_ causing bugs, it may point to a backend issue).

k8lyn6 (commented):
I tested viewing images and PDFs, as well as checking for existing and new tags, custom metadata, and locations, and all looked good. I also did a pretty thorough review of all major functionality on the site, and I don't see any issues that aren't reproducible on prod (to be specific, I did find a bug but it seems unrelated to this pull request, and doesn't appear to be a major issue. I've documented it on Jira).

slifty:
@liam-lloyd ty and good note regarding the getWithChildren v2 change.

PR #573, opened 2025-05-16
Revert "PER-9419-call-v2-get"

slifty:
Reverts PermanentOrg/web-app#551


slifty:
We're using this commit in #571

PR #574, opened 2025-05-27
PER-10200 Web app display error on multi part uploads

crisnicandrei:
Steps to test:

1. Upload a large file (200MB approx)
2. No error should be displayed after uploading


src/app/core/services/upload/uploader.ts
```diff
@@ -132,7 +133,7 @@ export class Uploader {
       eTags,
     );
 
-    const record = response.getRecordVO();
+    const record = response as unknown as RecordVO;
```
slifty, line 136: This is overriding type safety (if record was a number, a string, a null, etc it would not be treated as a RecordVO)

Is there a reason we're switching to this instead of `.getRecordVO()` as part of this PR?

slifty (changes requested):
I'm a little confused -- I don't see any actual changes to the uploader logic (just a change that forcefully overrides typing, and the addition of a test, neither of which would affect runtime).

What was the runtime issue that was resulting in errors?

src/app/core/services/upload/uploader.ts
crisnicandrei: @slifty the response is a Record, not a RecordVO, and you can see it's only used inside the dispatch, otherwise the response is being returned. Naturally Record objects do not have .getRecordVO()

slifty: Looking into this, response came from a call to `registerMultipartRecord`

The type definitions of this function are wonky:

```

  public async registerMultipartRecord(
    record: RecordVO,
    uploadId: string,
    key: string,
    eTags: string[],
  ): Promise<RecordResponse> {
    return getFirst(
      this.httpV2.post('/record/registerRecord', {
        displayName: record.displayName,
        parentFolderId: record.parentFolderId,
        uploadFileName: record.uploadFileName,
        size: record.size,
        multipartUploadData: {
          uploadId,
          key,
          parts: eTags.map((ETag, index) => ({
            PartNumber: index + 1,
            ETag,
          })),
        },
      }),
    ).toPromise() as unknown as RecordResponse;
  }
  ```
  
For one thing, the return type is being force-cast as a `RecordResponse`... is that true?  If not, we should adjust the type definitions.

Related: it looks like `post` takes a generic to indicate what the return type is intended to be, so we should probably be doing `.post<RecordResponse>`... though I am not seeing type errors regardless (e.g. if I say `.post<number>` it doesn't yell at me, so I'm not sure if type safety is actually being tracked in the first place here).

slifty: Andrei and I paired this morning on this and this is now resolved.

slifty (approved):
Looks great thank you!

And thank you again for the pairing this morning to talk through the typing.

omnignorant (approved):
I attempted two uploads of ~250 and ~380 GB. The first time, I got an error but they uploaded to the archive nevertheless. I refreshed the browser window wondering whether I had the most recently deployed app loaded and made a second attempt. In the second attempt, both completed successfully AND without error so I think we are OK. Detailing my entire experience for completeness sake.

PR #576, opened 2025-06-03
Per 10107 implement unlisted share page UI

crisnicandrei:
This pull request adds the new API changes and api fields to the share modal in the web-app. This also creates a new share link which displays the items in the preview view as before, but does not open them in the file viewer, because I will make a separate pull request for that, so this one doesn't get too large. 

Previously I did it in this pull request, but the new v2 api makes it a bit more challenging so I had to drop that idea and will move it into another pr which will be branched off of this one.


cecilia-donnelly:
I believe this has been superseded by #630, #631, and #633.

cecilia-donnelly:
This should be merged after #633.

slifty:
In the same spirit of "opening a new PR" I'm going to close this and open one directly -- also will be rebasing against PER-10107-share-link-settings-component which has some updates.

PR #577, opened 2025-06-03
PER-10096-archive-invitations

crisnicandrei:
How to test:
1. Have account A invite account B to Permanent (Account B must have an account)
2. Account B must receive an email whilst logged in and click the ling
3. Account B should be redirected to the pending archives modal


slifty (approved):
Code looks good -- ready for QA testing!

omnignorant (commented):
This does not appear to be working as described. I invited an *existing* account B to be a member of my archive from account A. I got an email from account A to account B with a link, then copy/pasted it to a browser window where account B was logged in. I was not directed to pending archives. See attached video:
* account A = robert@permanent.org
* account B = robert+prmnttst1150@permanent.org
* URL = https://dev.permanent.org/app/pr/manage?fullName=Um9iZXJ0IFRlc3QxMTUw&primaryEmail=cm9iZXJ0K3BybW50dHN0MTE1MEBwZXJtYW5lbnQub3Jn

I did not attempt to QA this for a member invite to a non-existant account since that should take the new account through onboarding to accept the invite.

https://github.com/user-attachments/assets/69b4cc08-4b3f-4b56-957d-3b1aeb247473

crisnicandrei:
@omnignorant you are right, when I implemented this the email system was not working on my side. Should I implement this only on the frontend or should we change the backend aswell with the correct url taking you to the pending archive?

CC: @slifty @liam-lloyd

liam-lloyd:
> @omnignorant you are right, when I implemented this the email system was not working on my side. Should I implement this only on the frontend or should we change the backend aswell with the correct url taking you to the pending archive? The curent email url only has 2 parameters which maybe in the future could also have other purposes, and I think the frontend should not handle this on its own, but I might be wrong and I am open to suggestions.
> 
> CC: @slifty @liam-lloyd

I think it does make sense for the URL to specify which tab in the modal should be selected, and to update the link the backend includes in this email to reflect that.

crisnicandrei:
@liam-lloyd this is how the url looks like to get redirected to pending archives: https://local.permanent.org/app/(private/0004-0006/75//dialog:archives/pending)

PR #582, opened 2025-06-19
Create file viewer v2

crisnicandrei:
This Pull request is the second part of the new unlisted share functionality

This adds a new file-viewer which handles v2 api calls and a file list v2 component


crisnicandrei:
@cecilia-donnelly yes it is

cecilia-donnelly:
This is a followup to #576.

slifty:
Now that we are shimming v2 calls to appear to our components the same as v1 calls these components shouldn't be needed!

Issue #590, opened 2025-07-09
Update to Angular 20


slifty:
Now that depedabot is enabled we are getting some pretty big version upgrades, which then also conflict with our current version of angular (17)

Angular 17 is no longer supported; let's update to Angular 20.

PR #606, opened 2025-07-11
Update to Angular 18

slifty:
This PR moves the front-end code base to Angular 18.  It's part of a larger set of transitions which will ultimately culminate in Angular 20.

I separated the commits into pieces to reflect the "stages" of the move to Angular 18.  Once the code is reviewed my plan would be to combine "ng build works" through to the end so there is a single "Angular 18" update which makes all of the needed migration changes in one place.

Related to #590


slifty:
@cecilia-donnelly @liam-lloyd For this PR what I'm thinking is...

I have the PR itself broken into commits where not every commit stands on its own, but it shows the progress of the migration.

I can then squash appropriately before merge it so there's a single "commit" for "update to Angular 18"

At any rate... I think this is ready for review.

.eslintrc.js
```diff
@@ -152,6 +152,7 @@ module.exports = {
         'vars-on-top': 'off',
         yoda: 'off',
         'prettier/prettier': 'off',
+        'import/order': 'off',
```
slifty, line 155: Looking at this PR: I'm gonna remove this commit and fix the offending lines instead!

slifty:
@liam-lloyd fixed up (separate commit since I'm not entirely sure which introduced the issue, but again I'll be squashing)

liam-lloyd (approved):

slifty:
Closing the loop on the above: we aren't able to reproduce this issue at this point, but will be doing another round of tests for the Angular 19 MR

PR #607, opened 2025-07-14
PER-10223 Fix Timeline navigation

crisnicandrei:
The timeline view was using dialogs which were preventing navigation. Remove the dialog and use the Angular Router instead

Steps to test:

1. Open a folder in timeline view that contains records.
2. Click a record
3. It should open it in the zoomable file viewer.

### DIFF
--- a/src/app/views/components/timeline-view/timeline-view.component.html
+++ b/src/app/views/components/timeline-view/timeline-view.component.html
@@ -56,3 +56,4 @@
     <p [innerHTML]="timelineRootFolder.description"></p>
   </div>
 </div>
+<router-outlet></router-outlet>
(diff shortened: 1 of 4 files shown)


slifty (dismissed):
Let's remove that use of `any`

liam-lloyd:
> @cecilia-donnelly is it possible for us to move this PR along during Dan's absence? It is waiting on his approval and then I guess my QA probably.

I've approved it, so it should be ready for QA now. And we will be able to deploy it past Dan's request for changes (which has been addressed) once it passes QA.

omnignorant:
@liam-lloyd ok then, I added myself to make sure I keep track of it.

PR #608, opened 2025-07-16
PER-10193- Disable verify phone number if no phone number

crisnicandrei:


slifty (approved):

omnignorant (approved):
It was not specified in the description, but this is only true for a new account. There is a separate issue of "verified" persisting even when there isn't a phone number for existing accounts (attached). BUT this issue is addressed for new accounts (attached). 
Screenshot 2025-07-21 at 11 51 24 AM
Screenshot 2025-07-21 at 11 48 43 AM

PR #613, opened 2025-07-21
PER-10194-make phone unverified

crisnicandrei:
When removing the phone number, set the status to unverified

Steps to test:

1. Have an account with a verified phone number
2. Remove the phone number
3. It should set the status to unverified


cecilia-donnelly (approved):
LGTM! Ready for QA.

aasandei-vsp (approved):

PR #620, opened 2025-07-23
Upgrade to Angular 19

slifty:
This PR updates us to Angular 19.

This results in some warnings around deprecation but I think a lot of these are going to be addressed through the linting project that can start after this upgrade.

Related to #590


cecilia-donnelly:
This is one that could use general QA @omnignorant since it's an overall upgrade.

omnignorant:
@cecilia-donnelly does this still require my attention after the work completed today?

PR #622, opened 2025-07-23
Modernize our linting

### LINKED ISSUE #591: Upgrade to eslint 9
We're doing this with our other projects -- let's do it here too.

Eslint 8 is no longer supported.  This will mean changing our lint ruleset to `love` as well.

slifty:
This PR does several very important things:

1. It updates ESLint to v9, which is a major change that requires a change to our configuration format (ESLint 8 is no longer maintained).
2. It moves our lint configuration sets to begin to align with our other project.  There is a lot of work still to be done to actually enable the majority of the lint rules these sets provide, but the current status at least creates that road map for us with a clear path forward
3. It adds more robust linting (and formatting) scripts for us to use, and also updates our CI to use them.
4. It moves our project to `tabs` instead of `spaces.  This is in line with all of our other typescript repositories and also in line with the standard that Prettier has created.

The byproduct of all this is that essentially every single line of this repository has been reformatted; there are NOT any actual changes to the application or any logic.  The main ramification is that open PRs are going to have gnarly merge conflicts, but this is one of those situations where "it's going to need to happen some time".

I'm happy to pair to resolve any conflicts this creates!

Resolves #591 

This should not be merged until after #620 is merged


liam-lloyd (approved):

cecilia-donnelly:
@omnignorant I am QA-ing this but tagging you too in case you have some time and want to take a look. This should only be formatting changes but it touches all of the code in the web-app.

cecilia-donnelly (approved):

cecilia-donnelly:
I "bypassed rules" because the linting was "expected" for a day - not sure why that was. The eslint check passed, though.

PR #630, opened 2025-07-25
Per 10107 api only

cecilia-donnelly:
Experimenting with breaking up one of Andrei's big PRs.


cecilia-donnelly:
@crisnicandrei can you rebase this on main? There will be a lot of linting/prettier conflicts because of the big eslint upgrade. It might be better for you to open a new PR so that you're the author. Then, can you write a description of exactly what's included in this PR? My original thought was that it would be only non-user-visible changes. Is that where it ended up?

slifty:
Closing this in preference to #669

PR #633, opened 2025-07-30
Per 10107 only v2 components

crisnicandrei:
This pull request adds the V2 components I have created for the new unlisted share links. This pull request adds the new V2 components which are going to be used with the new API calls (v2) for the unlsited shares. The look remains the same, but the api calls are different.


cecilia-donnelly:
No visible changes, but needs QA since API calls are updated.

PR #638, opened 2025-08-04
591 esline 9 rebased

cecilia-donnelly:
(This is just a redo of #622 since I forgot about the no-merges plan and merged #607 this morning! The conflicts weren't too bad since #607 was small, but I want to make sure I did them properly before I overwrite #622.)


cecilia-donnelly:
Closing this. I reverted the mistakenly merged PR instead.

PR #653, opened 2025-08-13
Upgrade to Angular 20

### LINKED ISSUE #590: Update to Angular 20

slifty:
This PR upgrades Angular to v20

Resolves #590 

It should not be merged until after #652


cecilia-donnelly:
@omnignorant this will need site-wide QA since it's another Angular upgrade. The last one for a while, though!

Issue #628, opened 2025-07-23
Enable disabled lint rules


slifty:
This project has a *lot* of disabled lint rules.  It's time for us to start turning them back on.  We have a good model for this which we've done in our other projects -- one commit per rule, with the associated changes alongside the commit.

PR #658, opened 2025-08-21
Enable lint rules (Part 1)

slifty:
This PR begins the process of enabling various linting rules that had been disabled on the front-end.

Related to #628


slifty:
I re-added the deleted test you noted (and provided an accessor to the private attribute that had led me to delete the test).

Merging after CI passes!

PR #659, opened 2025-08-21
Enable lint rules (Part 2)

slifty:
This PR continues the project of enabling various disabled lint rules.

Related to #628 

This shouldn't be merged until #658 is merged

### DIFF
--- a/src/app/file-browser/components/file-list/file-list.component.ts
+++ b/src/app/file-browser/components/file-list/file-list.component.ts
@@ -170,7 +170,7 @@ export class FileListComponent
 		});
 
 		this.visibleItemsHandlerDebounced = debounce(
-			() => this.loadVisibleItems(),
+			async () => await this.loadVisibleItems(),
 			VISIBLE_DEBOUNCE,
 		);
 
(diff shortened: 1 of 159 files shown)


src/app/file-browser/components/file-list/file-list.component.ts
```diff
@@ -346,9 +346,11 @@ export class FileListComponent
 
 		const queryParams = this.route.snapshot.queryParamMap;
 		if (queryParams.has('showItem')) {
-			const folder_linkId = Number(queryParams.get('showItem'));
+			const folderLinkId = Number(queryParams.get('showItem'));
```
cecilia-donnelly, line 349: This gives me a little pause because I know we discussed whether to rename `folder_link` to `folderLink` in some context and were worried about collisions. I don't remember the exact circumstances though. Does that sound familiar at all?

cecilia-donnelly (approved):
I can't remember an actual problem with `folder_link` / `folderLink` so approving this, but wanted to say it out loud in case it jogs your memory!

src/app/file-browser/components/file-list/file-list.component.ts
cecilia-donnelly: Right right, that makes sense! Thanks Dan. And yes, it does.

PR #661, opened 2025-08-22
Enable lint rules (Part 3)

slifty:
This PR continues the project of enabling (or removing explicit overrides) of various rules

Related to #628 
Should not be merged until #659


cecilia-donnelly (approved):
This looks good! It does look like we broke the build along the way, though, so this is also failing to build for me locally. I don't think that's a this-PR problem but we should check it once we fix the bug on main.

PR #673, opened 2025-08-27
Fix type includes for build

slifty:
With some recent lint changes we needed to add some expanded type definitions for `Window`.  These were included for things like linting, but not included for the build step which resulted in a broken build.

This fixes the tsconfig used by the build step.

### DIFF
--- a/src/tsconfig.app.json
+++ b/src/tsconfig.app.json
@@ -12,7 +12,7 @@
     "secrets.ts"
   ],
   "include": [
-    "/src/**/*.d.ts"
+    "**/*.d.ts"
   ],
   "exclude": [
     "**/*.stories.*"


src/tsconfig.app.json
```diff
@@ -12,7 +12,7 @@
     "secrets.ts"
   ],
   "include": [
-    "/src/**/*.d.ts"
+    "**/*.d.ts"
```
slifty, line 15: Note that this is loading from inside `src` since tsconfig.app.json is located there!

slifty:
@liam-lloyd build for this branch did pass: https://github.com/PermanentOrg/web-app/actions/runs/17268927019

liam-lloyd (approved):

PR #552, opened 2025-04-04
PER-10089-create-unlisted-ui

crisnicandrei:
This pull request adds the new unlisted share link UI, which is hidden behind a feature flag for now.

Steps to test: 
1. Add a new feature flag called unlisted-share and enable it
2. Go to any record and click share
3. Click "Show Link Settings"
4. The dropdown should appear

PR #697, opened 2025-09-03
Refactor share links to a separate component

slifty:
This PR creates a separate component for rendering and working with share links; this will ultimately make it easier for us to implement the unlisted share UI in future commits.

It is built on #669, and should not be merged until that is ready for merge as well.


cecilia-donnelly:
I'm not sure we need this since the share link components have already been edited in https://github.com/PermanentOrg/web-app/pull/552.

slifty:
Closing since we believe it is already implemented (or if not, it may need to be looked at with fresh eyes)

PR #707, opened 2025-09-04
[PER-10283] Change login screen graphic for legacy lab

aasandei-vsp:
https://permanent.atlassian.net/browse/PER-10283

### DIFF
--- a/src/app/auth/components/auth/auth.component.scss
+++ b/src/app/auth/components/auth/auth.component.scss
@@ -80,9 +80,9 @@
 	border-radius: 1em;
 	margin-left: 20px;
 
-@media screen and (max-width: 1440px) {
-  background-image: none;
-}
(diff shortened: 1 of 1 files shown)


crisnicandrei (approved):

cecilia-donnelly (changes requested):
Thank you @aasandei-vsp ! Can you make the text bigger? I know it's hard to match without an official design, but the fonts for the date and explanatory text should be closer to the title size - if you look at the example image, the text inside the circle image is about the same size as the list items and general explanation. On my screen, at least, those look smaller in this PR.

I did try different responsive screen sizes and at a certain point the explanatory text overlaps the image, but I think that's probably fine since it only appears to affect a specific (odd) screen size.

I love the action button - it looks beautiful!

aasandei-vsp:
We will not advertise this now, but do it in March.

PR #765, opened 2025-10-14
Create/edit/delete share links

cecilia-donnelly:
The first commit works to edit share links, but it has some issues with the initial creation of a share link. That is:
1. Open the sharing dialog
2. Create a share link
3. See that editing is unreliable until you close and reopen the modal.

This is most likely due to the old and new share link objects *both* existing in the component. To fix that, I started working on using the new API for creating and deleting share links, too. That is in the second commit here. The API calls are working fine, but the UI is not picking up the changes correctly. Maybe we can look at this together?

There are other odds and ends - notably the expiration date needs to be fixed and I need to move all the variable names to match the new share link object fields.

### DIFF
--- a/src/app/file-browser/components/sharing-dialog/sharing-dialog.component.spec.ts
+++ b/src/app/file-browser/components/sharing-dialog/sharing-dialog.component.spec.ts
@@ -13,7 +13,7 @@ import { ArchiveVO, RecordVO, ShareVO } from '@models';
 import { RelationshipService } from '@core/services/relationship/relationship.service';
 import { ApiService } from '@shared/services/api/api.service';
 import { ShareResponse } from '@shared/services/api/share.repo';
-import { AccessRoleType } from '@models/access-role';
+import { AccessRoleType, PermissionsLevel } from '@models/access-role';
 import { MessageService } from '@shared/services/message/message.service';
 import { Shallow } from 'shallow-render';
 import { NgModule } from '@angular/core';
@@ -251,6 +251,88 @@ describe('SharingDialogComponent', () => {
 		expect(confirmSpy).toHaveBeenCalled();
 		expect(component.shares.length).toBe(1);
 		expect(component.shares[0].accessRole).toBe(shareViewer.accessRole);
+	}));
+
+	it('should remove share link when confirmed', fakeAsync(() => {
+		const mockShareLink = {
+			id: '123',
+			token: 'test-token',
+			permissionsLevel: 'viewer' as PermissionsLevel,
+			accessRestrictions: 'none' as 'none' | 'account' | 'approval',
+			expirationTimestamp: null,
+			createdAt: new Date(),
+			updatedAt: new Date(),
+			itemId: '1',
+			itemType: 'record' as const,
+			maxUses: null,
+			usesExpended: null,
+		};
+
+		component.newShareLink = mockShareLink;
+		component.showLinkSettings = true;
+
+		const confirmSpy = spyOn(
+			TestBed.inject(PromptService),
+			'confirm',
+		).and.returnValue(Promise.resolve(true));
+
+		const deleteShareLinkSpy = spyOn(
+			TestBed.inject(ShareLinksApiService),
+			'deleteShareLink',
+		).and.returnValue(Promise.resolve());
+
+		component.removeShareLink();
+
+		tick();
+
+		expect(confirmSpy).toHaveBeenCalledWith(
+			'Remove link',
+			'Are you sure you want to remove this link?',
+			jasmine.any(Promise),
+			'btn-danger',
+		);
+
+		expect(deleteShareLinkSpy).toHaveBeenCalledWith('123');
+		expect(component.newShareLink).toBeNull();
+		expect(component.showLinkSettings).toBe(false);
+	}));
+
+	it('should not remove share link when cancelled', fakeAsync(() => {
+		const mockShareLink = {
+			id: '123',
+			token: 'test-token',
+			permissionsLevel: 'viewer' as PermissionsLevel,
+			accessRestrictions: 'none' as 'none' | 'account' | 'approval',
+			expirationTimestamp: null,
+			createdAt: new Date(),
+			updatedAt: new Date(),
+			itemId: '1',
+			itemType: 'record' as const,
+			maxUses: null,
+			usesExpended: null,
+		};
+
+		component.newShareLink = mockShareLink;
+		component.showLinkSettings = true;
+
+		const confirmSpy = spyOn(
+			TestBed.inject(PromptService),
+			'confirm',
+		).and.returnValue(Promise.reject(false));
+
+		const deleteShareLinkSpy = spyOn(
+			TestBed.inject(ShareLinksApiService),
+			'deleteShareLink',
+		).and.returnValue(Promise.resolve());
+
+		component.removeShareLink();
+
+		tick();
+
+		expect(confirmSpy).toHaveBeenCalled();
+		expect(deleteShareLinkSpy).not.toHaveBeenCalled();
+		expect(component.newShareLink).toEqual(mockShareLink);
+		expect(component.showLinkSettings).toBe(true);
 	}));
 });
 


cecilia-donnelly:
@aasandei-vsp, since I'm out tomorrow I would appreciate it if you could give this an initial review either Friday or Monday. I have to fix the tests, of course, but I think the actual functionality is working at this point. I'll leave the different commits for now so you can see my thought process.

There are some choices I am concerned about and would like your eyes on:
1. I needed some response from the `DELETE` endpoint and couldn't find the right place to get the status - maybe you know! I wrote more detail about this in the commit message for a9dd955ceb4.
2. I should add a call to https://permanentorg.github.io/stela/#tag/records/operation/get-record-share-link and the folder equivalent instead of double calling in the edit service. (Currently I call the old API to get the share link's id and then call the new API to get the share link object.)
3. I'm not thrilled with how I've handled the conversion between the UI elements (restricted/unrestricted, auto approve etc) and the share link object. If you have ideas to improve that, let me know!

src/app/file-browser/components/sharing-dialog/sharing-dialog.component.html
```diff
@@ -243,13 +237,12 @@
 							{{ linkCopied ? 'Copied' : 'Copy' }}
 						</button>
 					}
-					@if (!shareLink) {
+					@if (!newShareLink) {
 						<div class="share-link-container">
-							Generate a link to send via text or email to invite people to
-							share this content.
+							Generate a link to invite people to share this content.
```
aasandei-vsp: Just nit pick, but I am not sure `to invite people to share` sounds great, but maybe that's just my personal opinion and please take into consideration english is not my first language. :sweat_smile:

src/app/file-browser/components/sharing-dialog/sharing-dialog.component.ts
```diff
@@ -415,14 +471,14 @@ export class SharingDialogComponent implements OnInit {
 		}
 	}
 
-	getExpirationFromExpiresDT(expiresDT: string): Expiration {
-		if (!expiresDT) {
+	getExpirationFromExpirationTimestamp(expirationTimestamp: Date): Expiration {
+		if (!expirationTimestamp) {
```
aasandei-vsp: also move in mapping service

src/app/file-browser/components/sharing-dialog/sharing-dialog.component.ts
```diff
@@ -172,12 +179,61 @@ export class SharingDialogComponent implements OnInit {
...
+				return 'manager';
+			case 'access.role.owner':
+				return 'owner';
+			default:
+				return 'viewer';
+		}
+	}
+
+	permissionsLevelToAccessRole(permissionsLevel: string): AccessRoleType {
+		switch (permissionsLevel) {
```
aasandei-vsp: same as above, enum seems to be the way to go.

src/app/file-browser/components/sharing-dialog/sharing-dialog.component.ts
```diff
@@ -541,11 +627,29 @@ export class SharingDialogComponent implements OnInit {
 
 	async onShareLinkPropChange(propName: ShareByUrlProps, value: any) {
 		this.updatingLink = true;
+		let update: Partial<ShareLink> = {};
+
 		try {
-			const update = new ShareByUrlVO(this.shareLink);
-			update[propName] = value;
-			await this.api.share.updateShareLink(update);
-			this.shareLink[propName] = update[propName];
```
aasandei-vsp: looks like this could also be moved in a mapping service, but without the call to BE.

src/app/file-browser/components/sharing-dialog/sharing-dialog.component.html
cecilia-donnelly: You're right, I must have been looking at a different version of the design! Or not paying enough attention. Updated to match the text here: https://www.figma.com/design/AYxFdjQVLuK2iapWciBAJs/Permanent-Features?node-id=681-382&p=f&t=1wwjRuWU7ydSCP7R-0

src/app/file-browser/components/sharing-dialog/sharing-dialog.component.html
```diff
@@ -271,66 +264,24 @@
...
+						</select>
 					</div>
 				}
-				@if (showLinkSettings && shareLink) {
+				@if (showLinkSettings && newShareLink) {
 					<div class="share-link-settings" @ngIfScaleAnimation>
-						@if (linkType === 'private') {
-							<div class="link-settings-item">
-								<label>Share Preview</label>
-								<div class="form-check form-check-inline">
```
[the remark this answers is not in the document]
cecilia-donnelly: We are getting rid of the preview toggle, yes. The idea was that it's just a less full featured version of unlisted shares. It does not exist in the new endpoint / new model but I don't think we've done a full design of how it should behave for old links, so it will stay in the database. Thanks for checking! I'll make a ticket to follow up on this.

src/app/file-browser/components/sharing-dialog/sharing-dialog.component.ts
```diff
@@ -153,7 +160,7 @@ export class SharingDialogComponent implements OnInit {
 
 		this.shareItem = this.data.item as ItemVO;
 
-		this.displayDropdown = feature.isEnabled('unlisted-share');
+		this.displayDropdown = true;
```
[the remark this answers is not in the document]
cecilia-donnelly: Thank you!

cecilia-donnelly:
@slifty, requesting a review from you since Andreea is out the rest of the week. I believe I've addressed her comments. Thanks!

src/app/share-links/services/share-link-mapping.service.ts
```diff
@@ -0,0 +1,81 @@
...
+		}
+
+		const diff = differenceInHours(
+			new Date(expirationTimestamp),
+			new Date(createdAt),
+		);
+
+		if (diff <= 24 * ExpirationDays.Day) {
+			return Expiration.Day;
+		} else if (diff <= 24 * ExpirationDays.Week) {
```
slifty: This was initially confusing to me because we're using an enum as a numeric value for a formula (as opposed to a value representing a particular object state that can be referenced for logical flow, which I believe is closer to the intention of an enum)

I'd encourage we treat these as distinct concepts and have named constants to represent the values we actually want to use as values / in a calculation. 

For instance, this would mean:

```
const ONE_WEEK_IN_DAYS = 7

...

if (diff <= ONE_WEEK_IN_DAYS)
```

And, in fact, we might consider being more direct entirely and saying:

```
const ONE_WEEK_IN_HOURS = 168;

...

const hoursToExpiration = differenceInHours(
  new Date(expirationTimestamp),
  new Date(createdAt),
);

if (hoursToExpiration <= ONE_WEEK_IN_HOURS) {
  return Expiration.Week
}
```

^ note this also renames `diff` to something more meaningful, and now I think the if block is really clear around what exactly it's attempting to accomplish.

src/app/share-links/services/share-link-mapping.service.ts
```diff
@@ -0,0 +1,81 @@
...
+		} else if (diff <= 24 * 7 * ExpirationDays.Month) {
+			return Expiration.Month;
+		} else if (diff <= 24 * 7 * ExpirationDays.Year) {
+			return Expiration.Year;
+		} else {
+			return Expiration.Other;
+		}
+	}
+
+	getExpiresDTFromExpiration(
```
slifty: This isn't a change request; just an observation: having to duplicate this logic on the front-end (where it could drift from the actual logic used by the back-end) is a signal to me that we might want to consider having the expiration dateTime as a read-only attribute that is a stored at the schema level as a generated column.

Alternatively, the day / month / year / { custom date } could even be an *data entry UX level concept* and the ONLY thing actually stored in the database is the actual resolved expiration value.

I also imagine this approach would make queries simpler, and provide broader long term flexibility because we could change the UX without impacting schema / data structure (e.g. if we just store when the record expires we can do whatever UX we want to render the value meaningfully and collect the value simply)

slifty (commented):
Found a few things that probably should become enums, and some thoughts around typing related to mapping functions.

slifty (approved):
Understood that we're not using this moment to make major new patterns / changes overall (e.g. making a swath of new enums); given that, looks good I think!

PR #766, opened 2025-10-15
[PER-10314] Folder in folder unlisted share preview

aasandei-vsp:
https://permanent.atlassian.net/browse/PER-10314

Steps to test:
1. Create an account or use an existing one and log in
2. Create a folder that contains other folders and files, inside the folders, create more folders, so you can have a tree structure like of folders and files.
3. Create share link for the "root" folder, the one you created first
4. Make it unlisted(either via API in Postman or directly in the DB)
5. Open a tab in incognito mode and paste the generated link
6. Go through folders by single clicking them and also preview files(single click)
EXPECTED: Being able to navigate forward inside the folder structure.
7. After you have clicked several folders, click a folder name in the breadcrumbs.
EXPECTED: The breadcrumbs cut to that specific folder and you get the see the folder/files inside it

### DIFF
--- a/src/app/file-browser/components/file-list-item/file-list-item.component.ts
+++ b/src/app/file-browser/components/file-list-item/file-list-item.component.ts
@@ -339,7 +339,6 @@ export class FileListItemComponent
 	}
 
 	ngOnDestroy() {
-		this.dataService.unregisterItem(this.item);
 		unsubscribeAll(this.subscriptions);
 	}
 
@@ -560,7 +559,6 @@ export class FileListItemComponent
 			if (this.isUnlistedShare) {
 				this.itemClicked.emit({
 					item: this.item,
-					event: event as MouseEvent,
 					selectable: false,
 				});
 				return;
--- a/src/app/file-browser/components/file-list/file-list.component.ts
+++ b/src/app/file-browser/components/file-list/file-list.component.ts
@@ -92,6 +92,10 @@ export class FileListComponent
 	@ViewChildren(FileListItemComponent)
 	listItemsQuery: QueryList<FileListItemComponent>;
 
+	/**
+	 * currentFolder represents the shared folder in case of unlisted shares and
+	 * the rendered folder when the component is used in private/public/shared features
+	 */
 	currentFolder: FolderVO;
 	listItems: FileListItemComponent[] = [];
 
@@ -106,6 +110,11 @@ export class FileListComponent
 	public showFolderThumbnails = false;
 
 	@Input() allowNavigation = true;
+
+	/**
+	 * the ephemeralFolder is useful for using the component as is, without needing to change the route
+	 * also this will preserve the currentFolder, where the navigation by click starts from
+	 */
 	@Input() ephemeralFolder: FolderVO = null;
 
 	@Output() itemClicked = new EventEmitter<ItemClickEvent>();
@@ -116,6 +125,7 @@ export class FileListComponent
 	private reinit = false;
 	private inFileView = false;
 	private inDialog = false;
+	private isUnlistedShare = false;
 
 	@ViewChild('scroll') private scrollElement: ElementRef;
 	visibleItems: Set<FileListItemComponent> = new Set();
@@ -205,7 +215,7 @@ export class FileListComponent
 	}
 
 	async ngOnChanges() {
-		this.initializeFolder();
+		this.syncStateWithRouteandInput();
 	}
 
 	registerArchiveChangeHandlers() {
@@ -323,15 +333,15 @@ export class FileListComponent
 		}
 	}
 
-	refreshView() {
-		this.ngOnInit();
+	async refreshView() {
+		await this.ngOnInit();
 		setTimeout(() => {
 			this.ngAfterViewInit();
 		}, 1);
 	}
 
 	async ngOnInit() {
-		this.initializeFolder();
+		await this.syncStateWithRouteandInput();
 	}
 
 	ngAfterViewInit() {
@@ -532,7 +542,7 @@ export class FileListComponent
 		}
 
 		const itemsToFetch = visibleListItems.map((c) => c.item);
-		if (itemsToFetch.length && !this.shareLinksService.isUnlistedShare()) {
+		if (itemsToFetch.length && !this.isUnlistedShare) {
 			await this.dataService.fetchLeanItems(itemsToFetch);
 		}
 	}
@@ -546,7 +556,9 @@ export class FileListComponent
 		}
 	}
 
-	private initializeFolder() {
+	private async syncStateWithRouteandInput() {
+		this.isUnlistedShare = await this.shareLinksService.isUnlistedShare();
+
 		this.currentFolder =
 			this.ephemeralFolder || this.route.snapshot.data.currentFolder;
 		this.showSidebar = this.route.snapshot.data.showSidebar;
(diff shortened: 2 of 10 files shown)


src/app/file-browser/components/file-list-item/file-list-item.component.ts
```diff
@@ -556,6 +556,14 @@ export class FileListItemComponent
 		}
 
 		if (this.item.isFolder) {
+			if (this.isUnlistedShare) {
+				this.itemClicked.emit({
+					item: this.item,
+					event: event as MouseEvent,
```
slifty: Where is `event` defined?  It doesn't appear to be a parameter of `goToItem` so I imagine this is tapping into the (deprecated) `event` that is globally defined by the DOM which we shouldn't do.

src/app/shared/components/breadcrumbs/breadcrumbs.component.ts
```diff
@@ -182,8 +184,8 @@ export class BreadcrumbsComponent implements OnInit, OnDestroy {
 				new Breadcrumb(
 					rootUrl,
 					folder.pathAsText[i],
-					folder.pathAsArchiveNbr[i],
-					folder.pathAsFolder_linkId[i],
+					folder.pathAsArchiveNbr && folder.pathAsArchiveNbr[i],
```
slifty: It's not clear to me what these lines are doing -- could we possibly put this in a named function that is more clear about what's going on?

E.g. if this is attempting to coalesce two possible "shapes" of this path I think that warrants a function that explains the need (it also strikes me as a mistake that this could sometimes be an array and sometimes not...)

src/app/file-browser/components/file-list/file-list.component.ts
```diff
@@ -200,6 +207,26 @@ export class FileListComponent
 		}
 	}
 
+	async ngOnChanges() {
+		this.currentFolder =
+			this.dataFolder || this.route.snapshot.data.currentFolder;
+		if (this.dataFolder) {
```
slifty: The variable names here don't make it super easy to understand what is going on; this isn't your fault, but it does mean that we unfortunately might need to rely on comments to help explain what's happening and why.

(If we had infinite time I'd suggest we potentially come up with better names -- what is dataFolder exactly?  Does `currentFolder` mean `renderedFolder`?)

Anyway, let's at least toss in a few comments for future usses

src/app/file-browser/components/file-list/file-list.component.ts
```diff
@@ -561,4 +545,16 @@ export class FileListComponent
 			this.visibleItems.delete(event.component);
 		}
 	}
+
+	private initializeFolder() {
```
slifty: great drying!

One suggestion: this says `initializeFolder` but it seems to be called on both init and changes and it does more than just initialize a single folder.  Is that the right name (maybe it is!).  Is this more of a `syncState` or `

Possibly it could be something like: such as `syncStateWithRoute` which I think captures the intention.

src/app/share-preview/components/share-preview/share-preview.component.ts
```diff
@@ -103,7 +104,7 @@ export class SharePreviewComponent implements OnInit, OnDestroy {
 	public hideBannerSubject: Subject<void> = new Subject<void>();
 	public hideBannerObservable = this.hideBannerSubject.asObservable();
 	public isUnlistedShare = false;
-	public ephemeralFolder: any = null;
+	public ephemeralFolder: FolderVO = null;
```
slifty: I'm surprised this passes our tsc (maybe it infers that this is actually `FolderVO | null` -- but let's make that explicit.

src/app/share-preview/components/share-preview/share-preview.component.html
```diff
@@ -69,10 +69,12 @@
...
 </div>
 <div class="share-preview-content-breadcrumbs">
-	<pr-breadcrumbs></pr-breadcrumbs>
+	<pr-breadcrumbs
+		(onBreadcrumbClicked)="goToFolderFromBreadcrumb($event)"
+	></pr-breadcrumbs>
 </div>
 <div class="share-preview-content" [ngClass]="{ navigating: isNavigating }">
-	@if (true) {
+	@if (ephemeralFolder) {
```
slifty: Since `ephemeralFolder` isn't a boolean can we be more explicit about what we're checking here?  (e.g. `(ephemeralFolder !== null)`)

This will eventually be required by our linter, but for now we may as well follow the practice.

slifty (changes requested):
This is a first pass review -- I don't know *exactly* why some of these changes are being made as a reviewer (which means future devs won't necessarily be able to back-trace why we made these changes / what was diagnoses). All that is to say I would find it really helpful to have more of a narrative in the commit messages explaining what issues were found and why the commit solves the problem.

I flagged a few specific places where I think naming might make the code more self-documenting (which will also help me better understand the intention behind the changes so I can more confidently approve that the implementation aligns with the intent!)

src/app/share-preview/components/share-preview/share-preview.component.html
```diff
@@ -71,15 +71,18 @@
...
-<pr-file-list
-	[dataFolder]="someFolder"
-	(itemClicked)="showFolder($event)"
-></pr-file-list>
 <div class="share-preview-content" [ngClass]="{ navigating: isNavigating }">
-	<router-outlet
-		(activate)="subscribeToItemClicks($event)"
-		(deactivate)="unsubscribeFromItemClicks()"
-	></router-outlet>
+	@if (true) {
```
slifty: Is there any situation where the `else` clause would happen here?

(basically: what's the point of this `if`)

src/app/shared/components/breadcrumbs/breadcrumbs.component.ts
aasandei-vsp: This is indeed to create the path for the router, when the user will click on the breadcrumb. The only change I made was to add a check(made the check better in a new commit), in case those 2 properties are not defined. I would not refactor this, because I don't think it's relevant to this feature.

src/app/share-preview/feature-service.ts
```diff
@@ -0,0 +1,26 @@
+import { Injectable } from '@angular/core';
+
+@Injectable({
+	providedIn: 'root',
+})
+export class FeatureService {
```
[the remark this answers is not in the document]
aasandei-vsp: The service name is indeed very vague. It was a step in my thinking process, but I ended up not using this service. I will re-create the commit history, to make sure things are more clear and this kind of noise is removed.

src/app/file-browser/components/file-list-item/file-list-item.component.html
```diff
@@ -82,7 +86,7 @@
 			>
 				{{ item.displayName }}
 			</p>
-			@if (!showAccess && item.ShareVOs?.length) {
+			@if (!showAccess && item.ShareVOs?.length && !isUnlistedShare) {
```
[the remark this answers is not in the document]
aasandei-vsp: I will be committing a specific message for this, that's a great point. Just FYI, the icon for shared would appear on the thumbnail and it's not relevant for specific files, because we are in a folder that is shared.

src/app/file-browser/components/file-list-item/file-list-item.component.ts
aasandei-vsp: I just used the EventEmitter as it was already used, but the event does not seem to be useful, so I'll just remove it.

PR #836, opened 2025-11-24
[PER-10357] Add fallback for stela folder calls

aasandei-vsp:
Calls for folders and records on stela can be made with two types of authentication:
- auth token
- share token

When making a call with the share token, if that folder/record is not unrestricted, the endpoint will return an empty list. When this happens, if the user is logged in, we try to make the call for the folder/record again, using the auth token, because maybe the user does have access to it and we can still get it.

Issue: https://permanent.atlassian.net/browse/PER-10357

Linked issues:
- https://permanent.atlassian.net/browse/PER-10324
- https://permanent.atlassian.net/browse/PER-10356

HOW TO TEST:

Test for folder fallback: **"My Archive" button**
1. Create an account or use an existing one
2. Create an unlisted share
3. Open a new window and log in to a different account
4. Paste the unlisted share link
5. Click the "My Archive" button
EXPECTATION: Should be redirected to the logged in user's current archive

Test for properties check bug: **"View in my archive" button**
1. Create an account or use an existing one
2. Create a restricted share
3. Open a new window and log in to a different account
4. Paste the restricted share link
5. Accept the share
6. Paste the restricted share link again in the window
7. Click the "View in my archive" button
EXPECTATION: Should be redirected to the logged in user's current archive

Test for record fallback:
I did not find a place to test this, so I just faked an empty response from the code to make sure it works.
image


src/app/shared/services/api/base.ts
```diff
@@ -75,6 +76,7 @@ export class BaseResponse {
 	}
 }
 
+@Injectable()
```
cecilia-donnelly: What does this do?

cecilia-donnelly (changes requested):
I don't think we should make properties optional to make tests pass - it seems to me that the test setup should match reality. It feels much better to me to include the `location` and `files in the `fakeRecordVO` than to adjust the record model to not always have those properties.

One housekeeping note is that the first two commit messages are the same, which confused me a little. I love the overall organization though!

cecilia-donnelly:
> I believe it is better to show an empty field or some inconsistent value than blocking the whole interface.

This is a good point! I worry a bit about unintended/unexpected behavior if we don't have, say, a file. I can imagine what would happen with no thumbnails or location, though.

> the only difference is that one says folder, the other one record

I missed that! :woman_facepalming: Thanks :)

cecilia-donnelly (approved):

omnignorant (approved):
huzzah! this appears to work correctly now based on the test cases provided.

PR #844, opened 2025-11-26
Adjust dependabot cadence

slifty:
This PR updates dependabot to lower the cadence of PRs to monthly and to add a 7-day cooldown period before a given dependency update would be considered.

Since this is a change we want to make across several repositories I did not make an issue to capture the improvement, but we talked about it in Slack!


cecilia-donnelly (approved):
Thank you @slifty !

PR #855, opened 2025-12-09
[PER-10371] Unlisted share record preview fixes

aasandei-vsp:
Jira bug: https://permanent.atlassian.net/browse/PER-10371

STEPS TO TEST

**Record as unlisted share, no account:**
1. Create a new account or log in to existing account
2. Upload a record directly in the archive folder
3. Create an unlisted share link the record
4. Copy link and paste it in a new window
EXPECTED: The preview of the record will appear, with the correct thumbnail
5. Double click on the record thumbnail
EXPECTED: The modal with details will appear, click on the download button/resize buttons

**Record as unlisted share, with account:**
1. Create a new account or log in to existing account
2. Upload a record directly in the archive folder
3. Create an unlisted share link the record
4. Open a new window, create or log in to another account
5. Copy link and paste it in a new window
EXPECTED: The preview of the record will appear, with the correct thumbnail, My Archive button and thumbnail will be present
6. Double click on the record thumbnail
EXPECTED: The modal with details will appear, click on the download button/resize buttons

OBSERVATION: Other scenarios could be tested, just to make sure share works correctly:

- unlisted share folder
- restricted folder/record

### DIFF
--- a/src/app/share-preview/resolves/preview-resolve.service.ts
+++ b/src/app/share-preview/resolves/preview-resolve.service.ts
@@ -91,9 +91,9 @@ export class PreviewResolveService {
 			if (!sharePreviewVO.previewToggle && !showPreview) {
 				const dummy = dummyItems[0];
 				record = cloneDeep(record);
-				record.thumbURL200 = dummy.thumbURL200;
-				record.thumbURL500 = dummy.thumbURL500;
-				record.thumbURL1000 = dummy.thumbURL1000;
+				record.thumbURL200 = record.thumbURL200 || dummy.thumbURL200;
+				record.thumbURL500 = record.thumbURL500 || dummy.thumbURL500;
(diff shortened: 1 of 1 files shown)


cecilia-donnelly (changes requested):
The "My Archive" appears as expected, but clicking it doesn't do anything. That check is outside the test cases in the PR, so potentially it could be a separate piece of work, but I think showing a non-functional button instead of no button is a regression.

I tested with an account that only has one archive.

aasandei-vsp:
> The "My Archive" appears as expected, but clicking it doesn't do anything. That check is outside the test cases in the PR, so potentially it could be a separate piece of work, but I think showing a non-functional button instead of no button is a regression.
> 
> I tested with an account that only has one archive.

Indeed, the button does not work, but the issue is fixed in  the fallback PR that's in testing: https://github.com/PermanentOrg/web-app/pull/836 
Once that is merged, the button should work fine on all types of share.

PR #856, opened 2025-12-09
[PER-10382] Fix fuse options import

aasandei-vsp:
The package fuse.js was updated and it seems the way it exports the IFuseOptions has changed and this broke the build for FE.

Issue: PER-10382 Fix fuse import

This can be tested by searching in the global search.


aasandei-vsp:
Only the local build was affected, so this is not necessary.

PR #860, opened 2025-12-17
Update sourcemaps upload to work with the v3 sentry-cli

liam-lloyd:
The command for uploading sourcemaps has changed for the v3 sentry-cli, this commit updates our CI process to use the new command rather than the no-longer-supported v2 command.


cecilia-donnelly (approved):
Can you add a link to the documentation for the relevant change?

PR #861, opened 2025-12-18
Add support for web archive rendering

### LINKED ISSUE #863: Complete WebRecorder project
We've got a lot of good work done during our 2025 hackathon; the back end changes necessary to support Web Archives are in place, and the majority of the work for support in the web-app is complete.

The following items remain:

1. Ensure that S3 CORS settings allow for Range requests, across all environments.
2. Ensure that CloudFront / CDN settings are correct in order to support Range requests.
3. Host replay on a static page that is loaded via a subdomain across all environments (see [this post](https://lil.law.harvard.edu/blog/2022/09/15/opportunities-and-challenges-of-client-side-playback/) for more).  This might be be `replay.permanent.org` and `replay.dev.permanent.org` etc. and update the embedding to use that version accordingly.

A web archive is attached which, when replayed, will render some diagnostic information about what a given replayed script has access to when it comes to the parent page (e.g. cookies, etc).  We want this replay to show no cookie access and no ability to interact with anything related to whatever permanent.org page is rendering the archive.

Ironically I can't upload the wacz directly so here it is in a zip format first:

[iframe-security-test.wacz.zip](https://github.com/user-attachments/files/24247041/iframe-security-test.wacz.zip)

slifty:
This PR adds support for rendering Web Archives via the ReplayWeb interface, which allows permanent users to upload web archives that then become viewable artifacts.

The feature is currently behind a feature flag until we resolve security considerations (cookie access as well as UX to ensure viewers understand the replayed websites are NOT permanent sites)

Resolves #863

### DIFF
--- a/src/app/file-browser/components/file-viewer/file-viewer.component.spec.ts
+++ b/src/app/file-browser/components/file-viewer/file-viewer.component.spec.ts
@@ -12,6 +12,7 @@ import { TagsService } from '@core/services/tags/tags.service';
 import { PublicProfileService } from '@public/services/public-profile/public-profile.service';
 import { ShareLinksService } from '@root/app/share-links/services/share-links.service';
 import { ApiService } from '@shared/services/api/api.service';
+import { FeatureFlagService } from '@root/app/feature-flag/services/feature-flag.service';
 import { MockComponent } from 'ng-mocks';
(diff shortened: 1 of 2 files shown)


slifty:
This issue is related: https://github.com/PermanentOrg/web-app/issues/863

PR #871, opened 2026-01-07
Remove IDE configuration

slifty:
This PR removes IDE configuration files from version control (more detailed explanation can be found in the commit message)


aasandei-vsp (approved):
Makes sense to me too!

PR #872, opened 2026-01-07
WIP: GitHub actions

slifty:
This PR is not ready for review -- it's an attempt to (1) consolidate our actions and then (2) switch to environments.

Before all is said and done I will almost certainly decouple (2) from this PR and focus just on the action conslidation with our current secret setup.


slifty:
We have issues to track this work; no need to keep this PR open.

PR #874, opened 2026-01-07
PER-10160: Format negative storage amounts

cecilia-donnelly:
In the Transaction History and a few other places, we show bytes formatted into KB, GB, etc. This rounding only worked on positive numbers, which became obvious while reviewing
https://github.com/PermanentOrg/stela/pull/588. Also, add a human-friendly term for the admin adjustment.

Added @liam-lloyd as a reviewer since this was inspired by your change over in stela.

### DIFF
--- a/constants/master_en.json
+++ b/constants/master_en.json
@@ -595,6 +595,7 @@
 		"billing": {
 			"transfer": {
 				"account_to_creditcard": "Refund",
+				"admin_adjustment": "Adjustment applied by Permanent.org",
 				"creditcard_to_account": "Purchase",
 				"pledge_to_account": "Donation",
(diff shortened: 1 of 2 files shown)


omnignorant (approved):
Looks to me that we have the desired result. I have attached before/after screenshots of what I see.

For the description, let's go with "Adjustment applied by Permanent.org". It uses a neutral tone and doesn't specify "who" at Permanent.org is doing it, e.g. Admin which might feel intrusive. I like the word applied here because it feels curative. Just wanted to explain my logic.

Screenshot 2026-01-09 at 4 28 33 PM
Screenshot 2026-01-09 at 4 28 51 PM

Issue #869, opened 2026-01-02
Upgrade to Angular 21


slifty:
There's a new major version of Angular; We've given it some time to give various dependencies time to upgrade, and now it's time for us to see if it's possible

PR #876, opened 2026-01-08
Fix Angular templates to support migration

slifty:
This PR fixes a few of our templates which had invalid Angular commands.

These fixes will make the migration to Angular 21 a bit smoother.

Related to #869

### DIFF
--- a/src/app/file-browser/components/publish/publish.component.html
+++ b/src/app/file-browser/components/publish/publish.component.html
@@ -43,58 +43,60 @@
 				Publish
 			</button>
 		</div>
-		<ng-container *ngIf="publicItem" [ngSwitch]="publishIa">
+		<ng-container *ngIf="publicItem">
 			<hr />
 			<img
 				src="assets/img/internetarchive/IALogo.png"
 				alt="Internet Archive"
 				class="internet-archive-logo"
 			/>
-			<div class="public-link-button" *ngSwitchCase="null">
-				<p>
-					This item can also be published to the Internet Archive.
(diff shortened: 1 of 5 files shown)


src/app/file-browser/components/publish/publish.component.html
```diff
@@ -43,58 +43,60 @@
...
-				</p>
-				<input
-					type="text"
-					class="form-control"
-					readonly
-					[value]="publishIa.permalink"
-					#iaLinkInput
-				/>
-				<div class="buttons">
+			<ng-container [ngSwitch]="publishIa">
```
aasandei-vsp: This ngSwitch should be replaced with @switch or better a simple @if, taking into account that we only have 2 options.

src/app/shared/components/inline-value-edit/inline-value-edit.component.html
```diff
@@ -279,14 +279,14 @@
...
-			<button class="btn btn-primary btn-sm" (click)="save()" type="submit">
-				Save
-			</button>
-			<button class="btn btn-secondary btn-sm" (click)="cancel()" type="button">
-				Cancel
-			</button>
-		</div>
-	</ng-template>
 </ng-container>
+<ng-template #inlineControls>
```
aasandei-vsp: This template makes sense to be outside of the container, but I am not sure it is relevant to the commit at hand.

Also, there seem to be a few ngSwitches here that should be replaced with @switch.

PR #883, opened 2026-01-10
Migrate to `ng-mock`

### LINKED ISSUE #879: Remove `shallow-render` dependency
This is not a standard test tool (only used by 7k weekly downloads), and the project owner no longer works with angular.

We would benefit from moving to more standard testing patterns and removing this dependency.

slifty:
This PR removes our use of `shallow-render` and replaces it with `ng-mock` which is a commonly used mocking library with more community support.

This is an extensive set of changes, so it's broken into smaller parts for easier review; ultimately, however, we will want to squash the commits into a single migration commit.

Resolves #879


src/test.ts
```diff
@@ -36,16 +37,26 @@ window.doNotLoadGoogleMapsAPI = true;
...
+ngMocks.globalReplace(RouterModule, RouterTestingModule);
+
+// Never mock these modules/directives (keep them real)
+ngMocks.globalKeep(NgbDatepickerModule);
+ngMocks.globalKeep(NgbTimepickerModule);
+ngMocks.globalKeep(NgbTooltipModule);
+ngMocks.globalKeep(NgbDropdownModule);
+ngMocks.globalKeep(NgbPaginationModule);
+ngMocks.globalKeep(CountUpDirective);
+ngMocks.globalKeep(CountUpModule);
```
aasandei-vsp: I wonder why we are not mocking these too.

PR #887, opened 2026-01-13
Refactor index files

### LINKED ISSUE #885: Reduce size of `index.html`
Copying the body of the related JIRA ticket https://permanent.atlassian.net/browse/PER-9983:

> Our index.html files are pretty big and hard to manage and keep in sync with each other. This is mainly because of some environment-specific script files embedded into each. Figure out if there’s a way to hide these scripts away so the index.html files are smaller and easier to deal with.

slifty:
This PR improves our `index.html` in two ways:

1. We no longer have environment-specific index files, instead having just one with angular environments determining which dynamic values / tokens to use.

2. We no longer include vendor code in our index file, instead registering it as a dependency which is inserted via angular.

This is much cleaner and easier to maintain.

Resolves #885

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -1,4 +1,4 @@
-import { enableProdMode, ɵresetCompiledComponents } from '@angular/core';
+import { enableProdMode } from '@angular/core';
 import { platformBrowserDynamic } from '@angular/platform-browser-dynamic';
 import { BrowserAgent } from '@newrelic/browser-agent/loaders/browser-agent';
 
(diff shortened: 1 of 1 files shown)


aasandei-vsp:
This is not my area of expertise, but I tried at least to take it locally and see if it works and I get the following error:

- I ran npm in web-app, and ran again docker compose

image

aasandei-vsp:
Some questions:

- How do we test it on the upper env?
- Is there any pipeline change necessary?

aasandei-vsp (approved):
Made it work locally, thanks @slifty for the help!

PR #889, opened 2026-01-15
Update to Angular 21

### LINKED ISSUE #869: Upgrade to Angular 21

slifty:
This PR updates Angular to v21 (the current major release).  It also bumps related dependencies to versions that also support Angular 21.

This is now based on the branch reflected in #896 -- we should smoke test *this* to cover both changes, but this should be merged second.

Resolves #869


aasandei-vsp (commented):
I tried to run this locally and it didn't work for me. I got errors on both npm i and build web-app.
Could you please provide some step-by-step for making it work locally? Maybe I am doing smth wrong?

Also, in the code, there are still some *ngIfs present in src/stories/header.component.ts. Wouldn't this break when running the storybook? I can check it once I can make the branch work on my local.

slifty:
Merging this into the bootstrap upgrade

Issue #127, opened 2022-05-26
Use packages instead of vendored code


jasonaowen:
We have some third-party source code committed in `src/vendor` and `src/style/vendor`, and binary assets in `src/assets` and `src/styles/fonts`. Committing third-party code and assets to our repo makes it more difficult to keep up with upstream updates; makes it harder to properly assign credit; and, worse, allows for the possibility of local modifications - which makes it even harder to pull in upstream changes.

Replace the vendored code and assets with packaged versions.

PR #896, opened 2026-01-16
Move to packaged Bootstrap

### LINKED ISSUE #895: Remove vendored `bootstrap`
This is a sub-task of #127 but It's possible it will reflect its own chunk of work to warrant being pulled into its own issue.

We use angular + bootstrap, but we have bootstrap SCSS included as vendor code instead of as an npm dependency.

Let's fix this, for all the reasons outlined in #127.

slifty:
This PR removes our "vendor" setup for Bootstrap and moves us to using an npm package instead.  This will make it easier to avoid version drift and also allows us to remove third party code from our repository.

This also updates our version of Bootstrap to v5, which is the latest version and is compatible with our other angular dependencies.

🚨 🚨 This adds a new dependency, so you'll need to run docker compose build web-app 🚨 🚨 

Resolves #895
Related to #127


aasandei-vsp (changes requested):
It seems we might be needing to figure out multiple dependencies, not only the bootstrap one.

aasandei-vsp:
Also, the merge button is green and I do not think that is correct, right?

PR #902, opened 2026-01-17
Fix the storybook build

### LINKED ISSUE #901: Fix Storybook build
When migrating to `ng-mocks`, we broke the Storybook build. That is now blocking deployment of this repo. It looks like we no longer have `jasmine` and need to replace it in Storybook files.

slifty:
This PR fixes an issue where our test files were being included in the storybook build.  This wasn't causing issues before, but some new tooling caused an issue that raised the need to ignore it.

Resolves #901

### DIFF
--- a/.storybook/tsconfig.json
+++ b/.storybook/tsconfig.json
@@ -5,7 +5,7 @@
 		"allowSyntheticDefaultImports": true,
 		"resolveJsonModule": true
 	},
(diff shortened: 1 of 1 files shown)


slifty:
I ran dev build + deploy on this branch and it has fixed the storybook build / succeeded: https://github.com/PermanentOrg/web-app/actions/runs/21087412745/job/60653208915

PR #909, opened 2026-01-23
[PER-10322] Fix choose archive flow

aasandei-vsp:
Testing notes:

- Need a test account with multiple archives

- Check both restricted links and unrestricted

This only applies to restricted shares. We will address the multiple archive choice for unlisted shares when we implement the functionality for adding them to the user's archive.

Issue: https://permanent.atlassian.net/browse/PER-10322

### DIFF
--- a/src/app/shared/components/archive-switcher/archive-switcher.component.html
+++ b/src/app/shared/components/archive-switcher/archive-switcher.component.html
@@ -1,6 +1,6 @@
 <div class="archive-list-wrapper">
 	<div class="archive-list">
-		<div class="archive-list-title">Select archive to switch to:</div>
+		<div class="archive-list-title">Select archive to request access:</div>
 		@for (archive of archives; track archive) {
 			<pr-archive-small
 				[archive]="archive"
@@ -21,5 +21,7 @@
 	<button class="btn btn-primary" (click)="createArchiveClick()">
 		Create new archive
 	</button>
-	<button class="btn btn-primary" (click)="backButtonClick()">Back</button>
+	<button class="btn btn-primary" (click)="backButtonClick()">
+		Current Archive
+	</button>
 </div>


cecilia-donnelly (approved):
Testing notes:
- Need a test account with multiple archives
- Check both restricted links and unrestricted

It appears that this only applies to restricted links. I didn't see a way to add an unrestricted shared item (single record) to my archive, so there wasn't a "switch archives" flow either. Is that expected?

I think this is an error, but maybe not:
The "back" button closed the modal (correct) but also accepted the share and took me into my workspace. That was unexpected for me because I thought it would behave more like "cancel," but it's really a product question because I had at that point already clicked "Accept Share," so it might be fine.

Overall the code makes sense and this is a definite improvement, thank you @aasandei-vsp!

I'll approve and ask @omnignorant to address the question about the "back" button.

aasandei-vsp:
> Testing notes:
> 
> * Need a test account with multiple archives
> * Check both restricted links and unrestricted
> 
> It appears that this only applies to restricted links. I didn't see a way to add an unrestricted shared item (single record) to my archive, so there wasn't a "switch archives" flow either. Is that expected?
> 
> I think this is an error, but maybe not: The "back" button closed the modal (correct) but also accepted the share and took me into my workspace. That was unexpected for me because I thought it would behave more like "cancel," but it's really a product question because I had at that point already clicked "Accept Share," so it might be fine.
> 
> Overall the code makes sense and this is a definite improvement, thank you @aasandei-vsp!
> 
> I'll approve and ask @omnignorant to address the question about the "back" button.

Thank you @cecilia-donnelly These are great points, I copied the testing notes to the PRs description.

Related to the unlisted share, the story I used for implementation mentioned that we will not be adding them to the user's archive. It seems though that mobile does that, so we will add a story to do it on web as well.

The back button functionality is a bit weird, indeed, but as you mentioned, we do need a business decision and some design for that. I do have a story for improvements, so I'll add this as well: https://permanent.atlassian.net/browse/PER-10385

PR #924, opened 2026-02-19
[PER-10416] Add date and time selection modal

aasandei-vsp:
**!!!!!!!!!! !!!!! This PR will be broken in multiple ones, so there is no need for it to remain active. The branch and the code are still in use for creating the rest of the PRs for the EDTF date.**

Bellow is information about what has been developed on this task and how to test it. This summary and test cases have been generated using Claude Code.

============================================================
EDTF DATE EDITING - COVERAGE & MANUAL TEST PLAN (PER-10590)

This document captures three things:

What part of the EDTF standard the new UI lets users produce -
the things QA should check work end-to-end.

What it doesn't (and why) - so we don't end up filing
intentional limitations as bugs.

A structured manual-test checklist QA can walk through.

============================================================
PART 1 - WHAT WE SUPPORT, AND WHAT WE DON'T

The EDTF standard is layered. Level 0 covers everyday calendar
dates. Level 1 adds qualifiers ("approximately 1985") and intervals
(date ranges with open ends). Level 2 adds more exotic features
like seasons, sets of dates, and BCE years. The new editor is
focused on the parts our archivists actually use.

For every construct below, we mark it as one of:

Supported       The user can produce this from the UI today,
                and it round-trips back when they reopen the
                record.

Implicit only   The underlying parser can read it (so old data
                is preserved correctly), but the UI doesn't
                give users a way to create it from scratch. If
                we want users to be able to enter these values,
                we'll need follow-up tickets.

Not supported   Neither the UI nor the underlying service
                handles this construct at all.

Level 0 - everyday calendar dates

Just a year (e.g. 1985)
  Status: Supported

Year and month (e.g. 1985-04)
  Status: Supported

Full date (e.g. 1985-04-12)
  Status: Supported

Date with UTC time (e.g. 1985-04-12T23:20:30Z)
  Status: Supported

Date with offset time (e.g. 1985-04-12T23:20:30-04:00)
  Status: Supported

Closed date range (e.g. 1964/2008)
  Status: Supported

Level 1 - qualifiers and partial dates

Approximate date, "around" (e.g. 1984~)
  Status: Supported

Uncertain date, "we think but we're not sure" (e.g. 1984?)
  Status: Supported

Both at once, "around and unsure" (e.g. 1984%)
  Status: Supported

Year with unknown digits, e.g. some time in the 20th century (19XX)
  Status: Implicit only

Month with unknown value (e.g. 1985-XX)
  Status: Implicit only

Day with unknown value (e.g. 1985-04-XX)
  Status: Implicit only

Whole date is unknown (XXXX-XX-XX)
  Status: Supported

Range with open start (e.g. ../1985-04-12)
  Status: Supported

Range with open end (e.g. 1985-04-12/..)
  Status: Supported

Range bound marked as uncertain instead of open (e.g. ?/1985)
  Status: Not supported

Season - spring, summer, autumn, winter (e.g. 1985-22)
  Status: Not supported

Level 2 - advanced features we intentionally don't handle

Sub-year groupings - quarters, semesters, etc. (e.g. 1985-26)
  Status: Not supported

"One of these dates" (e.g. [1667,1668,1670..1672])
  Status: Not supported

"All of these dates" (e.g. {1667,1668})
  Status: Not supported

Per-component qualifiers, e.g. only the day is uncertain
(e.g. ?2004-06-~01)
  Status: Not supported - a qualifier always applies to the
          whole date

BCE / negative years (e.g. -0533)
  Status: Not supported

Long-form years, more than 4 digits (e.g. Y170000002)
  Status: Not supported

Things to know about time and timezones

The time fields use a 12-hour clock with AM/PM. Internally we
convert this to 24-hour format before saving, and convert back
when reading.

You can only attach a time to a complete date (year + month +
day). If you have only a year, or year + month, the time is
silently dropped on save. This is intentional - partial dates
with times don't really make sense - but it can look like a
bug if you don't know.

The timezone is picked from a fixed list of 32 GMT offsets
(e.g. "Eastern Standard Time / GMT-05:00"). We don't model
daylight savings, region-specific zones (like
"America/New_York"), or arbitrary offsets.

If you fill in seconds but leave minutes blank, the time is
rejected as invalid.

If you fill in minutes but leave seconds blank, seconds
default to 00 on save.

Validation messages users will see

Both Approximate and Uncertain are on, but the underlying
date is invalid:
  "The date is not valid as both approximate and uncertain.
  Please check the date values."

Approximate is on, the underlying date is invalid:
  "The date is not valid as approximate. Please check the
  date values."

Uncertain is on, the underlying date is invalid:
  "The date is not valid as uncertain. Please check the
  date values."

A date range with the end before the start, or otherwise
unparseable as an interval:
  "The date range is not valid. Please make sure the start
  date is before the end date."

Anything else the parser doesn't accept:
  "The date entered is not valid. Please check the values
  and try again."

============================================================
PART 2 - MANUAL TEST CASES

Section A - Entering single dates (Sidebar)

  A1. Enter year 1985 only and save.
        Saved EDTF: 1985
        Sidebar:    "1985"

  A2. Year 1985, month 04, then save.
        Saved EDTF: 1985-04
        Sidebar:    "April 1985"

  A3. Year 1985, month 04, day 12, then save.
        Saved EDTF: 1985-04-12
        Sidebar:    "April 12, 1985"

  A4. Continue from A3, add 02:30:00 AM with timezone "Eastern
      Standard Time", then save.
        Saved EDTF: 1985-04-12T02:30:00-05:00
        Sidebar:    "02:30:00 AM EST"

  A5. Continue from A3, add 12:00 AM (midnight), then save.
        Saved EDTF: ends in T00:00:00
        Sidebar:    "12:00:00 AM" preserved on reload

  A6. Continue from A3, add 12:00 PM (noon), then save.
        Saved EDTF: ends in T12:00:00
        Sidebar:    "12:00:00 PM" preserved on reload

  A7. Continue from A3, add 01:15 PM with timezone "Greenwich
      Mean Time", then save.
        Saved EDTF: 1985-04-12T13:15:00+00:00
        Sidebar:    "01:15:00 PM GMT"

  A8. After saving any date above, reopen the dropdown and click
      the Clear date and time (trash) button, then save.
        Saved EDTF: empty string
        Sidebar:    "Click to add date"

Section B - Date ranges (Modal)

  B1. Open the modal, toggle "Use a date range", enter 1985 then
      2000, save.
        Saved EDTF: 1985/2000

  B2. Range 1985-04-12 to 1985-04-20, save.
        Saved EDTF: 1985-04-12/1985-04-20

  B3. Same start and end date, but start time 09:00 AM EST and
      end time 05:00 PM EST, save.
        Saved EDTF:
          1985-04-12T09:00:00-05:00/1985-04-12T17:00:00-05:00

  B4. Open-end: enter only the start date 1985-04-12, leave the
      end blank, save.
        Saved EDTF: 1985-04-12/..

  B5. Open-start: leave the start blank, enter only the end date
      2000-01-01, save.
        Saved EDTF: ../2000-01-01

  B6. Range toggle is on but both start and end are blank, save.
        Saved EDTF: empty value (no date saved)

Section C - Qualifiers (Modal)

  C1. Year 1985, toggle Approximate on, save.
        Saved EDTF: 1985~
        UI:         sidebar shows a blue dot next to the date and
                    the word "Approximate"

  C2. Date 1985-04-12, toggle Uncertain on, save.
        Saved EDTF: 1985-04-12?
        UI:         sidebar shows the word "Uncertain"

  C3. Date 1985-04-12, toggle both Approximate and Uncertain on,
      save.
        Saved EDTF: 1985-04-12%
        UI:         sidebar shows "Approximate, Uncertain"

  C4. Toggle Unknown on, save.
        Saved EDTF: XXXX-XX-XX
        UI:         all date and time fields go grey and become
                    uneditable; sidebar shows "Unknown"

  C5. Enter a full date and time, toggle Unknown on, then
      immediately toggle it off again.
        Expected:   the original date and time should come back
                    unchanged. Verifies that turning Unknown on
                    doesn't permanently destroy the user's
                    previous input.

  C6. With Unknown on, try to toggle Approximate or Uncertain.
        Expected:   those toggles should be disabled and
                    unclickable.

  C7. Date 1985-04-12T02:30:00-05:00, toggle Approximate on, save.
        Saved EDTF: 1985-04-12T02:30:00-05:00~
        UI:         the qualifier marker (~) appears AFTER the
                    timezone, not in the middle of the string.

Section D - Reading existing values

  D1. Stored: 1985
        Modal: year 1985; month, day, time and qualifiers all empty

  D2. Stored: 1985-04-12T23:20:30Z
        Modal: date 1985-04-12, time 11:20:30 PM, timezone
               "Greenwich Mean Time"

  D3. Stored: 1985-04-12T23:20:30+05:30
        Modal: date populated, time populated, timezone
               "India Standard Time"

  D4. Stored: 1985~
        Modal: year 1985, Approximate toggle on

  D5. Stored: 1985%
        Modal: year 1985, Approximate on, Uncertain on

  D6. Stored: XXXX-XX-XX
        Modal: Unknown toggle on, all other fields blank
               and disabled

  D7. Stored: 1985/2000
        Modal: range toggle on, start year 1985, end year 2000

  D8. Stored: 1985-04-12/..
        Modal: range toggle on, start date filled in, end date
               blank or showing ".."

  D9. Stored: ../2000-01-01
        Modal: range toggle on, start blank or showing "..",
               end date filled in

Section E - Field-level validation while typing

  E1.  Year, type "abcd".
         Expected: letters are rejected; the field stays empty.

  E2.  Year, type "0985".
         Expected: the leading zero is rejected - years must not
                   start with 0.

  E3.  Year, type "19" then click away.
         Expected: accepted as in-progress; nothing is saved
                   until 4 digits are entered.

  E4.  Year, type "1985".
         Expected: after the 4th digit, focus jumps automatically
                   to the month field.

  E5.  Month, type "13".
         Expected: the second digit is rejected; only 01-12 are
                   allowed.

  E6.  Month, type "2" as the first digit.
         Expected: rejected; the first digit of a month must be
                   0 or 1.

  E7.  Month, type "02".
         Expected: focus jumps automatically to the day field.

  E8.  Day, in February of a leap year (e.g. 2000), type "29".
         Expected: accepted.

  E9.  Day, in February of a non-leap year (e.g. 1985), type "29".
         Expected: rejected - that year's February only has 28
                   days.

  E10. Day, in April, type "31".
         Expected: rejected - April only has 30 days.

  E11. Day, type "00".
         Expected: rejected.

  E12. Hour, type "13".
         Expected: rejected - the input is a 12-hour clock.

  E13. Hour, type "00".
         Expected: the single digit 0 is allowed in-progress, but
                   00 as a complete value is rejected.

  E14. Minutes, type "60".
         Expected: rejected.

  E15. Minutes, type "9" as the first digit.
         Expected: rejected - the first digit of minutes must be
                   0-5.

  E16. Seconds, enter seconds while the minutes field is still
       blank.
         Expected: the Save button stays disabled and the modal
                   shows the generic "date entered is not valid"
                   error.

Section F - Whole-date validation in the modal

  F1. Range with start year 2000 and end year 1985 (end before
      start).
        Error:  "The date range is not valid. Please make sure
                the start date is before the end date."
        Save:   disabled

  F2. Date 2024-02-30 (a day that doesn't exist).
        Error:  "The date entered is not valid. Please check
                the values and try again."
        Save:   disabled

  F3. Approximate on with otherwise-invalid data.
        Error:  "The date is not valid as approximate. Please
                check the date values."
        Save:   disabled

  F4. Uncertain on with otherwise-invalid data.
        Error:  "The date is not valid as uncertain. Please
                check the date values."
        Save:   disabled

  F5. Both qualifiers on with otherwise-invalid data.
        Error:  "The date is not valid as both approximate and
                uncertain. Please check the date values."
        Save:   disabled

  F6. Year only (1985) and try to add a time.
        Error:  none, but the time is silently dropped on save
                (intentional - time needs a full year/month/day).
        Save:   enabled

  F7. Open-start range: blank start, end 2000-01-01.
        Error:  none.
        Save:   enabled - saves as ../2000-01-01

Section G - Timezone dropdown

  G1. Open the dropdown, type "Eastern".
        Expected: the list filters to "Eastern Standard Time"
                  and "Eastern European Standard Time".

  G2. Type "+05:30".
        Expected: the list filters to "India Standard Time".

  G3. Type "xyzzy".
        Expected: the list is empty (apart from the "Select
                  timezone" reset row at the top).

  G4. Pick "India Standard Time", save, reopen the record.
        Expected: sidebar shows the abbreviation "IST"; modal
                  preselects "India Standard Time".

  G5. Pick "International Date Line West" (GMT-12:00), save.
        Expected: sidebar shows the abbreviation "IDLW".

  G6. After picking a timezone, reopen the dropdown and click the
      "Select timezone" row at the top.
        Expected: the timezone clears; the saved EDTF no longer
                  has an offset suffix.

  G7. Range toggle on; set the start timezone to "Eastern Standard
      Time" and the end timezone to "Japan Standard Time", save.
        Expected: each side of the saved range carries its own
                  offset - they don't have to match.

Section H - UI integration and regressions

  H1. Open the sidebar on a record where the user only has
      VIEWER permission.
        Expected: the date row shows the current value (or
                  "No date") with no edit icon, and nothing
                  happens when clicked.

  H2. Open the sidebar's date dropdown, then click More options.
        Expected: the dropdown closes and the full modal opens,
                  pre-populated with whatever was in the dropdown.

  H3. Inside the modal, change something, then click Cancel.
        Expected: the modal closes; the sidebar still shows the
                  original (pre-edit) value.

  H4. Inside the modal, enter valid data, then click Save.
        Expected: the modal closes; the sidebar reflects the new
                  value; reloading the page keeps it.

  H5. Click the date row of a record that already has qualifiers,
      a range, or "unknown" set.
        Expected: the modal opens directly - the lightweight
                  inline dropdown is skipped because the existing
                  data is too complex for it.

  H6. Inside the modal, click Clear date and time (the trash
      icon), then Save.
        Expected: the date is wiped; the sidebar shows the empty
                  state and reloading confirms the value is
                  actually gone.

  H7. Save the modal while disconnected from the network.
        Expected: a toast error appears (no silent failure).


src/app/shared/components/datepicker-input/datepicker-input.component.ts
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	private getMaxDaysInMonth(year: string, month: string): number {
+		const y = parseInt(year, 10);
+		const m = parseInt(month, 10);
+
+		if (!m || m < 1 || m > 12) return 31;
+
+		if (m === 2 && y) {
+			return this.isLeapYear(y) ? 29 : 28;
```
cecilia-donnelly: I see you handle February here, but it does seem strange to have to write this ourselves.

cecilia-donnelly (commented):
I'm sorry, I forgot to finish and submit this!

cecilia-donnelly (commented):
Can you base this on `PER-10475-add-patch-folder`, so we're not reviewing the same commits in two different places? Once that one is merged this one should automatically switch to being based on main.

src/app/file-browser/components/edit-date-time/edit-date-time.model.ts
```diff
@@ -0,0 +1,78 @@
...
+	{ offset: 'GMT+06:00', name: 'Bangladesh Standard Time' },
+	{ offset: 'GMT+07:00', name: 'Indochina Time' },
+	{ offset: 'GMT+08:00', name: 'China Standard Time' },
+	{ offset: 'GMT+09:00', name: 'Japan Standard Time' },
+	{ offset: 'GMT+09:30', name: 'Australian Central Standard Time' },
+	{ offset: 'GMT+10:00', name: 'Australian Eastern Standard Time' },
+	{ offset: 'GMT+11:00', name: 'Solomon Islands Time' },
+	{ offset: 'GMT+12:00', name: 'New Zealand Standard Time' },
+	{ offset: 'GMT+13:00', name: 'Tonga Standard Time' },
+];
```
cecilia-donnelly: Ideally we would not have a list of timezones hardcoded here, though I understand why you did it this way. Where did this list come from?

src/app/file-browser/components/edit-date-time-modal/edit-date-time-mapping.service.ts
```diff
@@ -0,0 +1,265 @@
...
+		if (hour24 === 12) {
+			return { hours: '12', amPm: Meridian.PM };
+		}
+		return { hours: String(hour24 - 12).padStart(2, '0'), amPm: Meridian.PM };
+	}
+
+	static to24Hour(hour12: number, amPm: Meridian): number {
+		if (amPm === Meridian.AM) return hour12 === 12 ? 0 : hour12;
+		return hour12 === 12 ? 12 : hour12 + 12;
+	}
```
cecilia-donnelly: I'm going commit by commit and I think I've seen another `to24Hour` along the way, but maybe it moved here?

src/app/file-browser/components/edit-date-time-modal/edit-date-time-mapping.service.ts
```diff
@@ -0,0 +1,265 @@
...
+		'GMT+06:00': 'BST',
+		'GMT+07:00': 'ICT',
+		'GMT+08:00': 'HKT',
+		'GMT+09:00': 'JST',
+		'GMT+09:30': 'ACST',
+		'GMT+10:00': 'AEST',
+		'GMT+11:00': 'SBT',
+		'GMT+12:00': 'NZST',
+		'GMT+13:00': 'TOT',
+	};
```
cecilia-donnelly: Same comment about not keeping our own list of timezones or, here, abbreviations.

cecilia-donnelly (commented):
This is a lot! It looks fantastic. I am definitely wondering about doing less of our own date building (as I was asking in my first encounter with this PR a while ago), but I love how you're using the EDTF library to validate.

src/app/shared/services/edtf-service/edtf.service.ts
```diff
@@ -1,14 +1,112 @@
...
+	{ offset: 'GMT+06:00', name: 'Bangladesh Standard Time' },
+	{ offset: 'GMT+07:00', name: 'Indochina Time' },
+	{ offset: 'GMT+08:00', name: 'China Standard Time' },
+	{ offset: 'GMT+09:00', name: 'Japan Standard Time' },
+	{ offset: 'GMT+09:30', name: 'Australian Central Standard Time' },
+	{ offset: 'GMT+10:00', name: 'Australian Eastern Standard Time' },
+	{ offset: 'GMT+11:00', name: 'Solomon Islands Time' },
+	{ offset: 'GMT+12:00', name: 'New Zealand Standard Time' },
+	{ offset: 'GMT+13:00', name: 'Tonga Standard Time' },
+];
```
cecilia-donnelly: I think it's important not to manage our own list of timezones. It looks like Luxon might be fairly commonly used nowadays, but we also already have `moment` which has the affiliated `moment-timezone`. I won't write too much but it looks like there are good reasons to move away from `moment` and to a newer library BUT that doesn't need to be part of this work unless you decide it makes sense. I assume you know much more about this! I will add some other similar comments elsewhere.

In any case, I do not want to commit a timezone list of our own. That is a problem we already have in the backend and we can avoid doing the same thing in the front end.

src/app/shared/services/edtf-service/edtf.service.ts
```diff
@@ -42,45 +140,130 @@ export interface DateTimeModel {
...
+
+		const match = offset.match(/GMT([+-])(\d{2}):(\d{2})/);
+		if (!match) return offset;
+		const sign = match[1];
+		const hrs = parseInt(match[2], 10);
+		const mins = parseInt(match[3], 10);
+		return mins === 0
+			? `UTC${sign}${hrs}`
+			: `UTC${sign}${hrs}:${String(mins).padStart(2, '0')}`;
+	}
```
cecilia-donnelly: I assume a library like `moment-timezone` would also handle the offset parsing though I may be wrong.

src/app/shared/services/edtf-service/edtf.service.ts
```diff
@@ -1,14 +1,112 @@
...
+
 export enum EdtfPrecision {
 	Time = 0,
 	Year = 1,
 	Month = 2,
 	Day = 3,
 }
 
-export interface DateQualifier {
+export const DAYS_IN_MONTH = [31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31];
```
cecilia-donnelly: I'd also rather not handle this ourselves -- just seems unnecessary given the complexity of leap years. https://momentjs.com/docs/#/displaying/days-in-month/

src/app/shared/services/edtf-service/edtf.service.ts
```diff
@@ -264,39 +471,188 @@ export class EdtfService {
...
+		if (value.length === 1) return num >= 0 && num <= 1;
+		return num >= 1 && num <= 12;
+	}
+
+	isValidMinutesSeconds(value: string): boolean {
+		if (!this.isNumeric(value) || value.length > 2) return false;
+		const num = parseInt(value, 10);
+		if (value.length === 1) return num >= 0 && num <= 5;
+		return num >= 0 && num <= 59;
+	}
```
cecilia-donnelly, line 453: Similar to timezones, I don't think we need to or should write all these validators ourselves.

https://momentjscom.readthedocs.io/en/latest/moment/05-query/09-is-leap-year/ for specifically that one

https://momentjscom.readthedocs.io/en/latest/moment/01-parsing/15-is-valid/ handles a bunch of these validation cases and then returns an object that itself has to be parsed so we can respond with a meaningful error to the user.

src/app/shared/components/datepicker-input/datepicker-input.component.ts
```diff
@@ -0,0 +1,145 @@
...
+		const input = event.target as HTMLInputElement;
+		const value = input.value;
+
+		if (
+			value !== '' &&
+			(!this.edtfService.isNumeric(value) || value.length > 4)
+		) {
+			input.value = this.date.year;
+			return;
+		}
```
cecilia-donnelly, line 86: I wonder if we can also use a library for some of this validation in the component.

Also for years specifically, EDTF does handle future years that are more than 4 digits, so that should be allowed. Note also that future years over 4 digits will not be numeric: "Letter-prefixed calendar year" in https://www.loc.gov/standards/datetime/.

cecilia-donnelly: (I know, talk about complication...)

cecilia-donnelly (changes requested):
This is such a major undertaking! I think it would be worth writing a blog post about it afterward, so perhaps someone else could benefit from your research and work.

At this point my overall feedback is to rely more on `moment` (since it's already in the project) or to choose another datetime library. The most urgent requirement for a library is timezone management, in my opinion, but it could also simplify the other code you've written for parsing dates. Obviously it won't work for everything, since EDTF dates aren't necessarily "real" dates from the perspective of a date library. It does seem like it could help with am/pm parsing, though, for example, and much of the validation you've written. Specific comments inline!

src/app/shared/components/datepicker-input/datepicker-input.component.ts
cecilia-donnelly: By the way, this is in Level 1. I'm now seeing in the test notes in this PR that this is marked as "Not supported" but they should be (probably just a Claude-ism, but figured I'd mention it explicitly).

src/app/shared/services/edtf-service/edtf.service.ts
```diff
@@ -198,11 +404,10 @@ export class EdtfService {
...
 
-		if (offset === 'Z') {
-			return { timezoneOffset: '+00:00', timezoneName: 'UTC' };
-		}
-
-		return { timezoneOffset: offset, timezoneName: '' };
+		// Normalise to GMT+hh:mm format so it matches the TIMEZONES list
+		const gmtOffset = offset === 'Z' ? 'GMT+00:00' : `GMT${offset}`;
+		const tz = TIMEZONES.find((t) => t.offset === gmtOffset);
+		return { timezoneOffset: gmtOffset, timezoneName: tz?.name || '' };
```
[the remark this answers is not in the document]
aasandei-vsp: The edtf library used here rejects any qualifier after a time-level EDTF string at the parse step — meaning:

- toEdtfDate (which produces these strings) calls edtf(stringDate) on line 262 to validate; it would throw before the string is ever returned.

- toDateTimeModel can't parse them either (the failing tests above hit the library's "Syntax error at line 1 col 21").

So strings like 1985-04-12T02:30:00-05:00~ can't be produced by the app today, and the regex limitation in extractTimezone is unreachable.

src/app/shared/services/edtf-service/edtf.service.ts
```diff
@@ -116,43 +307,51 @@ export class EdtfService {
...
 	): string {
 		const dateStr = this.buildDateString(date);
-		const timeStr = this.buildTimeString(time);
-		const utcSuffix = timeStr ? 'Z' : '';
+		const edtfObject = edtf(dateStr);
+		// Strip any time/timezone the library may append (e.g. T00:00:00.000Z)
+		let result = edtfObject.toEDTF().replace(/T.*$/, '');
 
+		const hasCompleteDate = !!(date.year && date.month && date.day);
+		const timeStr = hasCompleteDate ? this.buildTimeString(time) : '';
```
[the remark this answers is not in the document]
aasandei-vsp: Thanks — I traced this through and the current code already handles the blank-date case the way the spec wants, without throwing.

When date.year is empty, buildDateString pads it to 'XXXX' via ''.padEnd(4, 'X'), which is valid EDTF for "unspecified 4-digit year" and parses successfully (edtf('XXXX') returns an ExtDate). isEmptyDateTime only returns true when the time is also blank, so a time-with-blank-date input does reach normalizeEdtfString — but hasCompleteDate is false, so timeStr is set to '' and the time is silently dropped, exactly as the UI spec calls for. The function returns 'XXXX' and edtf(stringDate) on the round-trip validation also succeeds.

The only path that would throw inside normalizeEdtfString is a non-numeric year (e.g. 'abc' → 'abcX'), and that's already prevented upstream by isValidYear.

src/app/shared/services/edtf-service/edtf.service.ts
```diff
@@ -264,39 +471,188 @@ export class EdtfService {
 		return !date.year && !date.month && !date.day && !time?.hours;
 	}
```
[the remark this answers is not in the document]
aasandei-vsp: Same trace as the previous comment on normalizeEdtfString — this doesn't actually break. When date.year is empty, buildDateString pads to 'XXXX' (via ''.padEnd(4, 'X')), which is valid EDTF and parses successfully. Inside normalizeEdtfString, hasCompleteDate is false, so timeStr is set to '' and the time is silently dropped — which is exactly the behaviour you're suggesting. The function returns 'XXXX' and the final edtf(stringDate) validation also succeeds. No throw, no blocked save.

So isEmptyDateTime's purpose is narrower than the name suggests: it gates whether the start/end slot collapses to .. (open interval) vs. a real EDTF value. Treating "year-less + time" as fully empty would change interval semantics for the end-date case, which we don't want. The time-dropping behaviour you're after is already handled one layer down.

src/app/shared/services/edtf-service/edtf.service.ts
aasandei-vsp: TL;DR It can be done, but it's a bit more complicated than just using a library.

Summary: Timezone handling for EDTF date editing
The EDTF format constraint
EDTF only supports numeric UTC offsets (like -05:00 or Z) in its date strings. It has no way to carry a named timezone like America/New_York. This is a property of the spec, not something we can work around in our code.

The practical consequence: an EDTF value on its own is ambiguous about where the date was recorded. A -05:00 offset could be New York in winter, Bogotá, Lima, or several other places. If we want to preserve the user's actual zone choice, it has to be stored separately from the EDTF string.

Frontend changes
Stop maintaining our own timezone list. Today we ship a hardcoded list of 32 GMT offsets and a static abbreviation map inside EdtfService. Both should go — the timezone data should come from a library or from the browser itself.

Serialization stays correct across DST. When a user picks a named zone and a date, we ask the source of truth for the exact offset that applies on that specific date, and that offset is what we write into the EDTF string. So a New York event in January serializes with -05:00, and one in July with -04:00, automatically.

One UX call to make. Today's dropdown is 32 friendly GMT-offset rows. The full IANA list is roughly 430 entries. Two reasonable directions:

Show the full (or curated) IANA list and lean on the existing search filter to make it scannable — minimal component change
Group entries by region in the dropdown for easier browsing
Leaning toward option 1, since the search filter already does most of the work.

Where the timezone data should come from
A few realistic options, in order of how closely each matches "give me a nice timezone picker without owning anything":

@vvo/tzdb — the strongest match. A curated, deduplicated list of IANA timezones packaged as pure data (no date math, no parser). Each entry includes raw offset, current offset, standard and daylight abbreviations, the cities sharing the zone, and the country. It collapses the ~430 raw IANA zones into ~140 "main" ones, so the dropdown is shorter and friendlier out of the box. This is what most polished timezone pickers use.
timezones-list — much lighter, just label/value pairs ready to drop into a dropdown. Less metadata; useful if all we need is a list of strings.
countries-and-timezones — organized by country. Useful if we ever want a country-first picker.
Native browser APIs — modern browsers ship the IANA list via Intl.supportedValuesOf('timeZone'), and can derive DST-correct labels and abbreviations on demand. Zero dependencies, but gives us the raw ~430-entry list with no curation; we'd still want to filter or group it ourselves.
Recommendation: @vvo/tzdb if we want a polished picker without writing curation logic, native Intl if we're fine with a longer raw list and a search filter doing the work and want to add zero dependencies. Either path lets us delete the hardcoded list and abbreviation map.

Backend changes
If we want the user's named-zone choice to survive a save-and-reload, the backend needs to store it as a separate field alongside the EDTF value. The EDTF column stays spec-pure (offset only); the named zone lives next to it as metadata.

Without that field, we can still save and load the date correctly, but the dropdown will only be able to show the raw offset on re-display — the original "New York" choice would be lost. Worth checking how the backend handles zone information today before committing to a schema change.

Out of scope
Migrating off moment.js to a newer date library is a worthwhile but separate effort. The timezone work above doesn't depend on it — neither @vvo/tzdb nor native Intl is a date library, so this change is dependency-light regardless of which path we pick.

src/app/shared/services/edtf-service/edtf.service.ts
aasandei-vsp: Comment above relates to this, we'll have a clear answer only after we decide what direction we want to take with the timezone.

src/app/shared/components/datepicker-input/datepicker-input.component.ts
aasandei-vsp: I didn't support it, I have some validation in place specifically on the year input, but that's wrong and can be fixed easily, by just adapting that validation :D

Thank you for noticing that!

src/app/shared/services/edtf-service/edtf.service.ts
aasandei-vsp: Agreed, we have another library already `date-fns` that can do date formatting and validation, so I'll use it here, as it is actively maintained.

It is confusing, because we also use `moment` in some places and in others we use `date-fns`.

src/app/shared/services/edtf-service/edtf.service.ts
aasandei-vsp: I'll use `date-fns` for this as well. Thank you, that's a great catch!

aasandei-vsp:
I did go through all the comments, but changes are still in progress. I will let you guys know when it can be reviewed again!

src/app/shared/components/datepicker-input/datepicker-input.component.ts
aasandei-vsp: After the discussion we had, I understand we will not be supporting years that have more than 4 digits or years that get a letter prefix. Please let me know if there's something I'm missing.

As per validation, I will try and use the `date-fns` library wherever I can, so we do not reinvent the wheel of validation :)))

src/app/shared/services/edtf-service/edtf.service.ts
cecilia-donnelly: Thank you!

src/app/shared/services/edtf-service/edtf.service.ts
cecilia-donnelly: We're waiting on a bit of design help for this.

src/app/shared/components/datepicker-input/datepicker-input.component.ts
cecilia-donnelly: Agreed! Not supporting years with more than 4 digits or years with a letter prefix or negative years.

cecilia-donnelly:
https://docs.google.com/document/d/1jXnHCVYgiR7VpjwyY9kxi4F64MM1OR8D_-K1_5ICiF8/edit?tab=t.0#heading=h.65t919juh8gh has the current discussion.

aasandei-vsp:
This work has already been committed in other PRs.

PR #958, opened 2026-03-06
Per 9589 milestone sort

cecilia-donnelly:
Allow users to display milestones on the public profile either chronologically or reverse chronologically.

~I could use some help with making it obvious to the user (1) that this is a button and (2) what the button does. I don't know how to make sure that they understand that they are reordering the milestones on a different screen than the one they are looking at.~

I have changed this in response to feedback to let users change the order in archive settings. I also point to that setting from the profile editor where members create milestones.

### DIFF
--- a/src/app/core/components/profile-edit/profile-edit.component.html
+++ b/src/app/core/components/profile-edit/profile-edit.component.html
@@ -455,7 +455,13 @@
 						<div class="profile-section">
 							<em
 								>Milestones capture any significant event in the history of this
-								archive, and are ordered from most recent to least recent.</em
+								archive, and by default are ordered from most recent to least
+								recent. This order can be changed in
+								<a
+									href="#"
+									(click)="openArchiveSettings(); $event.preventDefault()"
+									>archive settings</a
+								>.</em
 							>
 						</div>
 						<div class="profile-section">
--- a/src/app/core/components/profile-edit/profile-edit.component.ts
+++ b/src/app/core/components/profile-edit/profile-edit.component.ts
@@ -34,6 +34,7 @@ import { copyFromInputElement } from '@shared/utilities/forms';
 import { EventService } from '@shared/services/event/event.service';
 import { LocationPickerComponent } from '@fileBrowser/components/location-picker/location-picker.component';
 import { DIALOG_DATA, DialogRef } from '@angular/cdk/dialog';
+import { ArchiveSettingsDialogComponent } from '@core/components/archive-settings-dialog/archive-settings-dialog.component';
 import {
 	PROFILE_ONBOARDING_COOKIE,
 	ProfileEditFirstTimeDialogComponent,
@@ -259,6 +260,10 @@ export class ProfileEditComponent implements OnInit, AfterViewInit {
 			this.updateProgress();
 			this.trackProfileEdit(item);
 		}
+	}
+
+	openArchiveSettings() {
+		this.dialog.open(ArchiveSettingsDialogComponent);
 	}
 
 	async onShareClick() {
--- a/src/app/core/components/public-settings/public-settings.component.spec.ts
+++ b/src/app/core/components/public-settings/public-settings.component.spec.ts
@@ -39,7 +39,7 @@ const mockApiService = {
 			updatedValues: { milestoneSortOrder?: string },
 		) => {
 			if (patchArchiveThrowError) {
-				throw { error: { message: 'Test Error' } };
+				throw { error: { error: { message: 'Test Error' } } };
 			}
 			patchArchiveCalled = true;
 			patchArchiveCalledWith = { archiveId, updatedValues };
--- a/src/app/core/components/public-settings/public-settings.component.spec.ts
+++ b/src/app/core/components/public-settings/public-settings.component.spec.ts
@@ -18,6 +18,13 @@ let archive: ArchiveVO;
 let throwError: boolean = false;
 let updatedDownload: boolean = false;
 let updated: boolean = false;
+let patchArchiveThrowError: boolean = false;
+let patchArchiveCalled: boolean = false;
+let patchArchiveCalledWith: {
+	archiveId: string;
+	updatedValues: { milestoneSortOrder?: string };
+} | null = null;
+let showErrorCalled: boolean = false;
 const mockApiService = {
 	archive: {
 		update: async (data: any) => {
@@ -27,6 +34,16 @@ const mockApiService = {
 			updated = true;
 			updatedDownload = data.allowPublicDownload as boolean;
 		},
+		patchArchive: async (
+			archiveId: string,
+			updatedValues: { milestoneSortOrder?: string },
+		) => {
+			if (patchArchiveThrowError) {
+				throw { error: { message: 'Test Error' } };
+			}
+			patchArchiveCalled = true;
+			patchArchiveCalledWith = { archiveId, updatedValues };
+		},
 	},
 };
 
@@ -35,6 +52,10 @@ describe('PublicSettingsComponent', () => {
 		throwError = false;
 		updatedDownload = null;
 		updated = false;
+		patchArchiveThrowError = false;
+		patchArchiveCalled = false;
+		patchArchiveCalledWith = null;
+		showErrorCalled = false;
 		archive = {
 			allowPublicDownload: true,
 		} as ArchiveVO;
@@ -46,7 +67,9 @@ describe('PublicSettingsComponent', () => {
 			.provide({
 				provide: MessageService,
 				useValue: {
-					showError: () => {},
+					showError: () => {
+						showErrorCalled = true;
+					},
 				},
 			});
 	});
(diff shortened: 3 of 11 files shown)


src/app/core/components/profile-edit/profile-edit.component.spec.ts
```diff
@@ -153,4 +158,86 @@ describe('ProfileEditComponent', () => {
...
+		it('should default to reverse_chronological when milestoneSortOrder is not set', async () => {
+			component.archive = new ArchiveVO({ archiveId: '789' });
+
+			await component.toggleMilestoneOrder();
+
+			expect(mockApiService.archive.patchArchive).toHaveBeenCalledWith(
+				'789',
+				'chronological',
+			);
+		});
```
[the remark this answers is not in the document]
cecilia-donnelly: I suppose this is a little confusing, but I do think it's correct - when it's not set on creation, the sort order is always `reverse_chronological`. So the patch update from the toggle would be `chronological`.

src/app/models/archive-vo.ts
```diff
@@ -43,6 +43,7 @@ export class ArchiveVO extends BaseVO implements DynamicListChild {
 	public allowPublicDownload: boolean;
 	public payerAccountId: string;
 	public public: number;
+	public milestoneSortOrder: 'chronological' | 'reverse_chronological';
```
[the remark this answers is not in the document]
cecilia-donnelly: It is set for existing archives: https://github.com/PermanentOrg/back-end/commit/a7787acb1dbacea09afc4bec2bc3be448fe5337d

aasandei-vsp (commented):
Just a small comment.

src/app/core/components/profile-edit/profile-edit.component.ts
```diff
@@ -261,6 +262,10 @@ export class ProfileEditComponent implements OnInit, AfterViewInit {
 		}
 	}
 
+	openArchiveSettings() {
+		this.dialog.open(ArchiveSettingsDialogComponent, { width: '1000px' });
```
cecilia-donnelly, line 266: @aasandei-vsp we are a little inconsistent about how we open modals - obviously a set width here is not responsive. Do you have a suggestion about the right way to do this?

aasandei-vsp: This is definitely not ideal, but from what I see in the code, this is the default, hardcoding it to 1000px. Also, looking at the modal behavior, it does not seem to have a fixed 1000px width, as it is responsive and shrinks/widens with the screen. I have removed the config and the modal seems to behave the same. I'd recommend removing the width property completely and check the modal behavior. If there is anything not working correctly, than at most I would add a min-width of whatever you think should be, taking into account that are many screen sizes, but I doubt that would be necessary.(trial and error)

src/app/core/components/public-settings/public-settings.component.ts
```diff
@@ -47,12 +49,29 @@ export class PublicSettingsComponent implements OnInit {
...
+	public async onMilestoneSortOrderChange() {
+		this.archive.milestoneSortOrder = this.milestoneSortOrder;
+		this.updating = true;
+		try {
+			await this.api.archive.patchArchive(
+				this.archive.archiveId,
+				this.milestoneSortOrder,
+			);
+		} catch (err) {
+			this.msg.showError({ message: err.error?.message, translate: true });
```
aasandei-vsp, line 69: Have a look with the inspect at the error it should show. In my case, for patch record/folder, it was err.error.error. The err.error.message might be the more generic one related to response status.

src/app/shared/services/api/archive.repo.ts
```diff
@@ -272,6 +272,15 @@ export class ArchiveRepo extends BaseRepo {
 		).toPromise();
 	}
 
+	public async patchArchive(
+		archiveId: string,
+		milestoneSortOrder: 'chronological' | 'reverse_chronological',
```
aasandei-vsp, line 277: This type appears in multiple places. It would be better to just put it in one place, maybe in the archive model. Something like `type MilestoneSortOrder =  'chronological' | 'reverse_chronological'`.

src/app/core/components/public-settings/public-settings.component.ts
```diff
@@ -47,12 +49,29 @@ export class PublicSettingsComponent implements OnInit {
...
+		this.milestoneSortOrder =
+			this.archive.milestoneSortOrder ?? 'reverse_chronological';
 		this.archiveType =
 			this.archive.type === 'type.archive.family'
 				? 'type.archive.group'
 				: this.archive.type;
 	}
 
+	public async onMilestoneSortOrderChange() {
+		this.archive.milestoneSortOrder = this.milestoneSortOrder;
```
slifty, line 61: This assignment changes *before* the patch call is made, which means if the patch call fails we now have the client / server disagreeing.

I think the logic there is we want the state to update right away (does it result in a visual change?) rather than waiting on a network call to resolve.  If that's the case, can you store the "old" value before making this assignment, and revert to that old value in the catch block?

If that *isn't* the case, can we move this down to the try block / after the patch call?

src/app/core/components/public-settings/public-settings.component.ts
slifty, line 69: Related to what @aasandei-vsp notes -- since this message is server generated we won't have a matching i8n string in the front end.  Instead of rendering a server value can we detect the issue and render a web-app-defined value (with translation strings?)

src/app/shared/services/api/archive.repo.ts
```diff
@@ -272,6 +272,15 @@ export class ArchiveRepo extends BaseRepo {
 		).toPromise();
 	}
 
+	public async patchArchive(
```
slifty, line 275: This is generically named `patchArchive` but it takes a single required `milestoneSortOrder`

Should we name this `updateMilestoneSortOrder`?

Or, if it's intended to be a true patch method, can we define the parameters in a more "patchy" way (e.g. it takes a `patchValues: Partial<ArchiveVO>`, or even a constrained `patchValues: Partial<Pick<ArchiveVO, 'milestoneSortOrder'>>`)

src/app/core/components/profile-edit/profile-edit.component.html
```diff
@@ -455,7 +455,11 @@
 						<div class="profile-section">
 							<em
 								>Milestones capture any significant event in the history of this
-								archive, and are ordered from most recent to least recent.</em
+								archive, and by default are ordered from most recent to least
+								recent. This order can be changed in
+								<a href="javascript:void(0)" (click)="openArchiveSettings()"
```
slifty, line 460: I'm told by robots that "inline-js href is an a11y smell"

> A `<button class="link-styled">` (or anchor with           
> role="button" + (keydown.enter)) preserves keyboard + screen-reader semantics.

src/app/public/components/public-profile/public-profile.component.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+		{ fieldNameUI: 'profile.milestone', string1: 'Oldest', day1: '1990-01-01' },
+		{ fieldNameUI: 'profile.milestone', string1: 'Middle', day1: '2000-06-15' },
+		{ fieldNameUI: 'profile.milestone', string1: 'Newest', day1: '2010-12-31' },
+	];
+
+	beforeEach(async () => {
+		await TestBed.configureTestingModule({
+			declarations: [PublicProfileComponent],
+			providers: [PublicProfileService],
+			schemas: [NO_ERRORS_SCHEMA],
```
slifty, line 24: This is also a claude-identified issue:

> NO_ERRORS_SCHEMA — this silences all unknown-element/attribute errors in the template, which can hide real regressions. Prefer importing the needed child components or using `CUSTOM_ELEMENTS_SCHEMA` only where necessary.

src/app/core/components/public-settings/public-settings.component.spec.ts
```diff
@@ -108,4 +125,102 @@ describe('PublicSettingsComponent', () => {
...
+		it('should set updating to true while saving and reset to false after', async () => {
+			defaultRender({
+				allowPublicDownload: false,
+				archiveId: 'abc123',
+			} as ArchiveVO);
+			const instance = ngMocks.findInstance(PublicSettingsComponent);
+
+			let updatingWhileSaving: boolean;
+			const originalPatchArchive = mockApiService.archive.patchArchive;
+			mockApiService.archive.patchArchive = async (id, order) => {
```
slifty, line 213: There's a weird race condition here!  This modifies a shared mock.  It is restored later (line 223), but if there is ever an exception in the test then we'd never get to line 223, and so future tests will use this modified mock.

We could use a try / finally to cover that case, we could make the mock itself more robust and add a flag, or we could spy on the mock (e.g. `spyOn(mockApiService.archive, 'patchArchive').and.callFake(...)`)

slifty (changes requested):

src/app/core/components/public-settings/public-settings.component.ts
cecilia-donnelly: 100% correct, thanks!

cecilia-donnelly: Yes, you're right - I was thinking that I'd make the stela error nicer, but I can use the existing setup for translating server errors to human-readable text.

src/app/core/components/public-settings/public-settings.component.ts
cecilia-donnelly: I think this is leftover from when I had this in a different place -- it does result in a visual change but on a completely different page, so I think it's fine to have it after the network call. Thanks!

src/app/shared/services/api/archive.repo.ts
cecilia-donnelly: Right! I think it should be a real patchArchive since we'll presumably add more fields to the stela counterpart in the future. Thanks :)

src/app/core/components/public-settings/public-settings.component.spec.ts
cecilia-donnelly: Ahhh, this is tricky, thanks! It looks like we use `spyOn` elsewhere so I will do that.

slifty (approved):
Looks great!  Thank you for the iteration :D

cecilia-donnelly:
To test:
1. Add milestones to an archive profile, then open the profile in a new tab so you can check the order there.
2. Open archive settings from the private workspace and change the order. Go to the profile tab and refresh to see if the new order is reflected correctly.
3. Go to archive profile and open archive settings from the link there.
4. Try to make the API return an error. If you can, check to make sure that the error appears correctly. It will *not* appear in the profile editor (that's an out of scope fix -- the error is there, just behind the profile editor), but it should appear if you open archive settings normally from the workspace. Unfortunately I don't have a good idea of how to make an error happen without directly manipulating the API server.

The error text is coming directly from stela right now, but I have a companion change there to have it use the translation service.

omnignorant (approved):
This is looking and working fantastically.

PR #964, opened 2026-03-11
Remove Onboarding feature flag

slifty:
This PR removes the old onboarding code, meaning the new glam onboarding (which is already live everywhere) is the only onboarding we need to worry about going forward.

(this PR is a draft until the changes to onboarding are merged in #952)


aasandei-vsp (approved):

cecilia-donnelly:
- regular signup
- signup with a specific code
- signup from share
- signup from invitation
- signup from invitation to an archive

omnignorant:
working through this and not sure how to test the "signup with a specific code" route. I guess I'd need to know an invitation code on dev? I'm not sure if they're the same as the codes on prod.

omnignorant (approved):
Tested all the routes except for invitation code. Only issues that I think are worth noting here.
* from share by email — got a fleeting yellow error message after creating archive before landing in private files but I successfully reached the shared folder as expected.
* I am not getting Mailchimp welcome emails on dev.

Basically, seems to be working as expected!

I did find a number of small areas of improvement but those are unrelated to this.

PR #971, opened 2026-03-20
[PER-10475] Add patch folder

aasandei-vsp:
Issue: https://permanent.atlassian.net/browse/PER-10475

 STEPS TO TEST FOLDER DATE:

1. Log in or create a new account
2. Navigate to your files and create a new folder or select an existing folder
3. Click on the folder to open the details sidebar
4. Click on the "Start date" field and modify it
EXPECTED: The start date field updates and displays the new value
5. Click on the "End date" field and modify it
EXPECTED: The end date field updates and displays the new value
6. Refresh the page
EXPECTED: Both start and end dates persist and display correctly
7. Modify only the end date without changing the start date
EXPECTED: The end date updates while the start date remains unchanged

STEPS TO TEST RECORD DATE:

1. Log in or create a new account
2. Upload a new single record
3. Click on the single record, so the details sidebar appears
4. Modify the date and/or time of the single record that appears in the sidebar
EXPECTED: The date/time field will be updated
5. Double click on the single record
6. Modify the date and/or time of the single record that appears in the right side of the preview
EXPECTED: The date/time field will be updated

!!!!OBSERVATION: Just to be extra sure, testing the update of another property from the sidebar, like the description would be useful.

### DIFF
--- a/src/app/core/services/edit/edit.service.spec.ts
+++ b/src/app/core/services/edit/edit.service.spec.ts
@@ -47,6 +47,9 @@ describe('EditService', () => {
 		} as unknown as RecordRepo;
 		apiService.folder = {
 			getFolderShareLink: jasmine.createSpy('getFolderShareLink'),
+			update: jasmine.createSpy('update'),
+			updateStelaFolder: jasmine.createSpy('updateStelaFolder'),
+			getStelaFolderVOs: jasmine.createSpy('getStelaFolderVOs'),
 		} as unknown as FolderRepo;
 		apiService.share = {
 			getShareLink: jasmine.createSpy('getShareLink'),
@@ -122,8 +125,11 @@ describe('EditService', () => {
 		expect(apiService.record.update).not.toHaveBeenCalled();
 	});
 
-	it('should call updateStelaRecord when recordKey is displayDT', async () => {
-		const record = new RecordVO({ recordId: 1, displayDT: '2024-01-01' });
+	it('should call updateStelaRecord when recordKey is displayTime', async () => {
+		const record = new RecordVO({
+			recordId: 1,
+			displayTime: '2024-01-01',
+		});
 		const updatedRecord = new RecordVO({
 			recordId: 1,
 			updatedDT: '2024-03-03',
@@ -136,21 +142,18 @@ describe('EditService', () => {
 		(apiService.record.updateStelaRecord as jasmine.Spy).and.returnValue(
 			Promise.resolve(mockResponse),
 		);
-		(apiService.record.update as jasmine.Spy).and.returnValue(
-			Promise.resolve([{ updatedDT: '2024-03-03' }]),
-		);
 		(apiService.record.get as jasmine.Spy).and.returnValue(
 			Promise.resolve(mockResponse),
 		);
 
-		await service.updateItems([record], ['displayDT']);
+		await service.updateItems([record], ['displayTime']);
 
 		expect(apiService.record.updateStelaRecord).toHaveBeenCalledWith(record);
-		expect(apiService.record.update).toHaveBeenCalled();
+		expect(apiService.record.update).not.toHaveBeenCalled();
 		expect(apiService.record.get).toHaveBeenCalledWith([record]);
 	});
 
-	it('should not call updateStelaRecord when recordKey does not contain displayDT', async () => {
+	it('should not call updateStelaRecord when recordKey does not contain displayTime', async () => {
 		const record = new RecordVO({ recordId: 1, displayName: 'Test' });
 		const updatedRecord = new RecordVO({
 			recordId: 1,
@@ -173,10 +176,10 @@ describe('EditService', () => {
 		expect(apiService.record.get).toHaveBeenCalledWith([record]);
 	});
 
-	it('should call both updateStelaRecord and update when recordKey has displayDT with other properties', async () => {
+	it('should call both updateStelaRecord and legacy update when recordKey has displayTime with other properties', async () => {
 		const record = new RecordVO({
 			recordId: 1,
-			displayDT: '2024-01-01',
+			displayTime: '2024-01-01',
 			displayName: 'Test',
 		});
 		const updatedRecord = new RecordVO({
@@ -197,7 +200,7 @@ describe('EditService', () => {
 			Promise.resolve(mockResponse),
 		);
 
-		await service.updateItems([record], ['displayDT', 'displayName']);
+		await service.updateItems([record], ['displayTime', 'displayName']);
 
 		expect(apiService.record.updateStelaRecord).toHaveBeenCalledWith(record);
 		expect(apiService.record.update).toHaveBeenCalled();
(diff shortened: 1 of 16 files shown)


cecilia-donnelly (commented):
Some thoughts, particularly around the use of `any` which we're trying to avoid.

src/app/core/services/edit/edit.service.ts
```diff
@@ -419,6 +415,18 @@ export class EditService {
 						newData.TimezoneVO = updatedItem.TimezoneVO;
 					}
 
+					if ((updatedItem as any).displayTime) {
+						(newData as any).displayTime = (updatedItem as any).displayTime;
+					}
```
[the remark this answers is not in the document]
aasandei-vsp: That is true! I've added it both in FolderVO and FolderVOData. I did that for record so it was the right way to do it here as well.

src/app/shared/services/api/folder.repo.ts
```diff
@@ -199,16 +214,52 @@ export class FolderRepo extends BaseRepo {
...
+	}
+
+	public async updateStelaFolder(folderVO: FolderVO): Promise<FolderResponse> {
+		const displayTime = this.buildEdtfInterval(
+			folderVO.displayDT,
+			folderVO.displayEndDT,
+		);
+
+		const payload = {
+			displayTime,
```
cecilia-donnelly: Wouldn't this need to be "displayTime": displayTime?

cecilia-donnelly (commented):
@aasandei-vsp I'm so sorry for being slow on this! I am a bit confused - is the UI supposed to be calling the new PATCH endpoint yet? It doesn't look like it is but your testing steps suggest that there is something to test in the UI. Changing the folder dates does work correctly, but it looks like they're still using the old `folder/update` endpoint.

slifty (dismissed):
A few suggestions for your consideration!

src/app/shared/services/api/folder.repo.ts
```diff
@@ -170,14 +170,29 @@ export class FolderRepo extends BaseRepo {
...
+		folderVOs: FolderVO[],
 		shareToken: string = null,
-	): Promise<StelaFolder> {
+	): Promise<StelaFolder[]> {
 		const queryData = {
-			folderIds: [folderVO.folderId],
+			folderIds: folderVOs.map((currentFolder) => currentFolder.folderId),
 		};
-		let folderResponse: PagedStelaResponse<StelaFolder>;
+		let folderResponse: PagedStelaResponse<StelaFolder> | any;
```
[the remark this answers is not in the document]
aasandei-vsp: I understand the confusion, but at the moment both the old update endpoint and the new one are called.
The old update endpoint supports changing multiple properties at the same time.
The new patch endpoint only supports changing the displayTime property.
The reason we are calling both of them is because we could be trying to update multiple properties at the same time and if one of them is related to date, make sure to use the patch folder endpoint for it, otherwise, just do the old update.
Because the update calls are made in parallel and are both updating the same object , we cannot rely on the response from either of the update calls, that's why after the update promises are resolved, a call to get the most recent version of the folder is needed.

cecilia-donnelly (commented):
We talked about this a bit in demos this morning, so maybe it's not expected to work here and only in the main dates PR - currently if I enter a "start date" later than the "end date" I get a 400 response from the new API call that is not shown to the user and the old API call succeeds. What's the expected behavior in this PR?

I know the edit service is difficult to deal with but I am concerned that the double call is not the right pattern.

src/app/shared/services/api/folder.repo.spec.ts
```diff
@@ -210,4 +210,197 @@ describe('Folder repo', () => {
...
+				'200': 'url200',
+				'256': 'url256',
+				'500': 'url500',
+				'1000': 'url1000',
+				'2000': 'url2000',
+			},
+			type: 'folder',
+			status: 'ok',
+			view: 'grid',
+		};
```
cecilia-donnelly: Is there a reason to define this twice?

src/app/core/services/edit/edit.service.spec.ts
```diff
@@ -220,6 +223,231 @@ describe('EditService', () => {
...
+		expect(apiService.folder.update).toHaveBeenCalledWith(mockFolders, [
+			'displayName',
+		]);
+
+		expect(apiService.folder.getStelaFolderVOs).toHaveBeenCalledWith(
+			mockFolders,
+		);
+	});
+
+	it('should call both updateStelaFolder and legacy update when whitelist has displayTime with other properties', async () => {
```
cecilia-donnelly, line 404: Currently we don't send multiple properties in one update call. I don't think we ever *want* to call both endpoints, because of the race condition @slifty mentioned. Let's remove this test and probably even be defensive about it in the edit service itself.

src/app/core/services/edit/edit.service.ts
```diff
@@ -680,9 +695,44 @@ export class EditService {
 		}
 
 		const archiveId = this.accountService.getArchive().archiveId;
-		promises.push(this.api.record.update(records, archiveId));
+		const legacyRecordKey = recordKey?.filter((k) => k !== 'displayTime');
+
+		if (!recordKey || legacyRecordKey?.length > 0) {
```
cecilia-donnelly, line 700: Possibly here we should guard against the case where we are sending both `displayTime` and legacy keys, so that we know we're *only* calling the old endpoint.

src/app/core/services/edit/edit.service.ts
```diff
@@ -680,9 +695,44 @@ export class EditService {
...
+					folders.map(
+						async (folder) => await this.api.folder.updateStelaFolder(folder),
+					),
+				),
+			);
+		}
+
+		const legacyFolderKeys = folderKeys?.filter((k) => k !== 'displayTime');
+
+		if (!folderKeys || legacyFolderKeys?.length > 0) {
```
cecilia-donnelly, line 730: Same here.

cecilia-donnelly (commented):
Thanks for separating out these commits so cleanly!

The part that worries me is the case where we might send both requests, so I added some comments about making sure we don't do that inline. Thank you so much for all the rounds of work on this @aasandei-vsp ! Looks good otherwise.

src/app/core/services/edit/edit.service.ts
aasandei-vsp: Looking at the signature of the method that calls the private updateRecords:
`async updateItems( items: any[], whitelist?: (keyof ItemVO)[], ): Promise<FolderResponse | RecordResponse | any>`
I am constrained to actually receive a list of keys for properties that need to be updated. I believe that changing the whole updateItems method is out of this scope and even though I haven't found a place where multiple keys are used, I am not 100% sure that this is the case.

src/app/core/services/edit/edit.service.ts
aasandei-vsp: Looking at the signature of the method that calls the private updateFolders:
`async updateItems( items: any[], whitelist?: (keyof ItemVO)[], ): Promise<FolderResponse | RecordResponse | any>`
I am constrained to actually receive a list of keys for properties that need to be updated. I believe that changing the whole updateItems method is out of this scope and even though I haven't found a place where multiple keys are used, I am not 100% sure that this is the case.

src/app/core/services/edit/edit.service.spec.ts
aasandei-vsp: The way update is built(both FE and BE also I would assume), you can actually send multiple properties for update, even if we indeed do not do it( I think, I am not 100% sure, I haven't found a place for it).
There is no racing condition, because we do not use any of the responses from update. The reason we are not doing it is because not all updates(patch record is the one missing the response object) send back the updated object.
So we first wait for the update promises to resolve and then we make a get call, to have the most recent record/folder.

src/app/shared/services/api/folder.repo.spec.ts
aasandei-vsp: Yes! But it is a bad reason :)) There are 2 different describe blocks. But taking into account that we are not actually modifying the object, it can safely be moved outside, defined once and be used everywhere.

src/app/core/services/edit/edit.service.ts
cecilia-donnelly: Right, that makes sense! What if `updateItems` calls either `updateRecordLegacyKey` or `updateStelaRecord` (names not important) on *one* field each, and for each field we route to the correct method? That way we always know what's going to happen rather than being unsure if one or both endpoints are being called.

That would involve splitting this `updateRecords` method into those two methods instead of having conditionals inside it. I don't think it would add more network calls because in practice we only update one field at a time but let me know if I'm way off.

cecilia-donnelly (commented):
Throwing out another idea.

aasandei-vsp:
@cecilia-donnelly Hm, not a bad idea at all! And definitely you are not off, let me tinker around and implement this. I think it should work well.

cecilia-donnelly (approved):
Woohoo! Thank you so much for all these conversations and updates, @aasandei-vsp. It looks awesome. I am excited to get this out!

slifty:
I tested folder and records -- it saved, persisted, and used the new endpoint!

slifty (approved):

omnignorant (changes requested):
Editing dates in the right hand side bar seems to work AOK. However, editing them in the full screen view does not work every time. Steps to reproduce:
* double click to view record in full screen
* change date in full screen view --> works!
* click on same record
* change date in right hand side info tab --> works!
* double click on same record to view in full screen
* change date in full screen view --> **does not work!**

problem even persists after refresh.

omnignorant:
screencapture showing the issue
https://github.com/user-attachments/assets/31afd056-eba1-4e6d-9b68-b98d9a01f5a9

aasandei-vsp:
> screencapture showing the issue https://github.com/user-attachments/assets/31afd056-eba1-4e6d-9b68-b98d9a01f5a9

Thank you! Managed to reproduce it and now it's fixed!

omnignorant (approved):
appears to be working now!

PR #972, opened 2026-03-20
[PER-10489] Fix unlisted share thumbnail display

aasandei-vsp:
STEPS FOR TESTING

**Restricted share and not logged in**
1. Log in or create an account
2. Upload a single record
3. Create a share link with the link type restricted for the uploaded record
4. Copy the share link
5. Open a new window using the incognito mode
6. Paste the share link
EXPECTED: The thumbnail of the shared single record should be a random one, not the record's real image.

**Restricted share and logged in**
1. Log in or create an account
2. Upload a single record
3. Create a share link with the link type restricted for the uploaded record
4. Copy the share link
5. Open a new window
6. Log in or create a new account
7. Paste the share link
EXPECTED: The thumbnail of the shared single record should be a random one, not the record's real image.
8. Accept the share
EXPECTED: The thumbnail of the shared single record in the shared workspace should be the record's real image.

**Unlisted share and not logged in**
1. Log in or create an account
2. Upload a single record
3. Create a share link with the link type anyone-can-view for the uploaded record
4. Copy the share link
5. Open a new window using the incognito mode
6. Paste the share link
EXPECTED: The thumbnail of the shared single record should be the record's real image.

**Unlisted share and logged in**
1. Log in or create an account
2. Upload a single record
3. Create a share link with the link type anyone-can-view for the uploaded record
4. Copy the share link
5. Open a new window
6. Log in or create a new account
7. Paste the share link
EXPECTED: The thumbnail of the shared single record should be the record's real image.

Issue: PER-10489

### DIFF
--- a/src/app/file-browser/components/file-list-item/file-list-item.component.html
+++ b/src/app/file-browser/components/file-list-item/file-list-item.component.html
@@ -47,7 +47,7 @@
 			</div>
 		}
 		@if (!item.isFolder && !isZip) {
-			<div prBgImage [bgSrc]="item | getThumbnail" [cover]="inGridView"></div>
+			<div prBgImage [bgSrc]="recordThumbnailUrl" [cover]="inGridView"></div>
 		}
 	</ng-container>
 	@if (inGridView && item.isFolder ? item.view : true) {
--- a/src/app/file-browser/components/file-list-item/file-list-item.component.spec.ts
+++ b/src/app/file-browser/components/file-list-item/file-list-item.component.spec.ts
@@ -1,7 +1,7 @@
 import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { ElementRef, Pipe, PipeTransform } from '@angular/core';
 import { of } from 'rxjs';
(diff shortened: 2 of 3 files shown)


cecilia-donnelly (commented):
This may be a bit more complicated because of the existence of the preview toggle for older share links. See https://permanent.atlassian.net/browse/PER-10333, which asks how these should be handled.

I know there was some conversation in Slack about this, but I'm not sure the question in PER-10333 was exactly answered. I think the key pieces are:

1. New unlisted share links - preview always on, works 👍 
2. New restricted share links - preview always off, works 👍 
3. Old share links (were always restricted) - preview depends on preview toggle setting. These will not show a preview even if the preview toggle was on, as I read this code. The problem is that we don't *know* whether the preview toggle was on or off because the new endpoint doesn't return that information. The note in the investigation ticket suggests showing some helptext, but ultimately this is a question for @omnignorant -- do we care about old restricted share links not showing previews when they originally would have?

cecilia-donnelly (approved):
Thanks for the conversation!

omnignorant (approved):
worked exactly as described in the testing steps

PR #994, opened 2026-04-16
PER-10528: Point top left logo to marketing site

cecilia-donnelly:
Instead of the root, this needs to explicitly point to the marketing site now that it's on a different host.


cecilia-donnelly:
I think not adding tests for this is okay, even though code coverage is complaining, but let me know what you think!

slifty (approved):

cecilia-donnelly:
Moving forward without QA on this one since it's a very small change.

PR #998, opened 2026-04-23
[PER-10451] Functional improvements to archive switcher

aasandei-vsp:
STEPS TO TEST:

**See the new design of the add to archive modal**
1. Log in to an existing account or create a new one
2. Upload a new file or add a new folder
3. Click the new file or folder and from the sidebar, open the share modal
4. Create a share link and select the restricted link type option
5. Copy the newly created link
6. In a different browser window, log in to another account or create a new one
7. After logging in, click on the email and from the dropdown choose Archives
8. Create multiple new archives
9. In the current browser window, paste the restricted share link
10. Click the "Accept share" button from the sticky footer.
EXPECTED: A modal will appear: 
 - The title: "Select archive to request access"
 - The list of all archives from the current account
 - The currently selected archive will have an orange background
 - There will be a "Create new archive" button
 - There will be a Cancel button
 
 **Click the cancel button**
 
 1. Follow the previous steps from **See the new design of the add to archive modal**
 2. Click the cancel button
 EXPECTED: The modal will disappear and the sticky footer with the "Accept share" button will appear again
 
 **Accept share in the current archive**
 1. Follow the previous steps from **See the new design of the add to archive modal**
 2. Click the current archive, the one with the orange background.
 EXPECTED: The modal will close and the user will be redirected to his share workspace, of his currently selected archive.
 
  **Accept share in a different archive**
 1. Follow the previous steps from **See the new design of the add to archive modal**
 2. Click any archive from the list, except the current one, that has an orange background.
 EXPECTED: A small modal appears on top of the archives one, with the buttons "Switch archive" and cancel.
 3. Click the "Switch archive" button
 EXPECTED: The modal will close and the user will be redirected to his share workspace, of the archive that has been selected in the modal.
 
  **Accept share in a newly created archive**
1. Follow the previous steps from **See the new design of the add to archive modal**
2. Click the "Create new archive" button
EXPECTED: The create new archive modal will appear.
3. Fill the form.
4. Click "Create archive"
EXPECTED: The "Switch archive" modal will appear.
5. Click the "Switch archive" button
EXPECTED: The modal will close and the user will be redirected to his share workspace, of the archive that has been created.

### DIFF
--- a/src/app/shared/components/archive-switcher/archive-switcher.component.spec.ts
+++ b/src/app/shared/components/archive-switcher/archive-switcher.component.spec.ts
@@ -6,12 +6,10 @@ import {
 } from '@angular/core/testing';
 import { AccountService } from '@shared/services/account/account.service';
 import { ApiService } from '@shared/services/api/api.service';
-import { DataService } from '@shared/services/data/data.service';
 import { PrConstantsService } from '@shared/services/pr-constants/pr-constants.service';
 import { PromptService } from '@shared/services/prompt/prompt.service';
 import { MessageService } from '@shared/services/message/message.service';
-import { Router } from '@angular/router';
-import { ArchiveVO, FolderVO } from '@root/app/models';
+import { ArchiveVO } from '@root/app/models';
 import { Component } from '@angular/core';
 import { DialogRef } from '@angular/cdk/dialog';
 import { ArchiveSwitcherComponent } from './archive-switcher.component';
@@ -19,7 +17,7 @@ import { ArchiveSwitcherComponent } from './archive-switcher.component';
 @Component({
 	selector: 'pr-archive-small',
 	template: '',
-	standalone: true, // ✅ This makes it standalone
+	standalone: true,
 })
 class MockArchiveSmallComponent {}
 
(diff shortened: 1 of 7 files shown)


src/app/shared/components/archive-switcher/archive-switcher.component.spec.ts
```diff
@@ -6,20 +6,18 @@ import {
...
-import { ArchiveVO, FolderVO } from '@root/app/models';
+import { ArchiveVO } from '@root/app/models';
 import { Component } from '@angular/core';
 import { DialogRef } from '@angular/cdk/dialog';
 import { ArchiveSwitcherComponent } from './archive-switcher.component';
 
 @Component({
 	selector: 'pr-archive-small',
 	template: '',
-	standalone: true, // ✅ This makes it standalone
```
cecilia-donnelly, line 22: What a helpful comment ha.

src/app/shared/components/archive-switcher/archive-switcher.component.spec.ts
```diff
@@ -124,114 +120,250 @@ describe('ArchiveSwitcherComponent', () => {
...
+
+		it('should not call setArchive when current archive is not in the fetched list', fakeAsync(() => {
+			mockAccountService.getArchive.and.returnValue(
+				new ArchiveVO({ archiveId: '999', fullName: 'Unknown' }),
+			);
+			component.ngOnInit();
+			tick();
+
+			expect(mockAccountService.setArchive).not.toHaveBeenCalled();
+		}));
```
cecilia-donnelly, line 150: Why would the current archive not be in the fetched list? (Just curious)

src/app/shared/components/archive-switcher/archive-switcher.component.spec.ts
```diff
@@ -124,114 +120,250 @@ describe('ArchiveSwitcherComponent', () => {
...
+			);
+
+			component.archiveClick(otherArchive);
+			tick();
+
+			expect(mockAccountService.changeArchive).not.toHaveBeenCalled();
+			expect(mockDialogRef.close).not.toHaveBeenCalled();
+		}));
+
+		it('should call accept before changeArchive for a pending archive', fakeAsync(() => {
```
cecilia-donnelly, line 218: This is interesting. I didn't realize that pending archives were returned by this endpoint.

I do like saying "if you want to put something in this archive then implicitly you're saying you want to be a member of the archive," but I think it would be clearer if pending archives didn't appear in the list of archives -- obviously a backend change!

cecilia-donnelly (approved):
This looks great! Thanks @aasandei-vsp.

omnignorant (dismissed):
This is working nearly perfectly, but there is a small change needed.

At minimum the new archive modal in share flow should work the same as the new archive modal in account settings, e.g. selecting the type toggles the relationship dropdown on/off. However, if we can eliminate that dropdown altogether and just default the relationship to "other". We do not really use this feature in the app at all.

aasandei-vsp:
> This is working nearly perfectly, but there is a small change needed.
> 
> At minimum the new archive modal in share flow should work the same as the new archive modal in account settings, e.g. selecting the type toggles the relationship dropdown on/off. However, if we can eliminate that dropdown altogether and just default the relationship to "other". We do not really use this feature in the app at all.

@omnignorant  I can make the change in this PR, but I think it might be more relevant to do it on the task that we already have in Jira:
https://permanent.atlassian.net/browse/PER-10385
This task involves doing some UI changes as well, which I think would make the experience better.
I think I could drag it this sprint, but I would need a bit of help with the design.

How would you like me to proceed?

omnignorant:
Are you saying that the new archive modal appears elsewhere so we should change it everywhere it appears? e.g. PER-10385 is for all create new archive flows? Is this because the create new archive modal that appears here is a component that is used elsewhere? So the issues I've pointed out here are also issues elsewhere?

Therefore, are you suggesting that we address my suggested changes separately because this PR is for the archive switcher and not for the create new archive modal?

If so, then yes. You can add my comments above to PER-10385 and merge this PR.

aasandei-vsp:
> Are you saying that the new archive modal appears elsewhere so we should change it everywhere it appears? e.g. PER-10385 is for all create new archive flows? Is this because the create new archive modal that appears here is a component that is used elsewhere? So the issues I've pointed out here are also issues elsewhere?
> 
> Therefore, are you suggesting that we address my suggested changes separately because this PR is for the archive switcher and not for the create new archive modal?
> 
> If so, then yes. You can add my comments above to PER-10385 and merge this PR.

@omnignorant This task PER-10385 was specifically created for the create new archive modal from the share flow. I would prefer to make the changes mentioned above on that specific task, because we would need some design. Right now, the modal for all archives has a very nice design made by Tibi and it would be nice for the create new archive modal from this flow to have a similar one.

So I'll just merge this one and add a comment with the requested changes on the task mentioned above.

PR #999, opened 2026-04-23
PER-10549: Convert double slashes to fix prod routing

cecilia-donnelly:
We use double slashes in Angular routing which WPEngine is collapsing with nginx's `merge_slashes` setting, making some urls unreachable. Notably, this affects the "redeem your promo code" link we send in our welcome email. This is a heavily Claude-guided remediation effort, so please tell me about any better way to do this, @aasandei-vsp!

### DIFF
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -12,6 +12,7 @@ import {
 	NavigationEnd,
 	ActivatedRoute,
 	NavigationStart,
+	UrlSerializer,
 } from '@angular/router';
 import {
 	HttpErrorResponse,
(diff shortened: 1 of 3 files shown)


aasandei-vsp (approved):
Ok, so I do not actually have experience with WPEngine, but I did do some research and it seems to be the correct solution for this situation.

From an architectural point of view it is the place to add this serializer.

I also looked if there is any package for this, but I didn't find any. I do believe it would have been too much to add something like that, when this solution is clean.

I did check this URL: https://local.permanent.org/app/(private//dialog:account) and it's working, but I guess just working on local is not saying that much.

LGTM

omnignorant (approved):
This is looking good. The modals I could find that use the /_/ links:

- account settings
- storage
- invitations
- profile
- connections
- members
- archive settings

PR #1000, opened 2026-04-24
[PER-20503] Update the image and text on the login screen

aasandei-vsp:
STEPS TO TEST:

1. Go to the login page
2. Check that the image from the left is the same with the one provided: https://drive.google.com/drive/folders/12--YowlE73jy8aX8hIwau_H2ZV4GP6J0
3. Check that the text on the image, in the white rectangle is the one from here: https://docs.google.com/document/d/1K0eLiD0nL_c6gPb6SV2JVKsQBkqnoKyZbH2WffrF_s0/edit?tab=t.0
4. Check that when you click the text "Learn more about this feature", you get redirected to: https://www.permanent.org/blog/you-can-now-share-your-private-archive-with-just-a-link/

### DIFF
--- a/src/app/auth/components/auth/auth.component.html
+++ b/src/app/auth/components/auth/auth.component.html
@@ -3,16 +3,12 @@
 	<div class="auth-container">
 		<div class="image">
 			<div class="testimony">
-				<p class="text">
-					At Permanent, we’re committed to helping communities preserve their
-					legacies. <b>Byte4Byte</b> is our storage grant program for nonprofits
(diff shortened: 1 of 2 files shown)


cecilia-donnelly (approved):
This looks great! I resized the screen and it was beautifully responsive. It also looks like a good choice to use a font size instead of the more complicated CSS we had before - thanks for that.

aasandei-vsp:
Alexa confirmed that the image and text are correct, so this can be merged.

PR #1002, opened 2026-04-27
Use new location metadata fields

### LINKED ISSUE #1001: Use new location metadata fields
We're working to update the location metadata system in permanent, which will come with some new fields.  We need to update our code to use these fields!

slifty:
This PR modifies the web app to use the new location metadata fields for both reading and writing data.

Resolves #1001

### DIFF
--- a/src/app/file-browser/components/location-picker/location-picker.component.spec.ts
+++ b/src/app/file-browser/components/location-picker/location-picker.component.spec.ts
@@ -1,25 +1,216 @@
-// import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+import { CUSTOM_ELEMENTS_SCHEMA } from '@angular/core';
+import { TestBed, ComponentFixture } from '@angular/core/testing';
+import { ApiService } from '@shared/services/api/api.service';
+import { MessageService } from '@shared/services/message/message.service';
+import { EditService } from '@core/services/edit/edit.service';
+import { ProfileService } from '@shared/services/profile/profile.service';
+import { PrLocationPipe } from '@shared/pipes/pr-location.pipe';
+import { LocationPickerComponent } from './location-picker.component';
 
-// import { LocationPickerComponent } from './location-picker.component';
+const fakeGoogleMaps = {
+	LatLng: class {
+		constructor(public input: unknown) {}
+	},
+	places: {
+		Autocomplete: class {
+			setFields(): void {}
+			addListener(): void {}
+			getPlace(): unknown {
+				return null;
+			}
+		},
+	},
+};
 
-// describe('LocationPickerComponent', () => {
-//   let component: LocationPickerComponent;
-//   let fixture: ComponentFixture<LocationPickerComponent>;
+const buildAddressComponents = (
+	overrides: Partial<
+		Record<string, { long_name: string; short_name: string }>
+	> = {},
+): google.maps.GeocoderAddressComponent[] => {
+	const defaults: Record<string, { long_name: string; short_name: string }> = {
+		street_number: { long_name: '55', short_name: '55' },
+		route: { long_name: 'Rue Plumet', short_name: 'Rue Plumet' },
+		locality: { long_name: 'Paris', short_name: 'Paris' },
+		postal_code: { long_name: '75007', short_name: '75007' },
+		administrative_area_level_1: {
+			long_name: 'Ile-de-France',
+			short_name: 'IDF',
+		},
+		country: { long_name: 'France', short_name: 'FR' },
+	};
+	const merged = { ...defaults, ...overrides };
+	return Object.entries(merged)
+		.filter(([, value]) => value !== undefined)
+		.map(([type, value]) => ({
+			long_name: value.long_name,
+			short_name: value.short_name,
+			types: [type],
+		})) as google.maps.GeocoderAddressComponent[];
+};
 
-//   beforeEach(async(() => {
-//     TestBed.configureTestingModule({
-//       declarations: [ LocationPickerComponent ]
-//     })
-//     .compileComponents();
-//   }));
+const buildPlace = (
+	overrides: Partial<google.maps.places.PlaceResult> = {},
+	addressOverrides:
+		| Parameters<typeof buildAddressComponents>[0]
+		| undefined = undefined,
+): google.maps.places.PlaceResult =>
+	({
+		name: "Jean Valjean's House",
+		address_components: buildAddressComponents(addressOverrides),
+		geometry: {
+			location: {
+				lat: () => 48.83,
+				lng: () => 2.3,
+			},
+		},
+		...overrides,
+	}) as unknown as google.maps.places.PlaceResult;
 
-//   beforeEach(() => {
-//     fixture = TestBed.createComponent(LocationPickerComponent);
-//     component = fixture.componentInstance;
-//     fixture.detectChanges();
-//   });
+describe('LocationPickerComponent', () => {
+	let fixture: ComponentFixture<LocationPickerComponent>;
+	let component: LocationPickerComponent;
+	const testWindow = window as unknown as { google?: unknown };
+	let previousGoogle: unknown;
+	let hadGoogle = false;
 
-//   it('should create', () => {
-//     expect(component).toBeTruthy();
-//   });
-// });
+	beforeAll(() => {
+		hadGoogle = 'google' in testWindow;
+		previousGoogle = testWindow.google;
+		testWindow.google = { maps: fakeGoogleMaps };
+	});
+
+	afterAll(() => {
+		if (hadGoogle) {
+			testWindow.google = previousGoogle;
+		} else {
+			delete testWindow.google;
+		}
+	});
+
+	beforeEach(async () => {
+		await TestBed.configureTestingModule({
+			declarations: [LocationPickerComponent],
+			providers: [
+				{ provide: ApiService, useValue: {} },
+				{ provide: MessageService, useValue: {} },
+				{ provide: EditService, useValue: {} },
+				{ provide: ProfileService, useValue: {} },
+				{ provide: PrLocationPipe, useValue: { transform: () => ({}) } },
+			],
+			schemas: [CUSTOM_ELEMENTS_SCHEMA],
+		}).compileComponents();
+
+		fixture = TestBed.createComponent(LocationPickerComponent);
+		component = fixture.componentInstance;
+	});
+
+	describe('createLocnFromPlace', () => {
+		it('populates the spec-aligned fields', () => {
+			const locn = component.createLocnFromPlace(buildPlace());
+
+			expect(locn.sublocation).toBe('55 Rue Plumet');
+			expect(locn.city).toBe('Paris');
+			expect(locn.adminOneName).toBe('Ile-de-France');
+			expect(locn.country).toBe('France');
+			expect(locn.postalCode).toBe('75007');
+			expect(locn.latitude).toBe(48.83);
+			expect(locn.longitude).toBe(2.3);
+		});
+
+		it('does not write deprecated legacy fields', () => {
+			const locn = component.createLocnFromPlace(buildPlace());
+
+			expect(locn.streetNumber).toBeUndefined();
+			expect(locn.streetName).toBeUndefined();
+			expect(locn.locality).toBeUndefined();
+			expect(locn.countryCode).toBeUndefined();
+			expect((locn as Record<string, unknown>).displayName).toBeUndefined();
+			expect(locn.adminOneCode).toBeUndefined();
+			expect(locn.adminTwoName).toBeUndefined();
+			expect(locn.adminTwoCode).toBeUndefined();
+		});
+
+		it('sets sublocation to null but keeps the name when only a name is available', () => {
+			const locn = component.createLocnFromPlace(
+				buildPlace({}, { street_number: undefined, route: undefined }),
+			);
+
+			expect(locn.sublocation).toBeNull();
+			expect(locn.name).toBe("Jean Valjean's House");
+		});
+
+		it('falls back to streetName alone when streetNumber is absent', () => {
+			const locn = component.createLocnFromPlace(
+				buildPlace({}, { street_number: undefined }),
+			);
+
+			expect(locn.sublocation).toBe('Rue Plumet');
+		});
+
+		it('writes name when the place name does not include the sublocation', () => {
+			const locn = component.createLocnFromPlace(buildPlace());
+
+			expect(locn.name).toBe("Jean Valjean's House");
+		});
+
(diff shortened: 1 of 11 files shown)


src/app/file-browser/components/location-picker/location-picker.component.ts
```diff
@@ -229,36 +229,32 @@ export class LocationPickerComponent implements OnInit, AfterViewInit {
...
 		function getComponentName(
 			addressComponents: google.maps.GeocoderAddressComponent[],
 			type,
-			getShortName = true,
 		) {
 			const component = find(addressComponents, (c) => c.types.includes(type));
-			return component
-				? component[getShortName ? 'short_name' : 'long_name']
-				: null;
+			return component ? component.long_name : null;
```
cecilia-donnelly, line 264: What was the short name?

src/app/file-browser/components/location-picker/location-picker.component.spec.ts
```diff
@@ -1,25 +1,158 @@
...
+		postal_code: { long_name: '75007' },
+		administrative_area_level_1: { long_name: 'Ile-de-France' },
+		country: { long_name: 'France' },
+	};
+	const merged = { ...defaults, ...overrides };
+	return Object.entries(merged)
+		.filter(([, value]) => value !== undefined)
+		.map(([type, value]) => ({
+			short_name: value.long_name,
+			long_name: value.long_name,
```
cecilia-donnelly, line 45: Probably this is the same answer as my other name confusion, but what's going on with short and long?

src/app/file-browser/components/location-picker/location-picker.component.spec.ts
```diff
@@ -1,25 +1,158 @@
...
+		name: "Jean Valjean's House",
+		address_components: buildAddressComponents(addressOverrides),
+		geometry: {
+			location: {
+				lat: () => 48.83,
+				lng: () => 2.3,
+			},
+		},
+		...overrides,
+	}) as unknown as google.maps.places.PlaceResult;
```
cecilia-donnelly, line 67: Why does it have to be `as unknown`?

src/app/file-browser/components/location-picker/location-picker.component.spec.ts
```diff
@@ -1,25 +1,158 @@
...
+			expect(locn.postalCode).toBe('75007');
+			expect(locn.latitude).toBe(48.83);
+			expect(locn.longitude).toBe(2.3);
+		});
+
+		it('does not write deprecated legacy fields', () => {
+			const locn = component.createLocnFromPlace(buildPlace());
+
+			expect(locn.streetNumber).toBeUndefined();
+			expect(locn.streetName).toBeUndefined();
```
cecilia-donnelly, line 124: The requirements say that we should use street addresses even though IPTC doesn't, so I wouldn't expect these to be in a list of "deprecated legacy fields."

src/app/file-browser/components/location-picker/location-picker.component.ts
```diff
@@ -229,36 +229,32 @@ export class LocationPickerComponent implements OnInit, AfterViewInit {
 
 	createLocnFromPlace(place: google.maps.places.PlaceResult) {
 		const addr = place.address_components;
+		const streetNumber = getComponentName(addr, 'street_number');
+		const streetName = getComponentName(addr, 'route');
+		const sublocation =
+			[streetNumber, streetName].filter(Boolean).join(' ') || null;
```
cecilia-donnelly, line 241: (TIL)

src/app/file-browser/components/location-picker/location-picker.component.spec.ts
cecilia-donnelly: Okay, reading the document more carefully I see that we expose the "sublocation" field (in the "UI Requirements") instead of separate street number and name fields, which is why these are no longer relevant.

cecilia-donnelly (commented):
I don't see anything obviously off here! I have some questions, likely answered by your partner PRs to this one, but wanted to ask them right away.

src/app/file-browser/components/location-picker/location-picker.component.ts
cecilia-donnelly: @slifty is going to figure out if we want to keep that after all.

src/app/file-browser/components/location-picker/location-picker.component.spec.ts
cecilia-donnelly: Because this is a test and we're just pretending it's a `PlaceResult` and don't care if it's exactly right.

src/app/file-browser/components/location-picker/location-picker.component.ts
slifty: The idea is that short vs long name isn't part of the new location specification, and so we just use the long name now because it's the only thing that will exist in the future.

Short name might have been "CA" when long name was "California"

src/app/file-browser/components/location-picker/location-picker.component.spec.ts
slifty: Fixing this up (the idea that led to this mistake in we don't use short name but need to populate it... but we should just populate it!  Sorry I missed this!)

src/app/file-browser/components/location-picker/location-picker.component.spec.ts
slifty: I mentioned this by voice but this is to allow our partial-mock to validate type safety within the test.  In actual code I wouldn't do this but for a test I think it's "ok" since the alternative is a lot of unused mocking just to satisfy tsc.

src/app/shared/services/api/record.repo.ts
```diff
@@ -157,14 +158,19 @@ export const convertStelaSharetoShareVO = (stelaShare: StelaShare): ShareVO =>
...
+): LocnVOData | null => {
+	if (!stelaLocation?.id) {
+		return null;
+	}
+	const { state, precision, ...rest } = stelaLocation;
+	return {
+		...rest,
+		locnId: Number.parseInt(stelaLocation.id, 10),
+		adminOneName: state ?? undefined,
+		locationPrecision: precision ?? undefined,
```
aasandei-vsp, line 189: These checks don't really make sense.

src/app/shared/services/api/record.repo.ts
```diff
@@ -71,16 +72,16 @@ interface StelaFile {
...
+	name?: string | null;
+	sublocation?: string | null;
+	city?: string | null;
+	state?: string | null;
+	postalCode?: string | null;
+	country?: string | null;
+	latitude?: number | null;
+	longitude?: number | null;
+	altitudeMeters?: number | null;
+	precision?: LocationPrecision | null;
```
aasandei-vsp, line 84: Why do we need them to be also null? they can just be skipped, as they are optional.

aasandei-vsp (approved):
LGTM, small observations.

src/app/shared/services/api/record.repo.ts
slifty: `null` is a value and `undefined` is a lack of any value.

```javascript
const foo = {
a: null,
b: undefined,
}

console.log(foo.a);
console.log(foo.b);
console.log(foo.c);
```

results in

```
null
undefined
undefined
```

---

`null` and `undefined` are both falsey, but in any case where stela could return `null` or provide no value at all we'd want the types to reflect both possibilities.

I do personally believe that it is better to never have optional properties in an API's entity shape (outside of `patch`), and instead have our API commit to there being a nullable value because of this very ambiguity!

All that said: I'll double check the stela API commitments.  If this can be null we should have null, and if not we shouldn't!

src/app/shared/services/api/record.repo.ts
slifty: It's because precision can be null but `locationPrecision` is not nullable.

These lines are mapping null to undefined (so if precision is null it would be translated to `undefined`, which is allowed)

src/app/shared/services/api/record.repo.ts
slifty: OK!  So currently the stela spec lists these as optional but not nullable; ironically the *implementation* is the inverse (stela can return null, and always returns something).

Given that I think we should keep this as written.  Null is a possible value, and in theory stela could choose to not return a given value.

slifty: Created a stela ticket about this: https://github.com/PermanentOrg/stela/issues/742

slifty:
In my own testing: I'd like to wait until the stela endpoints for location creation are live before moving the frontend to use the new properties.

omnignorant (dismissed):
Not 100% sure what I am testing or what the success condition should be, but I did my best to suss it out of the QA guide. The gist seems to be "does a location properly save using the current picker UI"

Location editing seems to work inconsistently. In the screen capture attached here, I seem to have determined the following:
- location seems to save on initial selection
- for records without existing location data, the changed metadata persists after a refresh;
- for records with existing location data, the metadata does not persist;
- location metadata in profile, both general and milestone persists.

https://github.com/user-attachments/assets/f59c0479-70d1-4464-b7ef-8214b0aee6b9

slifty:
@omnignorant sorry for the confusion -- I pulled this off of the QA pile since we need to do some additional changes for everything to work as expected!

slifty:
Before we merge this I want to re-check to ensure that records with legacy location metadata work as expected.

src/app/shared/services/api/record.repo.ts
aasandei-vsp: Makes sense and we can leave it at that. In my opinion, we should go with one or the other to better enforce the contract we're making with the BE, either always having a value or agree that it's optional so sometimes it might be undefined.

src/app/shared/services/api/record.repo.ts
```diff
@@ -157,14 +158,69 @@ export const convertStelaSharetoShareVO = (stelaShare: StelaShare): ShareVO =>
...
+	sublocation?: string;
+	city?: string;
+	state?: string;
+	postalCode?: string;
+	country?: string;
+	latitude?: number;
+	longitude?: number;
+	altitudeMeters?: number;
+	precision?: LocationPrecision;
+}
```
aasandei-vsp: Shouldn't these also be nullable? Maybe I'm not understanding this correctly, but I thought we are updating the same fields that we get with StelaLocation.

aasandei-vsp (approved):
Left a few comments, but overall looks good.

Checked all location instances that I know of(manual testing):

- Sidebar
- Archive settings --> Milestones --> Location
- Group info --> Location established
- Gallery --> current archive --> info

src/app/shared/services/api/record.repo.ts
slifty: Oh great catch yes I believe you are right.

slifty: Belated update(HA!)

This is the shape of the *Request*, rather than what stela can give back.  stela documentation doesn't allow null, so this is correct as written (even if there's an argument for stela accepting null in future)

cecilia-donnelly (commented):
QA findings: mostly this looks good!

An actual bug reproduction recipe:
1. Choose a record with no location
2. Set the location to an actual street address (I used 1600 Pennsylvania Avenue South, Minneapolis, MN, USA) and save.
3. I refreshed; this might not be critical
4. Change the location to one without a street address. I used Colorado National Monument, Grand Junction, Colorado.
5. Refresh (this one is definitely necessary)
6. After refresh, observe that the street number and name are now the previous values, so in my example I see "1600 Pennsylvania Avenue South
Grand Junction, Colorado"

Since location updates are broken altogether on main (PER-10630) this is definitely an improvement, but if we can fix it that would be ideal.

Probably related but not critical: If I search for Acadia National Park, the location that appears in the editing view (sidebar and full) is "ME, US, US"

I followed these tests, for posterity:

```
## Test 1 — Set a location on a file for the first time

1. Open the file browser and click on a single file that does not currently have a location set (the sidebar on the right should say "No location" or "Click to set location" next to "Location").
2. Click the "Location" row in the sidebar.
3. In the map dialog that opens, type a real address into the search box (e.g. your office address or a well-known landmark) and select it from the suggestions.
4. Confirm a pin drops on the map at the correct spot and a short address preview appears.
5. Click "Save".

**Expected result:** The dialog closes, and the sidebar now shows the address you picked next to "Location," along with a small map thumbnail.

---

## Test 2 — Set a location on a folder for the first time

1. Repeat Test 1, but select a folder instead of a file.

**Expected result:** Same as Test 1 — the folder's sidebar shows the new location and map thumbnail.

---

## Test 3 — Change an existing location

1. Select a file or folder that already has a location set (from Test 1 or 2).
2. Click the "Location" row again.
3. Search for and select a different address.
4. Click "Save".

**Expected result:** The sidebar updates to show the new address, not the old one.

---

## Test 4 — View a location in the full file details page

1. Double-click a file that has a location set to open its full details page.
2. Find the "Location" row in the details table.

**Expected result:** The same address and map thumbnail shown in the sidebar also appear here. Clicking it opens the same map dialog for editing.

---

## Test 5 — Cancel without saving

1. Select a file or folder with a known location already set.
2. Click "Location" to open the map dialog.
3. Search for and select a different address, but click "Cancel" instead of "Save."

**Expected result:** The dialog closes and the original location is unchanged — the new address you searched for was not saved.

---

## Test 6 — Set a location with limited address information

1. Select a file or folder without a location.
2. Open the location picker and search for a broad place name that doesn't have a specific street address (e.g. a city name or a large park/landmark rather than a full street address).
3. Select it and click "Save".

**Expected result:** A reasonable location still displays afterward (e.g. city and country, or coordinates) — it should not show blank, "undefined," or an error.

---

## Test 7 — Location persists after refresh

1. Set a location on a file (per Test 1) or use one already set.
2. Refresh the browser page, or navigate away to a different folder and then back.

**Expected result:** The location you set is still shown correctly — nothing reverts or disappears.

---

## Test 8 — No location set

1. Find or create a file/folder that has never had a location set.

**Expected result:** The sidebar and full details page clearly show "No location" or "Click to set location" (not blank, not an error, not leftover data from another item).

```

slifty:
Apparently the way I was trying to update locations didn't align with the backend -- fixed that up so now we're properly creating a new location instead of trying to update an existing one (which in turn was triggering some location reprocessing in the PHP side).

I also made two other little changes:

1. Location name is now rendered in the detail view in a way that matches the map view (it's bolded / treated as a title).  This is because sometimes the name is very similar to the sublocation and this helps convey the semantic role rather than looking like a duplication bug

2. I added some belt-and-suspenders support for legacy data since we won't be re-processing past location data for some time (once we do we'll consistently only have the new metadata fields, but for now we need the front end to gracefully fall back to the old values if new ones aren't present).

cecilia-donnelly (approved):
This looks great! Thanks for addressing my earlier comments, @slifty .

PR #1012, opened 2026-05-06
[PER-10590] Remove the drag while scrolling

aasandei-vsp:
While scrolling, there was a drag for the whole app. Removing the "mobile like" behavior that can be added automatically by css libraries, like bootstrap, the scroll won't be affected anymore and will work as expected on web.

### DIFF
--- a/src/styles/_base.scss
+++ b/src/styles/_base.scss
@@ -4,6 +4,7 @@ body {
 	height: 100%;
 	position: relative;
 	background: none;
+	overscroll-behavior: none;
 }
 
 pr-app-root {


cecilia-donnelly (approved):
I am not sure how to trigger the undesired behavior, so I couldn't verify that this stops it, but it makes perfect sense.

sergiu-tnv:
QA Approved

sergiu-tnv (approved):
QA Approved

omnignorant (approved):
works perfectly!

PR #1019, opened 2026-05-11
[PER-10593] Reset page scroll on public gallery navigation

aasandei-vsp:
When navigating into a sub-folder in a public archive, the page kept the parent view's scroll position so users landed at the bottom of the new sub-folder rather than the top. Reset window scroll to (0, 0) on every NavigationEnd inside PublicComponent so each folder load starts at the top.

Issue: PER-10593

### DIFF
--- a/src/app/public/components/public/public.component.ts
+++ b/src/app/public/components/public/public.component.ts
@@ -54,6 +54,7 @@ export class PublicComponent implements OnInit, OnDestroy {
 					this.isNavigating = true;
 				} else if (event instanceof NavigationEnd) {
 					this.isNavigating = false;
+					window.scrollTo(0, 0);
 				}
 			});
 	}


cecilia-donnelly (commented):
This didn't work for me. I went to the public view of an archive with a lot of content (locally). I scrolled to the bottom of the front page (local equivalent of https://app.permanent.org/p/archive/03dg-0000) and entered a folder. Then I was still at the bottom of the page inside the folder. I rebuilt the web-app from docker, but maybe there's something else I should do to make sure this is getting picked up?

aasandei-vsp:
@cecilia-donnelly Hmmm, I deployed it on dev as well and it worked for me, but maybe I fixed it in a different place than the one you/the ticket expects it.

The steps I followed:

1. In the private workspace, create a folder with a lot of content and some subfolders
2. Single click(select) the folder
3. Select the Publish action from the top of the menu
4. Click the publish button from the modal
5. Click the "View on web" button from the new modal or copy the generated link
6. The folder should load, so scroll to the bottom of the page and click a folder
EXPECTED: The subfolder will load and the page will scroll to the top.

Please let me know if this is the right place for the fix or if there are different steps that I need to follow.

cecilia-donnelly (approved):
Worked for me on dev so I probably had some problem with local not picking up the changes correctly. 🤷‍♀️ Sorry for the delay!

omnignorant (commented):
So, this is mostly working but maybe not on Firefox? But it was mostly working on Firefox. 

Here's how I broke it on Firefox:
1. Nav here: https://app.dev.permanent.org/p/archive/065t-0000/065t-00uj/325929
2. scroll to bottom click on "XXX Chickens Level 2" folder (the XXX is just to force it to the bottom of the page 🤪 )
3. I observe no scroll up

aasandei-vsp:
@omnignorant Hmm, that is really strange. I'm sorry, but I didn't manage to reproduce it. I have tried the exact scenario in Firefox and it worked for me. I used the build on dev that's already using this branch. This is a video of what I have seen:
https://www.loom.com/share/69e81c69602948ffaa2e8e250d58a2e4

slifty:
I can confirm this is working in firefox for me at this point.

slifty (approved):

PR #1020, opened 2026-05-11
Fix unlisted-share preview flicker

aasandei-vsp:
Resolve isUnlistedShare and ephemeralFolder before SharePreviewComponent renders so the first paint matches the share's actual type, instead of defaulting to the restricted-share UI and swapping after an async API call.

- Add primeForToken / isUnlistedShareSync to ShareLinksService and invalidate the cache on token change and on navigation off /share/
- Have ShareUrlResolveService prime the share-links cache in parallel with checkShareLink via Promise.all
- Add ShareUrlEphemeralFolderResolveService so ephemeralFolder is on route data before the component constructs
- Read isUnlistedShare synchronously in SharePreviewComponent and FileListItemComponent so blurred / real thumbnails are correct on first paint

Issue: PER-10352


aasandei-vsp:
This will actually not fix the root cause, there are actually 3 important problems: component rendering, routing and animation that would need a significant architectural change.

PR #1025, opened 2026-05-15
Ignore location `displayName`

### LINKED ISSUE #1024: Do not render the `displayName` of a location
The display name of a location is not actually stored by the PHP backend, which makes it an ephemeral artifact.  Currently it is populated by the `name` of the location object returned by the location lookup service (i.e. google maps).

This is confusing since it will appear initially but disappear on refresh.

We should just ignore it / not populate it / not render it in our UX.

slifty:
This PR removes recognition of the `displayName` attribute of a location.  This is a legacy field we are about to remove, but also it wasn't actually being stored which resulted in unexpected behavior (specifically, the value disappeared on refresh for users)

Resolves #1024

### DIFF
--- a/src/app/file-browser/components/location-picker/location-picker.component.html
+++ b/src/app/file-browser/components/location-picker/location-picker.component.html
@@ -15,10 +15,6 @@
 			@if (currentLocation) {
 				<div class="location-panel" @ngIfFadeInAnimation>
 					<div class="location-info">
-						@if (currentLocationDisplay.displayName) {
-							<span>{{ currentLocationDisplay.displayName }}</span>
-							<br />
-						}
 						<span>{{ currentLocationDisplay.line1 }}</span>
 						<br />
 						<span>{{ currentLocationDisplay.line2 }}</span>
--- a/src/app/file-browser/components/location-picker/location-picker.component.ts
+++ b/src/app/file-browser/components/location-picker/location-picker.component.ts
@@ -244,10 +244,6 @@ export class LocationPickerComponent implements OnInit, AfterViewInit {
 			countryCode: getComponentName(addr, 'country', true),
 		};
 
-		if (!place.name.includes(locn.streetNumber)) {
-			locn.displayName = place.name;
-		}
-
 		return locn;
 
 		function getComponentName(
--- a/src/app/models/locn-vo.ts
+++ b/src/app/models/locn-vo.ts
@@ -4,7 +4,6 @@ export interface LocnVOData extends BaseVOData {
 	locnId?: number;
 	timeZoneId?: number;
 
-	displayName?: string;
 	geoCodeLookup?: string;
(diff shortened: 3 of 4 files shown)


aasandei-vsp (commented):
Code wise, the change seems correct. But I have some issues understanding what has changed and where in the UI.

- Where exactly did the displayName appear in the UI?
- Can you give me some steps to reproduce before and after?

**This is the research I have done:**

**A. Location on sidebar(preview and full) On main and also on the branch I had the exact same result. Steps I tried:**
1. Select a file
2. Click on "Click to set location"
3. In the search bar, type Breazu.
4. Select "Breazu, Romania, RO"
5. Click save
EXPECTED: The map updates the pin, the location line says "Breazu, Romania, RO"
ACTUAL: As expected.
**?????? Not sure where the display name should be present, taking into account that I saw the same thing on both main and this branch.**
6. Refresh the page.
EXPECTED: The location should show "Breazu, Romania, RO" and the pin on the map should be correct.
ACTUAL: As expected.
7. Follow steps 2-6 again, but this time type "Iasi" in the search bar
EXPECTED: The location should show "Iași, IS, RO, RO" and the pin on the map should be correct.
ACTUAL: The location shows "Breazu, Romania, RO" and the pin on the map is for this location.

**B. Location pipe**
I have seen it used in the code in 2 places, on the public profile and somewhere in shared, but I cannot get to them in the UI. Could you help me with some steps for that, please?

slifty:
@aasandei-vsp I believe this only applies to *named locations* (e.g. landmarks such as "Eiffel Tower").  The landmark name would appear pre-save:

image

but then disappear on refresh:

image

Within this branch the landmark name is not rendered at all, just the address.

As for the location pipe, I'll look into it and try to get you instructions there.

aasandei-vsp:
@slifty Ah, thank you! And yes, this specific case works as expected, both in the sidebar and profile.

I've also figured out the location (there are actually 2 places that I've found) on the profile where you can edit:
Milestones location:
https://www.loom.com/share/837624ecab33442ba62758e0804b1264
Location Established from group information:
https://www.loom.com/share/5c49cb82c0ef48209f65acf6a0dc8c62

But also where these edits appear:
https://www.loom.com/share/9f8292ce708e46df943e8d7de54361fd


The only thing remaining is the fact that in the sidebar, after the first time you set the location, you cannot update it, but I am not sure it's relevant to this task.

I'll approve this PR, because the bug mentioned above is irelevant in this context, but we need to create a bug for it and maybe fix it before we start the UI work on location.

aasandei-vsp (approved):

slifty:
@cecilia-donnelly is going to give this a QA pass!

cecilia-donnelly (approved):
I checked in the sidebar and in the profile. Thanks!

PR #1029, opened 2026-05-27
[PER-10632] Add EDTF datepicker, timepicker, and timezone-dropdown components

aasandei-vsp:
These changes do not reflect in the UI at all, so no need for QA. Maybe just to check the app is loading.

The commits about 24H and unspecified digits should be checked thouroughly, the first main commit has already been checked on PR [#924](https://github.com/PermanentOrg/web-app/pull/924)

Issue: PER-10632

### DIFF
--- a/src/app/shared/components/datepicker-input/datepicker-input.component.ts
+++ b/src/app/shared/components/datepicker-input/datepicker-input.component.ts
@@ -85,11 +85,9 @@ export class DatepickerInputComponent implements OnInit, OnChanges {
 			return;
 		}
 
+		this.dateChange.emit({ ...this.date, year: value });
 		if (value.length === 4) {
-			this.dateChange.emit({ ...this.date, year: value });
 			this.monthInput.nativeElement.focus();
-		} else if (value.length === 0) {
-			this.dateChange.emit({ ...this.date, year: value });
 		}
 	}
 
--- a/src/app/shared/components/datepicker-input/datepicker-input.component.ts
+++ b/src/app/shared/components/datepicker-input/datepicker-input.component.ts
@@ -1,0 +1,156 @@
+import {
+	Component,
+	Input,
+	Output,
+	EventEmitter,
+	signal,
+	HostListener,
+	ElementRef,
+	ViewChild,
+	OnChanges,
+	SimpleChanges,
+	OnInit,
+} from '@angular/core';
+import { CommonModule } from '@angular/common';
+import { NgbDatepicker, NgbDateStruct } from '@ng-bootstrap/ng-bootstrap';
+import {
+	DateModel,
+	EdtfService,
+} from '@shared/services/edtf-service/edtf.service';
+
+@Component({
+	selector: 'pr-datepicker-input',
+	standalone: true,
+	imports: [CommonModule, NgbDatepicker],
+	templateUrl: './datepicker-input.component.html',
+	styleUrls: ['./datepicker-input.component.scss'],
+})
+export class DatepickerInputComponent implements OnInit, OnChanges {
+	@Input() date: DateModel = { year: '', month: '', day: '' };
+	@Input() disabled = false;
+
+	@Output() dateChange = new EventEmitter<DateModel>();
+
+	@ViewChild('yearInput') yearInput!: ElementRef<HTMLInputElement>;
+	@ViewChild('monthInput') monthInput!: ElementRef<HTMLInputElement>;
+	@ViewChild('dayInput') dayInput!: ElementRef<HTMLInputElement>;
+
+	showDatepicker = signal(false);
+	datepickerModel = signal<NgbDateStruct | null>(null);
+
+	constructor(
+		private elementRef: ElementRef,
+		private edtfService: EdtfService,
+	) {}
+
+	ngOnInit(): void {
+		this.updateDatepickerModel(this.date);
+	}
+
+	ngOnChanges(changes: SimpleChanges): void {
+		if (changes.date) {
+			this.updateDatepickerModel(this.date);
+		}
+	}
+
+	private updateDatepickerModel(date: DateModel): void {
+		const year = parseInt(date.year, 10);
+		const month = parseInt(date.month || '', 10);
+		const day = parseInt(date.day || '', 10);
+		if (year && month && day) {
+			this.datepickerModel.set({ year, month, day });
+		} else {
+			this.datepickerModel.set(null);
+		}
+	}
+
+	@HostListener('document:click', ['$event'])
+	onDocumentClick(event: MouseEvent): void {
+		if (!this.elementRef.nativeElement.contains(event.target)) {
+			this.showDatepicker.set(false);
+		}
+	}
+
+	toggleDatepicker(): void {
+		if (this.disabled) return;
+		this.showDatepicker.update((v) => !v);
+	}
+
+	updateYear(event: Event): void {
+		const input = event.target as HTMLInputElement;
+		const value = input.value;
+
+		if (!this.edtfService.isValidYear(value)) {
+			input.value = this.date.year;
+			return;
+		}
+
+		if (value.length === 4) {
+			this.dateChange.emit({ ...this.date, year: value });
+			this.monthInput.nativeElement.focus();
+		} else if (value.length === 0) {
+			this.dateChange.emit({ ...this.date, year: value });
+		}
+	}
+
+	updateMonth(event: Event): void {
+		const input = event.target as HTMLInputElement;
+		const value = input.value;
+
+		if (!this.edtfService.isValidMonth(value)) {
+			input.value = this.date.month ?? '';
+			return;
+		}
+
+		this.dateChange.emit({ ...this.date, month: value });
+		if (value.length === 2) {
+			this.dayInput.nativeElement.focus();
+		}
+	}
+
+	updateDay(event: Event): void {
+		const input = event.target as HTMLInputElement;
+		const value = input.value;
+
+		if (
+			!this.edtfService.isValidDay(value, this.date.year, this.date.month ?? '')
+		) {
+			input.value = this.date.day ?? '';
+			return;
+		}
+
+		this.dateChange.emit({ ...this.date, day: value });
+	}
+
+	onMonthKeydown(event: KeyboardEvent): void {
+		if (event.key !== 'Backspace') return;
+		const target = event.target as HTMLInputElement;
+		if (target.value !== '') return;
+		event.preventDefault();
+		const newYear = (this.date.year ?? '').slice(0, -1);
+		this.dateChange.emit({ ...this.date, year: newYear });
+		this.yearInput.nativeElement.focus();
+	}
+
+	onDayKeydown(event: KeyboardEvent): void {
+		if (event.key !== 'Backspace') return;
+		const target = event.target as HTMLInputElement;
+		if (target.value !== '') return;
+		event.preventDefault();
+		const newMonth = (this.date.month ?? '').slice(0, -1);
+		this.dateChange.emit({ ...this.date, month: newMonth });
+		this.monthInput.nativeElement.focus();
+	}
+
+	onDateSelect(newDate: NgbDateStruct): void {
+		this.datepickerModel.set(newDate);
+		const updatedDate = {
+			year: String(newDate.year),
+			month: String(newDate.month).padStart(2, '0'),
+			day: String(newDate.day).padStart(2, '0'),
+		};
+		this.date = updatedDate;
+		this.dateChange.emit(updatedDate);
+		this.showDatepicker.set(false);
+	}
+}
--- a/src/app/shared/services/edtf-service/edtf.service.spec.ts
+++ b/src/app/shared/services/edtf-service/edtf.service.spec.ts
@@ -77,6 +77,43 @@ describe('EdtfService', () => {
 
 				expect(result.qualifiers.unknown).toBe(false);
 			});
+
+			it('should strip trailing X from partial year', () => {
+				const result = service.toDateTimeModel('198X');
+
+				expect(result.date.year).toBe('198');
+				expect(result.date.month).toBe('');
+			});
+
+			it('should strip trailing X from partial month', () => {
+				const result = service.toDateTimeModel('1985-1X');
+
+				expect(result.date.year).toBe('1985');
+				expect(result.date.month).toBe('1');
+			});
+
+			it('should strip trailing X from partial day', () => {
+				const result = service.toDateTimeModel('1985-05-2X');
+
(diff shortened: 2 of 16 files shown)


cecilia-donnelly (commented):
I asked Claude to review so I could get you something today! It found a few things in the EDTF service (as you suspected). My comments in italics where I've quoted from the results.

1. It had a concern about existing data like `19XX` being transformed into `0019`. I don't think we have any existing data in an EDTF format in the database, since we haven't exposed that in the UI, so I'm not overly worried, but might be worth double checking that a date with a partially unknown year is saving / updating correctly.
2. "Clearing the year field while a month is already set silently produces '0000-MM' instead of treating the date as invalid." -- _but I think this is the expected behavior. I would expect the user to need to clear fields independently._
3. "isValidHour('0') returns true (0 ≤ 1 single-digit guard), so typing '0' in the hours field passes validation, but buildTimeString later throws when serializing it.", _and more detail:_ 
```
"User types '0' → isValidHour('0') = true → timeChange emits {hours:'0'}; when the form saves, buildTimeString is called with hours:'0' (truthy, no early-exit); parseTimeAs24Hour feeds '0' into date-fns parse('0:m:s a', 'h:m:s a') where the h token 
  requires 1–12 → isValid returns false → parseTimeAs24Hour returns null → buildTimeString throws 'Invalid time' → save fails."
```
4. "When showSeconds is false, onTimeSelect emits the stale this.time.seconds value instead of clearing it, causing stale seconds to persist silently in the EDTF output."
5. "normalizeEdtfString appends EDTF qualifiers after the time component (producing '1985-05-20T14:30:45~'), which is not a valid EDTF production; the edtf() validation call in toEdtfDate would then throw, making it impossible to save any record that has both a time 
  and an approximate/uncertain qualifier." _I don't think we'd ever get here, since we don't let users indicate time to the second *and* approximate/uncertain_
6. _And one non-EDTF service possible bug_
```
 "summary": "toggleAmPm cross-assigns am and pm by value, so if either optional field is absent the emitted model has undefined for one meridian, permanently corrupting am/pm on the next toggle.",
      "failure_scenario": "A consumer passes TimeModel {hours:'10', minutes:'30', pm:true} omitting the optional am field → toggleAmPm emits {am:true, pm:undefined}; a subsequent toggle emits {am:undefined, pm:true}; any code that reads time.am (e.g. parseTimeAs24Hour, which 
  defaults to AM when pm!==true) will permanently misread midnight/noon for this record."
```

I can definitely validate these, but it's the end of the day here and I wanted to give you some feedback for your Wednesday! I hope they aren't rabbit holes / definitely treat them with suspicion.

aasandei-vsp:
> I asked Claude to review so I could get you something today! It found a few things in the EDTF service (as you suspected). My comments in italics where I've quoted from the results.
> 
> 1. It had a concern about existing data like `19XX` being transformed into `0019`. I don't think we have any existing data in an EDTF format in the database, since we haven't exposed that in the UI, so I'm not overly worried, but might be worth double checking that a date with a partially unknown year is saving / updating correctly.
> 2. "Clearing the year field while a month is already set silently produces '0000-MM' instead of treating the date as invalid." -- _but I think this is the expected behavior. I would expect the user to need to clear fields independently._
> 3. "isValidHour('0') returns true (0 ≤ 1 single-digit guard), so typing '0' in the hours field passes validation, but buildTimeString later throws when serializing it.", _and more detail:_
> 
> ```
> "User types '0' → isValidHour('0') = true → timeChange emits {hours:'0'}; when the form saves, buildTimeString is called with hours:'0' (truthy, no early-exit); parseTimeAs24Hour feeds '0' into date-fns parse('0:m:s a', 'h:m:s a') where the h token 
>   requires 1–12 → isValid returns false → parseTimeAs24Hour returns null → buildTimeString throws 'Invalid time' → save fails."
> ```
> 
> 4. "When showSeconds is false, onTimeSelect emits the stale this.time.seconds value instead of clearing it, causing stale seconds to persist silently in the EDTF output."
> 5. "normalizeEdtfString appends EDTF qualifiers after the time component (producing '1985-05-20T14:30:45~'), which is not a valid EDTF production; the edtf() validation call in toEdtfDate would then throw, making it impossible to save any record that has both a time
>    and an approximate/uncertain qualifier." _I don't think we'd ever get here, since we don't let users indicate time to the second _and_ approximate/uncertain_
> 6. _And one non-EDTF service possible bug_
> 
> ```
>  "summary": "toggleAmPm cross-assigns am and pm by value, so if either optional field is absent the emitted model has undefined for one meridian, permanently corrupting am/pm on the next toggle.",
>       "failure_scenario": "A consumer passes TimeModel {hours:'10', minutes:'30', pm:true} omitting the optional am field → toggleAmPm emits {am:true, pm:undefined}; a subsequent toggle emits {am:undefined, pm:true}; any code that reads time.am (e.g. parseTimeAs24Hour, which 
>   defaults to AM when pm!==true) will permanently misread midnight/noon for this record."
> ```
> 
> I can definitely validate these, but it's the end of the day here and I wanted to give you some feedback for your Wednesday! I hope they aren't rabbit holes / definitely treat them with suspicion.

1. In the document "2026-01-21 Dates Times and Locations Metadata", we have a paragraph specifying that we only validate standard dates and times. From that, I understood that we do not support any X-es in the date or time: "USE STANDARD DATE/TIME VALIDATION, e.g. 4 digit year positive, 1-12 months, 1-31 days, 12/24 hours, etc.". So this is not a concern atm.

2. It is expected behavior, we can change it upon request.

3. It is expected behavior, I chose to implement it this way, because atm we do not actually "validate" the date and time, we just filter which characters we are allowed to fill in. Starting an hour with 0 is allowed, because 01 should be valid, so the user will first input 0. The general validation does fail because as a whole, the edtf date is invalid, but up until the user figures out what hour it needs to fill in.

4. The default value of that is true and we do not actually set it anywhere. We made it into an input because ngb-timepicker gets it as an input property. I'll just remove it and add it if we ever need to hide seconds.

5. Nowhere in the "Extended Date/Time Format (EDTF) Specification" document is mentioned that a date and time can have a qualifier. In all places where qualifiers are mentioned(level 1 and level 2), it is always only about date. Maybe I am missing something, so if there is a different interpretation of the document, please let me know.

6. The scenario is not correct, there is no bug, am and pm will be set correctly even if we send undefined. But indeed, the code is inconsistent, so we're going to coerce to true/false, so no undefined is sent.

I think I have answered all the questions, let me know if there is anything else that I need to address.

cecilia-donnelly:
Thank you for addressing all of these! I see that Robert added the note about only validating standard dates in that document (for #1) but that's very surprising to me since right along we've been talking about the X's as a very useful part of the EDTF standard. Let me clarify that with him.

aasandei-vsp:
> Thank you for addressing all of these! I see that Robert added the note about only validating standard dates in that document (for #1) but that's very surprising to me since right along we've been talking about the X's as a very useful part of the EDTF standard. Let me clarify that with him.

@cecilia-donnelly I know there has been a lot of talk about what we keep and what we don't, but what was put in writing, for me, it meant we are not supporting unspecified digits. The only place where unspecified digits are mentioned are actually in the unknown date description.

src/app/shared/services/edtf-service/edtf.service.ts
```diff
@@ -43,181 +64,168 @@ export interface DateTimeModel {
...
-		return timezone ? `${edtfString}${timezone}` : edtfString;
+		return result;
 	}
 
 	private buildDateString(date: DateModel): string {
-		const year = date.year.padEnd(4, 'X');
+		const year = date.year.padStart(4, '0');
 
 		if (!date.month) {
 			return year;
```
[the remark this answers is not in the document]
aasandei-vsp: This is replaced by unspecified digits.

src/app/shared/components/datepicker-input/datepicker-input.component.ts
```diff
@@ -0,0 +1,156 @@
...
+			input.value = this.date.year;
+			return;
+		}
+
+		if (value.length === 4) {
+			this.dateChange.emit({ ...this.date, year: value });
+			this.monthInput.nativeElement.focus();
+		} else if (value.length === 0) {
+			this.dateChange.emit({ ...this.date, year: value });
+		}
```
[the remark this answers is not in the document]
aasandei-vsp: This is irrelevant because the functionality was replaced by unspecified digits.

src/app/shared/services/edtf-service/edtf.service.ts
```diff
@@ -43,181 +64,173 @@ export interface DateTimeModel {
...
 
-	isValidDate(date: DateModel): boolean {
-		if (!date.year) {
-			return false;
-		}
+	private normalizeForParsing(edtfString: string): string {
+		// Replace +hh:mm / -hh:mm timezone offset with Z so the edtf library
+		// can parse it — the library requires a UTC designator and throws on offsets.
+		return edtfString.replace(/T([\d:]+)[+-]\d{2}:\d{2}/, 'T$1Z');
+	}
```
cecilia-donnelly, line 129: Okay, I think this is because we decided not to handle timezones in this first iteration. No change *necessarily* needed here, since this is for display as far as I can see, but the comment seems off and should be adjusted to explain that this is to preserve the local time for display.

On write we don't want to lose the offset information since that can go into the `displayTime` field (without needing a separate field that we had talked about for the specific timezone). The library has an `offsetToTimeZone` function, and the spec requires us to handle offsets, so we need to save them somehow. It looks like the offset handling was removed from `buildTimeString` and needs to go back in there so we don't lose the offset when we save a time.

When I suggested not handling timezones I was thinking about display, and I like that this always shows up as local time for now! But I don't want to strip out the offset on *save*.

src/app/shared/services/edtf-service/edtf.service.ts
```diff
@@ -249,54 +270,148 @@ export class EdtfService {
...
+	isValidDay(value: string, year: string, month: string): boolean {
+		if (value === '') return true;
+		// Defaults let day be typed before year/month are filled in.
+		// 2000 is a leap year (allows Feb 29); 01 has 31 days (most permissive).
+		const yearStr = year.length === 4 ? year : '2000';
+		const monthStr = month.length === 2 ? month : '01';
+		const dayStr = value.padStart(2, '0');
+		return isValid(
+			parse(`${yearStr}-${monthStr}-${dayStr}`, 'yyyy-MM-dd', new Date()),
+		);
```
cecilia-donnelly, line 439: This is nice, thanks!

src/app/shared/services/edtf-service/edtf.service.ts
```diff
@@ -43,181 +64,173 @@ export interface DateTimeModel {
...
-			return { timezoneOffset: '+00:00', timezoneName: 'UTC' };
+		const converted = this.parseTimeAs24Hour(time);
+		if (!converted) {
+			throw new Error('Invalid time');
 		}
 
-		return { timezoneOffset: offset, timezoneName: '' };
+		const pad = (n: number): string => String(n).padStart(2, '0');
+		return `T${pad(converted.hour)}:${pad(converted.minute)}:${pad(converted.second)}`;
 	}
```
cecilia-donnelly, line 235: I believe this is where we need to handle appending the offset so that it gets stored, but maybe it's *after* this is called in `normalizeEdtfString`.

src/app/shared/services/edtf-service/edtf.service.spec.ts
```diff
@@ -83,6 +77,43 @@ describe('EdtfService', () => {
...
+				expect(result.date.month).toBe('05');
+			});
+
+			it('should parse 198X-05-20 as partial year with full month and day', () => {
+				const result = service.toDateTimeModel('198X-05-20');
+
+				expect(result.date.year).toBe('198');
+				expect(result.date.month).toBe('05');
+				expect(result.date.day).toBe('20');
+			});
```
cecilia-donnelly, line 127: The name of this is a little confusing, but am I understanding correctly that this enforces the rightmost X condition?

cecilia-donnelly (changes requested):
Small change I hope, but we should preserve the offset when we save to the database, even though we show the time as a local string. This is amazing! I didn't look too hard at the components, as suggested, and focused on the EDTF service.

src/app/shared/services/edtf-service/edtf.service.ts
aasandei-vsp: Preserving the functionality is a very good point, I added timezones handling back, even though we are only using the browser's timezone. I wanted to use the offsetToTimeZone method you suggested, but it isn't in the package's public exports, so I added a small local equivalent. Let me know if there is anything else that I might be missing regarding to this.

Bellow is an example of how we are sending the timezone when we update the edtf date --> displayTime property.

image

src/app/shared/services/edtf-service/edtf.service.spec.ts
aasandei-vsp: I do understand the confusion, this is actually testing a more technical decision that I made, based on the fact that undefined/incomplete year/month/day input field translates into X-es into the edtf string date.
In the edtf string date, it will be correctly represented with an X, but it will just leave the input field blank. The DateTimeModel is actually built to make it easier to show all the parts of the edtf date string, into their respective inputs.
So in this case, the year property from the DateTimeModel will look incomplete, as it would look into the input field. But in the edtf string, it will always add the X.
Does this behavior make sense? Maybe there is a different approach that I am not seeing?

Also, related to the rightmost X level 1 rule, we are not restricting it to that, we support directly the level 2 unspecified digits. I have made this choice because both web and the backend support it and it was much easier to implement it this way, taking into account that level 2 was the end goal anyways. In the demo I have presented this and Robert seemed to be ok with the decision. Would it be ok to leave it like that?

aasandei-vsp: Just to show an example for undefined month in the form, but XX-es added in the edtf string at the bottom:

Image

src/app/shared/services/edtf-service/edtf.service.ts
aasandei-vsp: Yes, definitely part of the change!

src/app/shared/services/edtf-service/edtf.service.ts
```diff
@@ -213,8 +220,18 @@ export class EdtfService {
 			throw new Error('Invalid time');
 		}
 
+		const timezoneOffset = time.timezoneOffset ?? this.localTimezoneOffset();
```
cecilia-donnelly, line 223: Ah, this is tricky (of course, it all is). This assumes that the user enters a time in their local time (makes sense) and we just want to record where they are in relation to UTC. I like it!

cecilia-donnelly (approved):
Woohoo! Very excited to get this out!

PR #1050, opened 2026-06-08
[PER-10416] Remove the permanent file format from the sidebar

aasandei-vsp:
TASK: https://permanent.atlassian.net/browse/PER-10620

STEPS TO TEST:

1. Log in or create a new account
2. Upload a record
3. Click the record
EXPECTED: The sidebar appears with details about the record, but there is no "Permanent format" field or value.

PR #1054, opened 2026-06-10
Revert "[PER-10416] Remove the permanent file format from the sidebar"

aasandei-vsp:
Reverts PermanentOrg/web-app#1050


aasandei-vsp:
We do not need to revert this change, as it is very small.

PR #1056, opened 2026-06-11
[PER-10643] Add edtf date and time validation

aasandei-vsp:
# Manual test cases — EDTF date/time input

**Setup:** log in, upload a file, click it.
**EXPECTED:** sidebar with a Date section. Unless a test says "inline picker", open the edit modal via **More options** for each test. The modal footer shows a live preview of the resulting EDTF string when valid, or a red error message when invalid, and the **Save** button is disabled while invalid.

> [!WARNING]
> The date and time inputs are the **same shared component** in both the inline sidebar picker and the **More options** modal. Every test that only exercises those fields — **Basic date entry, Single-digit month/day, Invalid characters, Out-of-range values, Impossible calendar days, and Time** — must be run in **both** places, because their commit behavior differs: the modal disables **Save** while invalid, whereas the inline picker lets you attempt the save and surfaces the error as a toast. Qualifiers, Unknown, and Date ranges exist only in the modal.

> [!NOTE]
> Field-specific messages (invalid characters, out-of-range month/day/hours/minutes/seconds, day-for-month) appear **inline below the offending input**, with that input highlighted red. The modal footer separately shows the generic `The date entered is not valid. Please check the values and try again.` whenever anything is invalid (with **Save** disabled) — the specific message is not duplicated there. Two exceptions with no inline field surface keep their specific text in the footer: `A complete date is required when time is provided.` and `The date range is not valid. Please make sure the start date is before the end date.`

---

## Basic date entry
*(run in both the inline picker and the modal)*

### Year only

1. Type `2026` in the year, leave month and day empty.
   - **EXPECTED:** No error. Preview shows `2026`. Save is enabled.

### Year and month

1. Type `2026` in the year, `05` in the month, leave day empty.
   - **EXPECTED:** No error. Preview shows `2026-05`. Save is enabled.

### Full date

1. Type `2026`, `05`, `20`.
   - **EXPECTED:** No error. Preview shows `2026-05-20`. Save is enabled.

### Auto-advance between segments

1. Type `2026` in the year.
   - **EXPECTED:** Focus jumps to the month input once 4 digits are entered.
2. Type `05` in the month.
   - **EXPECTED:** Focus jumps to the day input once 2 digits are entered.

### Backspace navigation between segments

1. Enter `2026` / `05` / `20`, then place the cursor in the empty day input and press Backspace until day is empty, then Backspace again.
   - **EXPECTED:** Focus moves back to the month input and removes its last digit.
2. Continue pressing Backspace with the month empty.
   - **EXPECTED:** Focus moves back to the year input and removes its last digit.

### Calendar picker

1. Click the calendar icon; select a date.
   - **EXPECTED:** The picker fills year/month/day, clears any field errors, and closes. Preview reflects the selected date.
2. Focus the calendar icon and press Enter or Space.
   - **EXPECTED:** The calendar opens/closes (keyboard toggle works).

---

## Single-digit month/day (entered without zero-padding)
*(run in both the inline picker and the modal)*

### Single-digit month is accepted and zero-padded

1. Enter year `2026`, month `2` (single digit), day `15`.
   - **EXPECTED:** No error. Preview shows `2026-02-15` — the month is zero-padded automatically even though you typed `2`. Save is enabled.
2. Save, then reopen the item.
   - **EXPECTED:** The date persists and the month field now shows `02`.

### Single-digit day is accepted and zero-padded

1. Enter year `2026`, month `05`, day `5` (single digit).
   - **EXPECTED:** No error. Preview shows `2026-05-05`. Save is enabled.

### Both month and day single-digit

1. Enter year `2026`, month `3`, day `7`.
   - **EXPECTED:** No error. Preview shows `2026-03-07`.

### Single-digit month, year-month only (no day)

1. Enter year `2026`, month `4`, leave day empty.
   - **EXPECTED:** No error. Preview shows `2026-04`.

### No premature error while a single digit is typed

1. Type `2` in the month input and stop.
   - **EXPECTED:** No error appears — validation waits until the value is complete. Focus does **not** auto-advance to the day (auto-advance only fires after 2 digits are entered), so you must move to the day input yourself.

### Single-digit month is used for day validation

1. Enter year `2026`, month `2` (single digit), day `30`.
   - **EXPECTED:** `That day does not exist in the selected month and year.` below the day input — the day is validated against **February**, not January. (This is the regression that was fixed: an un-padded month must not be treated as `01`.)
2. Enter year `2021`, month `2`, day `29`.
   - **EXPECTED:** `That day does not exist in the selected month and year.`

### Single-digit `1` month is treated as January

1. Enter year `2026`, month `1` (single digit), day `31`.
   - **EXPECTED:** No error (January has 31 days). Preview shows `2026-01-31`.

### Lone `0` month is rejected (inline, on the month field)

1. Enter year `2026`, month `0` (just a zero), day `15`.
   - **EXPECTED:** `Month must be between 1 and 12.` appears **below the month input** and the month field is highlighted red. A lone `0` is treated as an unfinished value, not a valid month. In the modal the footer shows the generic error and **Save** is disabled; in the inline picker a save attempt surfaces the error.

### Lone `0` day is rejected (inline, on the day field)

1. Enter year `2026`, month `05`, day `0` (just a zero).
   - **EXPECTED:** `Day must be between 1 and 31.` appears **below the day input** and the day field is highlighted red (same behavior as above).

### Lone `0` month is attributed to the month field, not the day

1. Enter year `2024`, month `0`, day `31`.
   - **EXPECTED:** Only `Month must be between 1 and 12.` appears (on the month field). The day field does **not** show `That day does not exist…` — an invalid month is not mislabeled as a day error.

---

## Invalid characters
*(run in both the inline picker and the modal)*

### Invalid characters in date and time

1. Type `19ab` in the year, then `ab` in the hours.
   - **EXPECTED:** The values stay visible (not silently reverted), with `The date contains invalid characters.` / `The time contains invalid characters.` below the inputs and red styling.
2. Clear the inputs.
   - **EXPECTED:** Errors disappear.

---

## Out-of-range values
*(run in both the inline picker and the modal)*

### Out-of-range month

1. Type `13` in the month input.
   - **EXPECTED:** The value stays visible and `Month must be between 1 and 12.` appears below the input.
2. Change the month to `12`.
   - **EXPECTED:** The error disappears.

### Out-of-range day

1. Enter year `2026`, month `01`, then type `32` in the day input.
   - **EXPECTED:** `Day must be between 1 and 31.` appears below the input.
2. Change the day to `31`.
   - **EXPECTED:** The error disappears.

### Out-of-range hours (AM/PM mode)

1. With the format toggle on `AM` or `PM`, type `13` in the hours input.
   - **EXPECTED:** The value stays visible and `Hour must be between 1 and 12.` appears below the time input.
2. Change the hours to `10`.
   - **EXPECTED:** The error disappears.
3. Type `13` again (error reappears), then toggle the format until it shows `24H` (leave the hours as `13`).
   - **EXPECTED:** The error disappears — `13` is valid in 24-hour mode — without editing the hours value. Toggling back to `AM`/`PM` makes `Hour must be between 1 and 12.` reappear.

### Out-of-range hours (24H mode)

1. Toggle the format until it shows `24H`, type `24` in the hours input.
   - **EXPECTED:** `Hour must be between 0 and 23.` appears below the time input.
2. Change the hours to `23`.
   - **EXPECTED:** The error disappears.
3. Type `24` again (error reappears), then toggle the format to `AM` (leave the hours as `24`).
   - **EXPECTED:** The message changes to `Hour must be between 1 and 12.` — `24` is invalid in both modes, but the message reflects the active format's range. (A 12-hour clock's valid hours are `1`–`12`; midnight is `12 AM` and noon is `12 PM`, so there is no `0` and the range is intentionally 1–12, not 0–12.)

### Out-of-range minutes

1. Type `60` in the minutes input.
   - **EXPECTED:** `Minutes must be between 0 and 59.` appears below the time input.
2. Change the minutes to `59`.
   - **EXPECTED:** The error disappears.

### Out-of-range seconds

1. Type `60` in the seconds input.
   - **EXPECTED:** `Seconds must be between 0 and 59.` appears below the time input.
2. Change the seconds to `59`.
   - **EXPECTED:** The error disappears.

---

## Impossible calendar days
*(run in both the inline picker and the modal)*

### Feb 29 in a non-leap year

1. Enter year `2021`, month `02`, then type `29` in the day input.
   - **EXPECTED:** `That day does not exist in the selected month and year.` appears below the day input (the value is **not** silently autocorrected to `2021-03-01`). Save is disabled.
2. Change the year to `2024`.
   - **EXPECTED:** The error disappears (2024 is a leap year). Preview shows `2024-02-29`.

### Feb 29 with a single-digit month

1. Enter year `2021`, month `2` (single digit), then type `29` in the day input.
   - **EXPECTED:** Same as above — `That day does not exist in the selected month and year.` (this previously slipped through with a single-digit month).

### Re-validation when month changes

1. Enter year `2026`, month `01`, day `31` (valid).
   - **EXPECTED:** No error.
2. Change the month to `04`.
   - **EXPECTED:** `That day does not exist in the selected month and year.` appears (the day is re-checked against the new month).

---

## Time
*(run in both the inline picker and the modal)*

### Format toggle (AM / PM / 24H)

1. Enter a time, then click the format toggle repeatedly.
   - **EXPECTED:** It cycles `AM` → `PM` → `24H` → `AM`. The hour is re-validated against the new format (e.g. `13` is invalid in AM/PM but valid in 24H).

### Midnight and noon

1. Enter time `12:00:00` with `AM`.
   - **EXPECTED:** Preview stores midnight (`T00:00:00…`).
2. Enter time `12:00:00` with `PM`.
   - **EXPECTED:** Preview stores noon (`T12:00:00…`).

### Seconds optional

1. Enter hours and minutes only, leave seconds empty.
   - **EXPECTED:** No error; seconds default to `00` in the saved value.

### Time requires a complete date

1. Enter year `2026` only (no month/day), then enter a time.
   - **EXPECTED:** `A complete date is required when time is provided.` in the footer and Save is disabled.
2. Fill in month and day.
   - **EXPECTED:** The error disappears.

---

## Unspecified digits (X)
*(modal only)*

### Unknown month with a known day

1. Enter year `2026`, leave month empty, enter day `11`.
   - **EXPECTED:** No error. Preview shows `2026-XX-11`. Save enabled.

### Partial year

1. Enter `198` in the year (3 digits), leave month/day empty.
   - **EXPECTED:** Preview shows `198X`.
2. Enter `19` in the year (2 digits).
   - **EXPECTED:** Preview shows `19XX`.

### Unknown year with a known month

1. Leave year empty, enter month `05`.
   - **EXPECTED:** Preview shows `XXXX-05`.

---

## Qualifiers
*(modal only)*

### Approximate

1. Enter a full date `2026`/`05`/`20`, toggle **Approximate** on.
   - **EXPECTED:** Preview shows `2026-05-20~`. Save enabled.

### Uncertain

1. Enter a full date, toggle **Uncertain** on.
   - **EXPECTED:** Preview shows `2026-05-20?`.

### Approximate + Uncertain (combined)

1. Enter a full date, toggle both **Approximate** and **Uncertain** on.
   - **EXPECTED:** Preview shows `2026-05-20%`.

### Qualifier with unspecified digits

1. Enter year `2026`, leave month empty, enter day `11`, toggle **Approximate** on.
   - **EXPECTED:** No "not valid" error. Preview shows `2026-XX-11~`. Save enabled.
2. Save, then reopen the item.
   - **EXPECTED:** Month is blank, day is `11`, and **Approximate** is still on (the value round-trips).

### Qualifier combined with a time is rejected

1. Enter a full date `2026`/`05`/`20`, enter a time (e.g. `10:30`), then toggle **Approximate** (or **Uncertain**) on.
   - **EXPECTED:** The footer shows `The date entered is not valid. Please check the values and try again.` and Save is disabled (a qualifier with a time is unsupported).

### Unknown

1. Toggle **Unknown** on.
   - **EXPECTED:** The date/time fields become disabled/cleared. Preview shows `XXXX-XX-XX`. Save enabled.
2. Toggle **Unknown** off.
   - **EXPECTED:** The previous field values are restored.

### Unknown is mutually exclusive with Approximate/Uncertain

1. Toggle **Unknown** on.
   - **EXPECTED:** **Approximate/Uncertain** are disabled.

---

## Date ranges
*(modal only)*

### Basic range

1. Toggle **Use a date range** on. Enter start `2026`/`05` and end `2027`/`06`.
   - **EXPECTED:** Preview shows `2026-05/2027-06`. Save enabled.
   
### Range with time on both sides

1. Toggle **Use a date range** on. Enter start `2026`/`05`/`20` with time `10:30:00`, and end `2027`/`06`/`15` with time `14:00:00`.
   - **EXPECTED:** No error. Preview shows the full interval with both times and their timezone offsets, e.g. `2026-05-20T10:30:00+HH:MM/2027-06-15T14:00:00+HH:MM`. Save enabled.
2. Save, then reopen the item.
   - **EXPECTED:** Both sides round-trip — each side shows its own date, wall-clock time, AM/PM (or 24H), and offset unchanged (not shifted to UTC).

### Range with time on only one side

1. With a range, enter start `2026`/`05`/`20` with time `10:30:00`, and end `2027`/`06`/`15` with **no** time.
   - **EXPECTED:** No error. Preview shows `2026-05-20T10:30:00+HH:MM/2027-06-15` (a timed start with a date-only end is allowed). Save enabled.
2. Repeat with the time on the **end** side only (start date-only, end timed).
   - **EXPECTED:** Preview shows `2026-05-20/2027-06-15T14:00:00+HH:MM`. Save enabled.

### A range side with a time still requires a complete date on that side

1. With a range, enter start year `2026` only (no month/day) but add a time to the start, and a valid end date.
   - **EXPECTED:** The footer shows `A complete date is required when time is provided.` and Save is disabled (the requirement applies per side).
2. Fill in the start month and day.
   - **EXPECTED:** The error disappears.

### Range validation (start after end)

1. With a range, enter start `2027` and end `2026`.
   - **EXPECTED:** `The date range is not valid. Please make sure the start date is before the end date.` in the footer and Save is disabled.

### Open-ended range — "sometime after"

1. With a range, enter a start date and leave the end empty.
   - **EXPECTED:** Preview shows `2026-05/..`. In the sidebar it reads **Sometime after**.

### Open-ended range — "sometime before"

1. With a range, leave the start empty and enter an end date.
   - **EXPECTED:** Preview shows `../2027-06`. In the sidebar it reads **Sometime before**.

### Per-side qualifiers on a range

1. With a range, toggle **Approximate** on the start side and **Uncertain** on the end side.
   - **EXPECTED:** Preview shows the qualifier on each side, e.g. `2026-05~/2027-06?`.

### Unspecified digits on one side of a range

1. With a range, enter start `2026`/blank month/`11` with **Approximate**, and end `2027`/`01`.
   - **EXPECTED:** Preview shows `2026-XX-11~/2027-01`. Round-trips on reopen.

### Unknown on one side of a range

1. With a range, toggle **Unknown** on the start side and enter an end date.
   - **EXPECTED:** The start side shows as **Unknown**; Save enabled.

---

### Clear date and time (inline picker)

1. On an item that already has a date, open the inline sidebar picker and click **Clear date and time**, then save.
   - **EXPECTED:** The date is removed from the item (stored value becomes empty/null).

### Clear start / clear end (modal)

1. In the modal with a range, click **Clear start date and time**.
   - **EXPECTED:** Only the start side is cleared.
2. Click **Clear end date and time**.
   - **EXPECTED:** Only the end side is cleared.

---

### Save disabled while invalid (modal)

1. Enter any invalid value (e.g. month `13`).
   - **EXPECTED:** The **Save** button is disabled, the field shows its specific message inline, and the footer shows the generic error.
2. Correct the value.
   - **EXPECTED:** **Save** becomes enabled and the footer shows the valid preview.

### Cancel discards changes

1. Change a value, then click **Cancel** (or click outside the inline picker).
   - **EXPECTED:** No change is saved; the previously stored value is restored.

### Round-trip persistence

1. For each of: full date, year-only, unspecified month, partial year, approximate/uncertain/combined, unknown, a date range, an open-ended range, and a full date + time — save it, then reopen the item.
   - **EXPECTED:** Each value displays correctly in the sidebar and reopens in the modal with the same fields, qualifiers, and time populated.
  
 ### New folder persistence 
1. Create a new folder and add an edtf date to it.
  - **EXPECTED:** Value displays correctly in the sidebar.
2. Change the date on the folder
  - **EXPECTED:** The new value displays correctly in the sidebar and modal, with no general error or BE error.

### DIFF
--- a/src/app/shared/components/datepicker-input/datepicker-input.component.html
+++ b/src/app/shared/components/datepicker-input/datepicker-input.component.html
@@ -1,51 +1,57 @@
-<div class="pr-date-input-group" [class.active]="showDatepicker()">
-	<input
-		#yearInput
-		type="text"
-		class="pr-date-segment year"
-		[value]="date.year"
-		[disabled]="disabled"
-		(input)="updateYear($event)"
-		placeholder="YYYY"
-		maxlength="4"
-	/>
-	<span class="pr-separator">/</span>
-	<input
-		#monthInput
-		type="text"
-		class="pr-date-segment"
-		[value]="date.month ?? ''"
-		[disabled]="disabled"
-		(input)="updateMonth($event)"
-		(keydown)="onMonthKeydown($event)"
-		placeholder="MM"
-		maxlength="2"
-	/>
-	<span class="pr-separator">/</span>
-	<input
-		#dayInput
-		type="text"
-		class="pr-date-segment"
-		[value]="date.day ?? ''"
-		[disabled]="disabled"
-		(input)="updateDay($event)"
-		(keydown)="onDayKeydown($event)"
-		placeholder="DD"
-		maxlength="2"
-	/>
-	<div class="pr-calendar-icon-wrapper">
-		<span
-			class="pr-icon-button"
-			role="button"
-			tabindex="0"
-			[class.disabled]="disabled"
-			(click)="toggleDatepicker()"
-			(keydown.enter)="toggleDatepicker()"
-			(keydown.space)="toggleDatepicker()"
-		>
-			<i class="material-icons">calendar_today</i>
-		</span>
+<div class="pr-date-input-wrap">
+	<div
+		class="pr-date-input-group"
+		[class.active]="showDatepicker()"
+		[class.has-error]="!!currentError()"
+	>
+		<input
+			#yearInput
+			type="text"
+			class="pr-date-segment year"
+			[value]="date.year"
+			[disabled]="disabled"
+			(input)="updateYear($event)"
+			placeholder="YYYY"
+			maxlength="4"
+		/>
+		<span class="pr-separator">/</span>
+		<input
+			#monthInput
+			type="text"
+			class="pr-date-segment"
+			[value]="date.month ?? ''"
+			[disabled]="disabled"
+			(input)="updateMonth($event)"
+			(keydown)="onMonthKeydown($event)"
+			placeholder="MM"
+			maxlength="2"
+		/>
+		<span class="pr-separator">/</span>
+		<input
+			#dayInput
+			type="text"
+			class="pr-date-segment"
+			[value]="date.day ?? ''"
+			[disabled]="disabled"
+			(input)="updateDay($event)"
+			(keydown)="onDayKeydown($event)"
+			placeholder="DD"
+			maxlength="2"
+		/>
+		<div class="pr-calendar-icon-wrapper">
+			<span
+				class="pr-icon-button"
+				[class.disabled]="disabled"
+				(click)="toggleDatepicker()"
+			>
+				<i class="material-icons">calendar_today</i>
+			</span>
+		</div>
 	</div>
+
+	@if (currentError()) {
+		<span class="pr-input-error">{{ currentError() }}</span>
+	}
 </div>
 
 @if (showDatepicker()) {
--- a/src/app/shared/components/datepicker-input/datepicker-input.component.html
+++ b/src/app/shared/components/datepicker-input/datepicker-input.component.html
@@ -41,8 +41,12 @@
 		<div class="pr-calendar-icon-wrapper">
 			<span
 				class="pr-icon-button"
+				role="button"
+				tabindex="0"
 				[class.disabled]="disabled"
 				(click)="toggleDatepicker()"
+				(keydown.enter)="toggleDatepicker()"
+				(keydown.space)="toggleDatepicker()"
 			>
 				<i class="material-icons">calendar_today</i>
 			</span>
--- a/src/app/shared/components/timepicker-input/timepicker-input.component.html
+++ b/src/app/shared/components/timepicker-input/timepicker-input.component.html
@@ -48,8 +48,12 @@
 		<div class="pr-time-icon-wrapper">
 			<span
 				class="pr-icon-button"
+				role="button"
+				tabindex="0"
 				[class.disabled]="disabled"
 				(click)="toggleTimepicker()"
+				(keydown.enter)="toggleTimepicker()"
+				(keydown.space)="toggleTimepicker()"
 			>
 				<i class="material-icons">access_time</i>
 			</span>
--- a/src/app/shared/components/timepicker-input/timepicker-input.component.html
+++ b/src/app/shared/components/timepicker-input/timepicker-input.component.html
@@ -1,54 +1,64 @@
-<div class="pr-time-input-group" [class.active]="showTimepicker()">
-	<input
-		#hoursInput
-		type="text"
-		class="pr-time-segment"
-		[value]="time.hours ?? ''"
-		[disabled]="disabled"
-		(input)="updateTime($event, 'hours', minutesInput)"
-		placeholder="hh"
-		maxlength="2"
-	/>
-	<span class="pr-colon">:</span>
-	<input
-		#minutesInput
-		type="text"
-		class="pr-time-segment"
-		[value]="time.minutes ?? ''"
-		[disabled]="disabled"
-		(input)="updateTime($event, 'minutes', secondsInput?.nativeElement)"
-		(keydown)="onMinutesKeydown($event)"
-		placeholder="mm"
-		maxlength="2"
-	/>
-	<span class="pr-colon">:</span>
-	<input
-		#secondsInput
-		type="text"
-		class="pr-time-segment"
-		[value]="time.seconds ?? ''"
-		[disabled]="disabled"
-		(input)="updateTime($event, 'seconds')"
-		(keydown)="onSecondsKeydown($event)"
-		placeholder="ss"
-		maxlength="2"
-	/>
-	<button class="pr-am-pm-toggle" [disabled]="disabled" (click)="cycleFormat()">
-		{{ formatLabel() }}
-	</button>
-	<div class="pr-time-icon-wrapper">
-		<span
-			class="pr-icon-button"
-			role="button"
-			tabindex="0"
-			[class.disabled]="disabled"
-			(click)="toggleTimepicker()"
-			(keydown.enter)="toggleTimepicker()"
-			(keydown.space)="toggleTimepicker()"
+<div class="pr-time-input-wrap">
+	<div
+		class="pr-time-input-group"
+		[class.active]="showTimepicker()"
+		[class.has-error]="!!currentError()"
+	>
+		<input
+			#hoursInput
+			type="text"
+			class="pr-time-segment"
+			[value]="time.hours ?? ''"
+			[disabled]="disabled"
+			(input)="updateTime($event, 'hours', minutesInput)"
+			placeholder="hh"
+			maxlength="2"
+		/>
+		<span class="pr-colon">:</span>
+		<input
+			#minutesInput
+			type="text"
+			class="pr-time-segment"
+			[value]="time.minutes ?? ''"
+			[disabled]="disabled"
+			(input)="updateTime($event, 'minutes', secondsInput)"
+			(keydown)="onMinutesKeydown($event)"
+			placeholder="mm"
+			maxlength="2"
+		/>
+		<span class="pr-colon">:</span>
+		<input
+			#secondsInput
+			type="text"
+			class="pr-time-segment"
+			[value]="time.seconds ?? ''"
+			[disabled]="disabled"
+			(input)="updateTime($event, 'seconds')"
+			(keydown)="onSecondsKeydown($event)"
+			placeholder="ss"
+			maxlength="2"
+		/>
+		<button
+			class="pr-am-pm-toggle"
+			[disabled]="disabled"
+			(click)="cycleFormat()"
 		>
-			<i class="material-icons">access_time</i>
-		</span>
+			{{ formatLabel() }}
+		</button>
+		<div class="pr-time-icon-wrapper">
+			<span
+				class="pr-icon-button"
+				[class.disabled]="disabled"
+				(click)="toggleTimepicker()"
+			>
+				<i class="material-icons">access_time</i>
+			</span>
+		</div>
 	</div>
+
+	@if (currentError()) {
+		<span class="pr-input-error">{{ currentError() }}</span>
+	}
 </div>
 
 @if (showTimepicker()) {
--- a/src/app/shared/services/edtf-service/edtf.service.ts
+++ b/src/app/shared/services/edtf-service/edtf.service.ts
@@ -34,6 +34,8 @@ export const UNKNOWN_VALUE = 'XXXX-XX-XX';
 
 export const MONTH_RANGE_ERROR = 'Month must be between 1 and 12.';
 export const DAY_RANGE_ERROR = 'Day must be between 1 and 31.';
+export const INVALID_DAY_FOR_MONTH_ERROR =
+	'That day does not exist in the selected month and year.';
 
 const DIGITS_ONLY = /^\d*$/;
 
@@ -336,7 +338,26 @@ export class EdtfService {
 		if (!hasDay) return `${year}-${month}`;
 
 		const day = this.padMonthOrDay(date.day);
-		return `${year}-${month}-${day}`;
+		const dateStr = `${year}-${month}-${day}`;
+
+		// A fully-numeric date with an in-range month and day can still be an
+		// impossible calendar day (e.g. Feb 29 in a non-leap year, or Apr 31). The
+		// edtf library silently rolls those forward (2021-02-29 becomes 2021-03-01),
+		// so reject them here instead of storing the shift
+		const monthNumber = parseInt(month, 10);
+		const dayNumber = parseInt(day, 10);
+		if (
+			/^\d{4}-\d{2}-\d{2}$/.test(dateStr) &&
+			monthNumber >= 1 &&
+			monthNumber <= 12 &&
+			dayNumber >= 1 &&
+			dayNumber <= 31 &&
+			!isValid(parse(dateStr, 'yyyy-MM-dd', new Date()))
+		) {
+			throw new Error(INVALID_DAY_FOR_MONTH_ERROR);
+		}
+
+		return dateStr;
 	}
 
 	// Shared by serialization (toEdtfDate) and display (formatDateForDisplay)
@@ -510,11 +531,10 @@ export class EdtfService {
 			return 'A complete date is required when time is provided.';
 		}
 
-		if (message.includes('must be between')) {
-			// Already a human-readable segment range message (month/day).
-			return (error as Error).message;
-		}
-
+		// Segment errors (month/day range, day-for-month) are shown inline under
+		// the offending field, so the footer/toast only needs the generic message —
+		// no need to duplicate the specific one here. Errors without an inline
+		// counterpart (the date-range and complete-date cases above) keep theirs.
 		return 'The date entered is not valid. Please check the values and try again.';
 	}
 
@@ -681,8 +701,11 @@ export class EdtfService {
 		if (/^\d$/.test(value)) return true;
 		// Defaults let day be typed before year/month are filled in.
 		// 2000 is a leap year (allows Feb 29); 01 has 31 days (most permissive).
+		// A single-digit month is a complete value (e.g. '2' is February), so pad
+		// it rather than discarding it — otherwise a bad day like Feb 29 would be
+		// validated against January and wrongly pass.
 		const yearStr = year.length === 4 ? year : '2000';
-		const monthStr = month.length === 2 ? month : '01';
+		const monthStr = month ? month.padStart(2, '0') : '01';
 		const dayStr = value.padStart(2, '0');
 		return isValid(
 			parse(`${yearStr}-${monthStr}-${dayStr}`, 'yyyy-MM-dd', new Date()),
@@ -704,5 +727,71 @@ export class EdtfService {
 			return options.rangeMessage ?? null;
 		}
 		return null;
+	}
+
+	getMonthError(
+		value: string,
+		options: { invalidCharsMessage: string },
+	): string | null {
+		// A lone "0" is an unfinished value ("05" minus a keystroke), not a valid
+		// month — flag it inline instead of leaving it to serialization.
+		if (value === '0') return MONTH_RANGE_ERROR;
+
+		return this.getSegmentError(value, {
+			invalidCharsMessage: options.invalidCharsMessage,
+			isWithinRange: (month) => this.isValidMonth(month),
+			rangeMessage: MONTH_RANGE_ERROR,
+		});
+	}
+
+	// Validates a day segment in tiers so each failure gets a specific message:
+	// a lone "0", then the plain 1–31 range, then whether that day actually
+	// exists in the selected month/year (e.g. Feb 29 in a non-leap year).
+	getDayError(
+		value: string,
+		year: string,
+		month: string,
+		options: { invalidCharsMessage: string },
+	): string | null {
+		// A lone "0" is an unfinished value ("05" minus a keystroke), not a valid
+		// day — flag it inline instead of leaving it to serialization.
+		if (value === '0') return DAY_RANGE_ERROR;
+
+		const rangeError = this.getSegmentError(value, {
+			invalidCharsMessage: options.invalidCharsMessage,
+			isWithinRange: (day) => this.isDayInRange(day),
+			rangeMessage: DAY_RANGE_ERROR,
+		});
+		if (rangeError) return rangeError;
+
+		// Only cross-check the day against the month when the month is itself a
+		// valid month; when it is not (e.g. "0" or "13"), the error belongs to the
+		// month field, so we must not mislabel it as a day-for-month problem.
+		if (
+			value.length === 2 &&
+			this.isResolvableMonth(month) &&
+			!this.isValidDay(value, year, month)
+		) {
+			return INVALID_DAY_FOR_MONTH_ERROR;
+		}
+
+		return null;
+	}
+
+	// The plain 1–31 numeric range, independent of month/year — the first tier of
+	// getDayError, kept separate from the calendar check so an out-of-range day
+	// and a day that does not exist in the month get different messages.
+	private isDayInRange(value: string): boolean {
+		const dayNumber = parseInt(value, 10);
+		return dayNumber >= 1 && dayNumber <= 31;
+	}
+
+	// A month that resolves to a real 1–12 value (single digit or two digits).
+	// Used to gate the day-for-month check so an invalid month does not surface
+	// as a day error.
+	private isResolvableMonth(month: string): boolean {
+		if (!/^\d{1,2}$/.test(month)) return false;
+		const monthNumber = parseInt(month, 10);
+		return monthNumber >= 1 && monthNumber <= 12;
 	}
 }
--- a/src/app/shared/services/edtf-service/edtf.service.ts
+++ b/src/app/shared/services/edtf-service/edtf.service.ts
@@ -25,6 +25,8 @@ export enum EdtfPrecision {
 
 export const UNKNOWN_VALUE = 'XXXX-XX-XX';
 
+const DIGITS_ONLY = /^\d*$/;
+
 export const DEFAULT_TIME: TimeModel = {
 	hours: '',
 	minutes: '',
@@ -197,13 +199,19 @@ export class EdtfService {
 		time: TimeModel,
 		qualifiers?: DateQualifierFlags,
 	): string {
+		const hasCompleteDate = !!(date.year && date.month && date.day);
+		const hasTime = !!time?.hours;
+
+		if (hasTime && !hasCompleteDate) {
+			throw new Error('A complete date is required when time is provided.');
+		}
+
 		const dateStr = this.buildDateString(date);
 		const edtfObject = edtf(dateStr);
 		// Strip any time/timezone the library may append (e.g. T00:00:00.000Z)
 		let result = edtfObject.toEDTF().replace(/T.*$/, '');
 
-		const hasCompleteDate = !!(date.year && date.month && date.day);
-		const timeStr = hasCompleteDate ? this.buildTimeString(time) : '';
+		const timeStr = hasTime ? this.buildTimeString(time) : '';
 
 		if (timeStr) {
 			result = `${result}${timeStr}`;
@@ -231,13 +239,18 @@ export class EdtfService {
 		const year = this.padWithX(date.year, 4);
 		if (!hasMonth && !hasDay) return year;
 
-		const month = hasMonth ? this.padWithX(date.month, 2) : 'XX';
+		const month = hasMonth ? this.padMonthOrDay(date.month) : 'XX';
 		if (!hasDay) return `${year}-${month}`;
 
-		// A day is a discrete value, so a single digit is zero-padded on the
-		// left ("9" -> "09"); X-padding would produce an invalid day (90-99).
-		const day = date.day.padStart(2, '0');
+		const day = this.padMonthOrDay(date.day);
 		return `${year}-${month}-${day}`;
+	}
+
+	private padMonthOrDay(value: string): string {
+		// A single 1–9 digit is treated as a complete value (e.g. '5' → '05'),
+		// since the user can't be mid-typing a two-digit value starting with 2–9.
+		if (/^[1-9]$/.test(value)) return `0${value}`;
+		return this.padWithX(value, 2);
 	}
 
 	private padWithX(value: string, width: number): string {
@@ -360,6 +373,10 @@ export class EdtfService {
 			message.includes('invalid upper bound')
 		) {
 			return 'The date range is not valid. Please make sure the start date is before the end date.';
+		}
+
+		if (message.includes('complete date is required')) {
+			return 'A complete date is required when time is provided.';
 		}
 
 		return 'The date entered is not valid. Please check the values and try again.';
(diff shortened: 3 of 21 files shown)


aasandei-vsp:
@cecilia-donnelly @slifty I have managed to rebase this one on the branch that's been checked out from too :D

cecilia-donnelly:
Thanks for the branch management! For others reading this (QA), this needs a feature flag to be turned on: `edtf-date`.

src/app/shared/components/datepicker-input/datepicker-input.component.html
```diff
@@ -1,51 +1,57 @@
...
-	/>
-	<div class="pr-calendar-icon-wrapper">
-		<span
-			class="pr-icon-button"
-			role="button"
-			tabindex="0"
-			[class.disabled]="disabled"
-			(click)="toggleDatepicker()"
-			(keydown.enter)="toggleDatepicker()"
-			(keydown.space)="toggleDatepicker()"
```
cecilia-donnelly, line 44: I think we still want these keyboard-only interactions to be possible.

src/app/shared/components/timepicker-input/timepicker-input.component.html
```diff
@@ -1,54 +1,64 @@
...
-	</button>
-	<div class="pr-time-icon-wrapper">
-		<span
-			class="pr-icon-button"
-			role="button"
-			tabindex="0"
-			[class.disabled]="disabled"
-			(click)="toggleTimepicker()"
-			(keydown.enter)="toggleTimepicker()"
-			(keydown.space)="toggleTimepicker()"
```
cecilia-donnelly, line 47: Same note on keyboard accessibility.

cecilia-donnelly (commented):
I love the way this is looking! The "sometime before"/"sometime after" language for an open ended range is very cool.

- I did see that if I enter February 29 in a non-leap-year, the error is "Day must be between 1 and 31" which is not exactly what we want to express there. I'm not sure what this should be, but maybe @omnignorant knows.
- If I leave both start and end dates empty and indicate that this is a range then I get "sometime before .." which isn't exactly right either
- If I have an open range (no start or end dates), then remove the "range" indicator, the new value is an empty string which gets an error from the backend. Maybe that's fine, but maybe we should instead send "Unknown"? QA/design question.
- the sidebar behaves slightly differently from the modal for invalid data. The sidebar allows you to click "save" with only a time filled (no date) while the modal blocks that path. The time-only value doesn't save, but in the sidebar the user gets a "We could not update" error in a toast instead of the nicer inline "A complete date is required when time is provided."

None of these are showstoppers, but if possible would be nice to fix!

src/app/shared/components/datepicker-input/datepicker-input.component.html
aasandei-vsp: That's a great catch! Thank you! I removed it by mistake, so I'll put it back.

src/app/shared/components/timepicker-input/timepicker-input.component.html
aasandei-vsp: That's a great catch! Thank you! I removed it by mistake, so I'll put it back.

aasandei-vsp:
When the date and time fields are undefined, we send null to the BE PATCH for the displayTime property and then we show the date field empty in the sidebar.
The problem is, the displayTime is set to null even if it has never been set by the user, so we cannot rely on the displayDT fallback.
@cecilia-donnelly How should we deal with this situation? Right now, if we just remove the displayDT fallback for the sidebar, the list shows the displayDT and it looks inconsistent.

image

src/app/shared/services/edtf-service/edtf.service.ts
```diff
@@ -231,15 +239,20 @@ export class EdtfService {
...
-		// left ("9" -> "09"); X-padding would produce an invalid day (90-99).
-		const day = date.day.padStart(2, '0');
+		const day = this.padMonthOrDay(date.day);
 		return `${year}-${month}-${day}`;
 	}
 
+	private padMonthOrDay(value: string): string {
+		// A single 1–9 digit is treated as a complete value (e.g. '5' → '05'),
+		// since the user can't be mid-typing a two-digit value starting with 2–9.
+		if (/^[1-9]$/.test(value)) return `0${value}`;
```
slifty, line 370: This is true for 2-9 but it isn't true for 1, specifically (since it could be `10` or `11` or `12`).

That could be the intent, but the commented reasoning leaves it somewhat open for interpretation, and you might want to use `1` as the example rather than `5` to make that intention very very clear.

Side note: I do think that it's can be nice to have systems that work predictbly / explicitly and communicate requirements well, rather than ones that might change input shape in unexpected ways (e.g. inserting a 0 as opposed to simply signaling an issue)

How can we be sure that a user typing `2020-1` intended `2020-01` as opposed to either `2020-1X`?

src/app/shared/services/edtf-service/edtf.service.ts
```diff
@@ -231,15 +239,20 @@ export class EdtfService {
...
 		if (!hasMonth && !hasDay) return year;
 
-		const month = hasMonth ? this.padWithX(date.month, 2) : 'XX';
+		const month = hasMonth ? this.padMonthOrDay(date.month) : 'XX';
 		if (!hasDay) return `${year}-${month}`;
 
-		// A day is a discrete value, so a single digit is zero-padded on the
-		// left ("9" -> "09"); X-padding would produce an invalid day (90-99).
-		const day = date.day.padStart(2, '0');
+		const day = this.padMonthOrDay(date.day);
```
slifty, line 340: padMonthOrDay will turn `0` into `0X` -- is that the intended behavior?

I suppose this is another case of "can we know for certain the intention of the user?  Do we want to change it for them as opposed to simply (and clearly) telling them that the formatting is incorrect and guiding them on how to correct it?

src/app/shared/services/edtf-service/edtf.service.ts
```diff
@@ -231,15 +239,20 @@ export class EdtfService {
...
 		return `${year}-${month}-${day}`;
 	}
 
+	private padMonthOrDay(value: string): string {
```
slifty, line 365: Because we're changing the input's meaning in this service translates text into something different than what was input; claude flagged that this is implemented different from how it is rendered via `formatDate`.

I do think we need to have the logic for translating user input into serialized format in *one* place.  Maybe the serializer / normalizing logic lives in a common area that is used both for rendering and writing?

Or maybe the normalizer (which is really more of an "inference attempt") shouldn't happen at all in a hidden or automatic way.  What I mean is if the string the user is typing isn't valid EDTF we could just say "this isn't valid as written, here's a suggested edit" and then they can opt to use the edit, or they can ignore it.

slifty (commented):
I might be coming at this with the wrong design philosophy, so please forgive me for possibly questioning defined specification!

I think right now we have code that is silently attempting to make inferences about user intent which may or may not actually be correct.

I'd encourage we reconsider this approach, and instead simply convey "this isn't correct, you need to fix it by X" (or possibly offering a suggested correction).

---

All that said, if we do decide to keep attempted inference we should make sure that the logic for save / serialization matches the logic for formatting / rendering.

cecilia-donnelly:
@aasandei-vsp and I talked about the null problem in `displayTime` during a meeting. What we will do is remove the fallback on these branches and complete the migration where we copy `displayDt` values to null `displayTime` values before we flip the feature flag. In the old UI, there is no way for a user to clear a date, so the fallback will work as intended. This requires that these intermediate PRs also consider the feature flag in the file list, though, because we should keep the fallback in deployed code until the new UI is released (i.e., the feature flag is turned on).

More succinctly:
- Behind the `edtf-date` feature flag, remove the fallback in the file list
- Do the same on other branches working on this feature

src/app/shared/services/edtf-service/edtf.service.ts
cecilia-donnelly: The "suggested edit" is a good idea, but ultimately a product question.

However, I do agree that the same rules should be applied for rendering and writing and it seems like a great idea to have them in the same place.

cecilia-donnelly (changes requested):
Officially requesting changes so this doesn't show up in my "review requested" list anymore. The specific change requested is to remove the fallback in the file list behind the feature flag.

aasandei-vsp:
> Officially requesting changes so this doesn't show up in my "review requested" list anymore. The specific change requested is to remove the fallback in the file list behind the feature flag.

I have created a commit specifically for this situation, making it easier to review.

aasandei-vsp:
> I might be coming at this with the wrong design philosophy, so please forgive me for possibly questioning defined specification!
> 
> I think right now we have code that is silently attempting to make inferences about user intent which may or may not actually be correct.
> 
> I'd encourage we reconsider this approach, and instead simply convey "this isn't correct, you need to fix it by X" (or possibly offering a suggested correction).
> 
> All that said, if we do decide to keep attempted inference we should make sure that the logic for save / serialization matches the logic for formatting / rendering.

There are indeed a lot of scenarios and dealing with each one of them proved to be slightly complex. Your suggestions are great and correct, but at some point it was decided that we need to silently make some decisions for the user, as not everyone is savvy when it comes to edtf dates. I'll be addressing the rest of the comments one by one.

src/app/shared/services/edtf-service/edtf.service.ts
aasandei-vsp: TL;DR We can't, really, with the current design/specs. So we made the decision to choose `2020-01`

The padding runs when the user hits Save, not per keystroke — so at that point 1 is the finished value, not the first digit of 10–12.  I've reworded the comment to use 1 as the example, and the preview now renders the same interpretation ("January 2020") because display and serialization share the same helper (see the one-place refactor below).
The discussion related to adding X-es in the date was, we do not allow the user to do it, we infer it from the user leaving the field empty. Which again, opened the door to a lot of interpretation, so I had to make some technical decisions about it myself. We do have a task for edtf improvements were we could refine even further the use of unspecified digits.

src/app/shared/services/edtf-service/edtf.service.ts
aasandei-vsp: Agreed — changed. A lone 0 can never become a valid month/day, so instead of guessing 0X we now reject it at save with "Month must be between 1 and 12." / "Day must be between 1 and 31.", shown inline.

src/app/shared/services/edtf-service/edtf.service.ts
aasandei-vsp: Done — the display formatting moved into EdtfService.formatDateForDisplay, built on the same padding helpers the serializer uses, so rendering and writing can't diverge anymore (there's a spec asserting they agree).

src/app/shared/services/edtf-service/edtf.service.ts
```diff
@@ -362,6 +375,10 @@ export class EdtfService {
 			return 'The date range is not valid. Please make sure the start date is before the end date.';
 		}
 
+		if (message.includes('complete date is required')) {
+			return 'A complete date is required when time is provided.';
```
cecilia-donnelly, line 531: This is smart.

src/app/shared/components/datepicker-input/datepicker-input.component.html
cecilia-donnelly, line 44: Why remove these?

src/app/shared/components/timepicker-input/timepicker-input.component.html
cecilia-donnelly, line 47: Same question - why remove this? (Maybe this isn't acting as a button anymore?)

cecilia-donnelly (changes requested):
The code makes sense but there are a couple things I found while testing it out:

- Clear date doesn't work because somewhere we're sending an empty string instead of explicit null
- I can't enter a month value like "3" -- the sidebar simply doesn't show anything if I try. "03" works, but I should be able to enter a plain "3" (or any other single digit) and have it be padded with a zero afterward
- It looks like the same is happening for hours
- It lets me enter a value like "2026-XX-10" but then I can't set it to "uncertain"  or "approximate." That might be an EDTF restriction, do you know?

I would go back through and do some testing again -- I know this has been waiting a while so maybe there are some odd interactions?

Also I see that you put the keyboard interactions back, disregard my inline comments! Thank you!

src/app/shared/components/datepicker-input/datepicker-input.component.html
aasandei-vsp: It's true that at some point in the past Claude removed them for some reason and I didn't notice and I'm glad you did! Now it's all in there, thank you for having such a thorough look!

src/app/shared/components/timepicker-input/timepicker-input.component.html
aasandei-vsp: It's true that at some point in the past Claude removed them for some reason and I didn't notice and I'm glad you did! Now it's all in there, thank you for having such a thorough look!

aasandei-vsp:
> The code makes sense but there are a couple things I found while testing it out:
> 
> * Clear date doesn't work because somewhere we're sending an empty string instead of explicit null
> * I can't enter a month value like "3" -- the sidebar simply doesn't show anything if I try. "03" works, but I should be able to enter a plain "3" (or any other single digit) and have it be padded with a zero afterward
> * It looks like the same is happening for hours
> * It lets me enter a value like "2026-XX-10" but then I can't set it to "uncertain"  or "approximate." That might be an EDTF restriction, do you know?
> 
> I would go back through and do some testing again -- I know this has been waiting a while so maybe there are some odd interactions?
> 
> Also I see that you put the keyboard interactions back, disregard my inline comments! Thank you!

**Clear date+only one digit**
I have checked the mentioned behaviors and the first three I couldn't reproduce. I have added a small video, from the branch deployed on dev, where things seem to be working as expected. 
The only thing worth mentioning, when you only insert one digit, the next input won't be focused, but that makes sense.
Please let me know if I'm missing something:
https://www.loom.com/share/c0639bb8f5ba4275993d71afa584b61b

**Unspecified digits+qualifiers**
TL;DR Specs unclear to me(no example to actually confirm this behavior).  BE supports it. Made workaround in FE.

The library's grammar treats "unspecified digits" (X) and "qualification" ( ~ / ? / % ) as separate, non-combinable features, so it throws a syntax error on anything like 2026-XX-10~ — regardless of where the qualifier is placed. That error was being caught and surfaced as the generic "date is not valid" message, which is why you could enter 2026-XX-10 but not mark it approximate/uncertain.

I verified the backend happily accepts and stores these strings (2026-XX-10~ and friends), so the restriction was purely FE. The fix works around the library by stripping the trailing qualifier before parsing/validating, then reattaching the flags to the model ourselves — so 2026-XX-10~ now serializes, saves, and parses back into the picker correctly, in both directions.

One intentional boundary: a qualifier combined with a time (e.g. 2026-01-01T10:00:00~) is still rejected — the library doesn't support that combination either, and since an unspecified date and a time can't coexist anyway, we kept that blocked rather than silently allowing it.

cecilia-donnelly (commented):
Thanks for this, @aasandei-vsp! I tested on dev this time to avoid any local environment oddities. I did find one edge case. When I put in "2021-2-29" (i.e., a leap day in a non-leap year) instead of erroring, the UI autocorrected to "2021-03-01" (no doubt a Date library fix). I think we want to give an error there. Final little things!

cecilia-donnelly:
So much here! So close! Thank you for this long list of tests.

I did find another weird case. I made a new folder on dev ("edtf-tests") for the testing. I intended to test on a record in that folder, but accidentally ran one on the folder itself. I tried to edit the date from "1990-09" to "1990-07." I got a 200 response from the PATCH call in devtools with the correct new value, but the main (top of screen) error toast said that it couldn't "update the item you selected" and in fact it did not. In the console I see `Failed to save item property TypeError: can't access property "update" of undefined`. I'm sorry, I really wanted to be done with edge cases! I am going through the manual tests you gave above and will edit this comment with anything else I find. I assume this is a problem with folders and records being treated differently somehow. I did test on dev so hopefully you can reproduce this problem on a folder.

- Minor: range errors (e.g., seconds greater than 59) appear correctly inline, but don't block save (in the sidebar only, correct in modal)
- Just a question and maybe a need for a design change: Do you know why this range is invalid? `"displayTime" failed custom validation because 2020-11~/2024-08-31T04:30:00-05:00 is not valid Level 2 EDTF: Invalid interval start: Invalid date format: 2020-11~`? (It's a backend error). If we can't have an approximate start date, which seems like it might be the case, then maybe we should remove the qualifier toggles from the start date on a range?

Some notes that aren't blockers for this PR, but would be improvements:
- Side note/different ticket, but is there a way to re-select the same record after refresh?
- Hitting "enter" should save if possible, except if focus is on one of the elements that has a different "enter" definition (datepicker, am/pm, etc)
- dates with qualifiers get "Invalid date" in the file list -- will be addressed in PER-10655, just wanted to call it out as an explicit case

aasandei-vsp:
> So much here! So close! Thank you for this long list of tests.
> 
> I did find another weird case. I made a new folder on dev ("edtf-tests") for the testing. I intended to test on a record in that folder, but accidentally ran one on the folder itself. I tried to edit the date from "1990-09" to "1990-07." I got a 200 response from the PATCH call in devtools with the correct new value, but the main (top of screen) error toast said that it couldn't "update the item you selected" and in fact it did not. In the console I see `Failed to save item property TypeError: can't access property "update" of undefined`. I'm sorry, I really wanted to be done with edge cases! I am going through the manual tests you gave above and will edit this comment with anything else I find. I assume this is a problem with folders and records being treated differently somehow. I did test on dev so hopefully you can reproduce this problem on a folder.
> 
> * Minor: range errors (e.g., seconds greater than 59) appear correctly inline, but don't block save (in the sidebar only, correct in modal)
> * Just a question and maybe a need for a design change: Do you know why this range is invalid? `"displayTime" failed custom validation because 2020-11~/2024-08-31T04:30:00-05:00 is not valid Level 2 EDTF: Invalid interval start: Invalid date format: 2020-11~`? (It's a backend error). If we can't have an approximate start date, which seems like it might be the case, then maybe we should remove the qualifier toggles from the start date on a range?
> 
> Some notes that aren't blockers for this PR, but would be improvements:
> 
> * Side note/different ticket, but is there a way to re-select the same record after refresh?
> * Hitting "enter" should save if possible, except if focus is on one of the elements that has a different "enter" definition (datepicker, am/pm, etc)
> * dates with qualifiers get "Invalid date" in the file list -- will be addressed in PER-10655, just wanted to call it out as an explicit case

This is quite an edge case, thank you for looking into it. I couldn't get it to reproduce on my end, but the error you pasted helped me track it down. You were right that the save itself worked; the problem was in the code that syncs the server's response back onto the folder afterward. It matches the response to the folder by folderId, but the folder in memory holds that id as a number while the response comes back with it as a string, so the match failed and we ended up trying to update undefined — which is what reverted your change and threw that error.
I've fixed it by normalizing the id on both sides so it matches reliably, and added a guard so a miss like this can never blow up a save that already succeeded.

* The range errors has a fix on PR #1065 , because actually @slifty found it there. So the errors and the disabling of save will be the same both in modal and sidebar.

* Invalid range with qualifier
image

image

We can definitely have a range with qualifiers. The difference is the FE accepts the start date to have a qualifier and the end date to have time attached to it, while the BE does not accept this combination. My take would be just to leave it like this, because the user does get an error shown. Otherwise, I can look into it and see how to sync with the BE. But I feel that there might still be validation differences between BE and FE and not sure we could address them all.

NOTES

* Hm, there are ways to do this, none of them easy or straight forward. We would have to alter the route, save some data in the session storage or worst case scenario set a "selected" property on the record(FE+BE) changes. 

* No sure it's doable, because we would need to figure out the focus from multiple components. Also, the A11y way is focus the save button and hit enter, which atm works.

* Yes! We definitely need to figure out dates in the list and how to make them look good and also be sortable.

cecilia-donnelly (approved):
Hurray! Let's go to QA on this one. Thanks for the detailed response to my last comment.

aasandei-vsp:
@cecilia-donnelly Thank you for putting this up for QA. I have added a manual test for the edge case you found with the new folder.

omnignorant (approved):
**UAT: approved**

Tested on `app.dev.permanent.org` against branch `PER-10643-edtf-validation` (commit `5a8cd85`), deployed via Dev build and deploy run #1639, with the `edtf-date` flag on. Two independent passes covering the full QA Instructions list. Full report attached.

**The regression this PR targets is fixed.** Single-digit month and day entry was exercised across 13 scenarios on both the sidebar picker and the modal. A month entered as `2` validates the day against February, not January - `2026/2/30` and `2021/2/29` are both rejected with "That day does not exist in the selected month and year", `2026/1/31` is accepted, a lone `0` in either field is rejected and attributed to the correct field, and changing the month or year re-validates the day immediately. Values zero-pad on save and round-trip correctly.

**The folder bug from the July pass does not reproduce.** Created a new folder, set `1990-09`, changed it to `1990-07`: saves cleanly, no "could not update the item you selected" toast, no console error, persists across reload. Confirmed on both the sidebar and the modal.

**The approximate-start-on-a-range rejection is also gone.** Both `1975~/..` and `1975~/1980` save and round-trip without backend error.

Also confirmed working: clear date and time, Cancel discarding changes, Unknown, qualifier round-trips, out-of-range and invalid-character errors inline on both surfaces, time-without-a-date blocked on both, segment auto-advance and backspace navigation, the calendar picker, AM/PM/24H re-validation, midnight and noon boundaries, and the range preview strings. The in-scope file list change is correct - a cleared date leaves the column empty rather than falling back to `displayDT`.

**Findings for a follow-up PR, none blocking:**

- A date range with both sides empty saves and renders as "Sometime before ..". Save stays enabled.
- A blank end date allows save but does not persist, and Unknown on the end side is ignored. A blank start date blocks save when the end carries a time. The start and end paths validate differently.
- Typing `X` directly into a field is not accepted, though blanks are converted to `X` on save.
- Blank seconds save as `00` - confirm whether that is intended.
- The clock picker has no confirm button.
- Enter does not save from the sidebar (previously raised by @cecilia-donnelly as a non-blocking improvement).
- A failed inline save closes the picker and discards the typed entry.

Separately, "Invalid date" appears in the file list for qualified dates and also for Unknown - assigned to PER-10655, noting the Unknown case in case that ticket is scoped only to qualifiers.

Approving so this can merge. The items above are worth a follow-up ticket rather than another round here.

PR #1065, opened 2026-06-17
[PER-10415] Edtf date record viewer

aasandei-vsp:
# Manual test cases — EDTF date in the full-screen record view

**Setup:** log in with the `edtf-date` feature flag **on**, upload a file, click it to open the full-screen record view. The metadata panel now shows a **Date picker above the metadata table** (the old `Date` row inside the table is gone).

> [!WARNING]
> This PR also changes the **inline picker's commit behavior everywhere it appears** (file-list sidebar *and* the full-screen view): it no longer lets you attempt an invalid save and surface the error as a toast. It now shows a **general error message inside the picker and disables Save**, same gating as the modal. The "run in both places" tests from the PER-10643 test plan should be re-run with the *new* expectation below.

> [!NOTE]
> Field-specific messages (invalid characters, out-of-range month/day, day-for-month) still appear **inline below the offending input**. What's new is the picker-level line above the footer: the generic `The date entered is not valid. Please check the values and try again.` for anything with an inline message, and `A complete date is required when time is provided.` keeps its specific text (it has no inline field to attach to).

---

### Flag on
1. Open a record full-screen.
   - **EXPECTED:** The EDTF date picker renders above the metadata table. There is **no** `Date` row inside the metadata table.

### Flag off
1. Turn the `edtf-date` flag off and open a record full-screen.
   - **EXPECTED:** The old behavior: `Date` row inside the metadata table with the legacy inline edit, no EDTF picker.

---

## General error & Save gating in the inline picker ⭐

*(run in **both** the file-list sidebar picker and the full-screen view picker — same component, both surfaces must gate)*

### Invalid field value disables Save and shows the general error
1. Open the inline picker and enter month `13` (or day `32`).
   - **EXPECTED:** The field shows its specific message inline, **and** a red general error appears above the footer: `The date entered is not valid. Please check the values and try again.` The **Save button is disabled** (greyed out, not-allowed cursor).
2. Click Save anyway / press Enter.
   - **EXPECTED:** Nothing happens — no save request, no toast, the dropdown **stays open**.
3. Correct the value.
   - **EXPECTED:** The general error disappears and Save becomes enabled.

### Impossible calendar day
1. Enter year `2021`, month `02`, day `29`.
   - **EXPECTED:** `That day does not exist in the selected month and year.` inline below the day input, the generic error above the footer, Save disabled.

### Time without a complete date
1. Enter a time (e.g. `02:30 PM`) with the date fields empty or partial.
   - **EXPECTED:** The picker error shows the specific text `A complete date is required when time is provided.` (no inline field message for this one). Save is disabled; clicking it does nothing and the dropdown stays open.

### Valid value
1. Enter a complete valid date.
   - **EXPECTED:** No error line, Save enabled, save works as before.

---

### Inline picker save
1. In the full-screen view, open the picker, enter `1985` / `05` / `20`, Save.
   - **EXPECTED:** The picker closes and displays `1985-05-20`. No error toast.
2. Close the full-screen view and check the file list / sidebar, then reload the page and reopen the record.
   - **EXPECTED:** The value persists everywhere.

### More options modal
1. Enter a partial value inline (e.g. year `1990`), then click **More options**.
   - **EXPECTED:** The modal opens prefilled with what you typed inline.
2. Add a qualifier or a range in the modal and Save.
   - **EXPECTED:** The full-screen picker displays the saved value; it persists after reload.
3. Reopen the modal, add a different qualifier or range and Cancel.
   - **EXPECTED:** No change is saved; the stored value is untouched.

### Record without a date (displayDT fallback)
1. Open a freshly uploaded record that has never had a date set.
   - **EXPECTED:** The Date section in the sidebar does not show a date, but the message "Click to add date"

---

## Clearing the date

1. On a record with a saved date, open the picker, clear every field, Save.
   - **EXPECTED:** Save succeeds. The picker shows no date.
2. Reload and reopen the record.
   - **EXPECTED:** The date stays cleared — it does **not** fall back to any other date value.

---

## Failed save re-syncs the picker

1. On a record with date `2000-01-01`, go offline (DevTools → Network → Offline) or block the update request, then save `2010-06-15` from the inline picker.
   - **EXPECTED:** An error banner appears **and is fully visible on top of the full-screen view** (it used to render behind it). The picker **re-syncs back to `2000-01-01`** instead of keeping the unsaved `2010-06-15`.
2. Repeat via the More options modal.
   - **EXPECTED:** Same — banner visible, picker shows the stored value after the failure.
3. Go back online and save again.
   - **EXPECTED:** Works normally.

---

### Viewer access
1. Open a record shared with you with **Viewer** access (below Editor) full-screen.
   - **EXPECTED:** The date section is visible but **disabled** — it cannot be opened/edited.

### Unlisted share link
1. Open a record via an unlisted share link.
   - **EXPECTED:** Picker disabled.

### Public archive
1. Open a record in a public archive view.
   - **EXPECTED:** Picker disabled.

---

## Navigating between records in the viewer

1. With several files in a folder (different dates, including one with no date), open one full-screen and use the arrow keys / swipe to move to the next and previous records.
   - **EXPECTED:** The picker updates to each record's own date on every navigation — no stale value carried over from the previous record.

### DIFF
--- a/src/app/file-browser/components/file-viewer/file-viewer.component.ts
+++ b/src/app/file-browser/components/file-viewer/file-viewer.component.ts
@@ -453,8 +453,11 @@ export class FileViewerComponent implements OnInit, OnDestroy {
 	}
 
 	private updateDisplayTimeObject(): void {
-		const timeSource =
-			this.currentRecord?.displayTime || this.currentRecord?.displayDT;
+		const edtfDate = this.currentRecord;
+		const hasExplicitlyClearedDate = edtfDate?.displayTime === null;
+		const timeSource = hasExplicitlyClearedDate
+			? null
+			: edtfDate?.displayTime || edtfDate?.displayDT;
 		try {
 			this.displayTimeObject = timeSource
 				? this.edtfService.toDateTimeModel(timeSource)
@@ -481,7 +484,8 @@ export class FileViewerComponent implements OnInit, OnDestroy {
 
 	private async saveDisplayTime(result: DateTimeModel): Promise<void> {
 		try {
-			const newDisplayTime = this.edtfService.toEdtfDate(result);
+			const edtfDate = this.edtfService.toEdtfDate(result);
+			const newDisplayTime = edtfDate === '' ? null : edtfDate;
 			await this.onFinishEditing('displayTime', newDisplayTime);
 		} catch (err) {
 			this.message.showError({ message: err?.message });
--- a/src/app/file-browser/components/file-viewer/file-viewer.component.ts
+++ b/src/app/file-browser/components/file-viewer/file-viewer.component.ts
@@ -31,7 +31,13 @@ import { ShareLinksService } from '@root/app/share-links/services/share-links.se
 import { ApiService } from '@shared/services/api/api.service';
 import { FeatureFlagService } from '@root/app/feature-flag/services/feature-flag.service';
 import { environment } from '@root/environments/environment';
+import {
+	DateTimeModel,
+	EdtfService,
+} from '@shared/services/edtf-service/edtf.service';
+import { MessageService } from '@shared/services/message/message.service';
 import { TagsService } from '../../../core/services/tags/tags.service';
+import { EditDateTimeModalService } from '../edit-date-time-modal/edit-date-time-modal.service';
 
 @Component({
 	selector: 'pr-file-viewer',
@@ -63,6 +69,12 @@ export class FileViewerComponent implements OnInit, OnDestroy {
 
 	public canEdit: boolean;
 
+	public showEdtfDatePicker = false;
+
+	public editingDate: boolean = false;
+
+	public displayTimeObject: DateTimeModel | null = null;
+
 	// Swiping
 	private touchElement: HTMLElement;
 	private thumbElement: HTMLElement;
@@ -78,10 +90,10 @@ export class FileViewerComponent implements OnInit, OnDestroy {
 
 	// UI
 	public useMinimalView = false;
-	public editingDate: boolean = false;
 	private bodyScrollTop: number;
 	private itemTagsSubscription: Subscription;
 	private tagsSubscription: Subscription;
+	private dateModalSubscription?: Subscription;
 	private isUnlistedShare = true;
 
 	constructor(
@@ -89,6 +101,7 @@ export class FileViewerComponent implements OnInit, OnDestroy {
 		private route: ActivatedRoute,
 		private element: ElementRef,
 		private dataService: DataService,
+		private message: MessageService,
 		@Inject(DOCUMENT) private document: any,
 		public sanitizer: DomSanitizer,
 		private accountService: AccountService,
@@ -98,9 +111,13 @@ export class FileViewerComponent implements OnInit, OnDestroy {
 		private shareLinksService: ShareLinksService,
 		private api: ApiService,
 		private feature: FeatureFlagService,
+		private edtfService: EdtfService,
+		private editDateTimeModalService: EditDateTimeModalService,
 	) {
 		// store current scroll position in file list
 		this.bodyScrollTop = window.scrollY;
+
+		this.showEdtfDatePicker = this.feature.isEnabled('edtf-date');
 
 		const resolvedRecord = route.snapshot.data.currentRecord;
 		this.allTags = tagsService.getTags();
@@ -187,6 +204,7 @@ export class FileViewerComponent implements OnInit, OnDestroy {
 		});
 		this.itemTagsSubscription.unsubscribe();
 		this.tagsSubscription.unsubscribe();
+		this.dateModalSubscription?.unsubscribe();
 	}
 
 	private setRecordsToPreview(resolvedRecord: RecordVO) {
@@ -245,6 +263,7 @@ export class FileViewerComponent implements OnInit, OnDestroy {
 			this.replayUrl = this.getReplayUrl();
 		}
 		this.setCurrentTags();
+		this.updateDisplayTimeObject();
 	}
 
 	toggleSwipe(value: boolean) {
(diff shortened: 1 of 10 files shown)


src/app/file-browser/components/file-viewer/file-viewer.component.ts
```diff
@@ -244,6 +262,7 @@ export class FileViewerComponent implements OnInit, OnDestroy {
 			this.replayUrl = this.getReplayUrl();
 		}
 		this.setCurrentTags();
+		this.updateDisplayTimeObject();
```
slifty, line 266: We should have this behind `showEdtfDatePicker` flag too, right?  We don't want validation running that could render an edf error even though the picker isn't visible.

src/app/file-browser/components/file-viewer/file-viewer.component.ts
```diff
@@ -432,6 +451,54 @@ export class FileViewerComponent implements OnInit, OnDestroy {
...
+		try {
+			const edtfDate = this.edtfService.toEdtfDate(result);
+			const newDisplayTime = edtfDate === '' ? null : edtfDate;
+			await this.onFinishEditing('displayTime', newDisplayTime);
+		} catch (err) {
+			this.message.showError({ message: err?.message });
+		} finally {
+			// Recompute so the picker re-syncs to the stored value, whether the
+			// save came from the inline picker or the modal, and on failure too.
+			this.updateDisplayTimeObject();
```
slifty, line 498: Does this mean invalid input will get discarded (rather than the user being told it's invalid and having an opportunity to make a correction)?

src/app/file-browser/components/file-viewer/file-viewer.component.ts
```diff
@@ -432,6 +451,54 @@ export class FileViewerComponent implements OnInit, OnDestroy {
...
+				await this.saveDisplayTime(result);
+			}
+		});
+	}
+
+	private async saveDisplayTime(result: DateTimeModel): Promise<void> {
+		try {
+			const edtfDate = this.edtfService.toEdtfDate(result);
+			const newDisplayTime = edtfDate === '' ? null : edtfDate;
+			await this.onFinishEditing('displayTime', newDisplayTime);
```
slifty, line 492: This doesn't actually wait for the save to finish -- which means if the server returns an error, the value is reverted but the display would have already been updated below below in the `finally` block.

I believe we need to somehow detect when the server response has come back and then update the UX accordingly at that point.

cecilia-donnelly:
+1 to @slifty 's comments and I'll remove myself as a reviewer for now.

src/app/file-browser/components/file-viewer/file-viewer.component.ts
aasandei-vsp: Good catch — done. I put the guard inside updateDisplayTimeObject() itself so every caller is covered, and added a regression test: flag off + unparseable displayTime → no error shown.

src/app/file-browser/components/file-viewer/file-viewer.component.ts
aasandei-vsp: Good question — split by path:

- From the **modal**, invalid input can't be submitted at all: it validates the EDTF live and disables Save, showing the error inline.

- The **inline picker** did have this gap — it emitted without validating, so invalid input showed an error banner but got discarded. Fixed by giving it the same live validation as the modal (Save disabled + inline error), so invalid input can no longer be submitted from either surface.

- On **backend failure**, the value reverts and an error is shown — the same optimistic-update pattern every other field edit uses. Keeping the modal open until the BE responds would mean moving the save call into the modal, which I'd rather not couple; the finally re-sync just makes the picker reflect the reverted value.

src/app/file-browser/components/file-viewer/file-viewer.component.ts
aasandei-vsp: Good catch — onFinishEditing here wasn't awaiting saveItemVoProperty, so the re-sync ran before the server responded. Fixed by awaiting it (matching the sidebar component), so on failure the finally recompute now runs after EditService reverts the value, and the picker snaps back to the stored date alongside the error banner. No response-detection needed — EditService already reverts synchronously with the failed response.

aasandei-vsp:
> +1 to @slifty 's comments and I'll remove myself as a reviewer for now.

Very good points, I've addressed all of them.

slifty:
@aasandei-vsp I'm back from vacation -- I imagine we'll catch up on standup shortly about the status of this PR / if you're looking for another review!

cecilia-donnelly:
@slifty, I've been working with @aasandei-vsp on https://github.com/PermanentOrg/web-app/pull/1056, which I've just approved. This one probably needs conflicts updated because of our work over there, and then yes I expect it is ready for another round of review. Correct me if I'm wrong about any of that, @aasandei-vsp !

aasandei-vsp:
@cecilia-donnelly You are exactly right about the situation :D 
@slifty I have rebased this and it's now ready for review again. On top of the manual test cases mentioned in the comment of this PR, it's worth to maybe have a look at the tests from the [EDTF epic](https://permanent.atlassian.net/browse/PER-9355) as well.

src/app/file-browser/components/file-viewer/file-viewer.component.ts
```diff
@@ -433,11 +452,62 @@ export class FileViewerComponent implements OnInit, OnDestroy {
...
 
+	private updateDisplayTimeObject(): void {
+		if (!this.showEdtfDatePicker) {
+			return;
+		}
+		const edtfDate = this.currentRecord;
+		const hasExplicitlyClearedDate = edtfDate?.displayTime === null;
+		const timeSource = hasExplicitlyClearedDate
+			? null
+			: edtfDate?.displayTime || edtfDate?.displayDT;
```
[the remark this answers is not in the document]
slifty: Agree with this suggestion

slifty (approved):
Copilot made a var name suggestion I agree with -- approving so you don't get blocked on needing another review after that though, looks good!

aasandei-vsp:
> Copilot made a var name suggestion I agree with -- approving so you don't get blocked on needing another review after that though, looks good!

@slifty That is indeed much better, fixed! Thank you!

omnignorant (approved):
## UAT — PASS (18/18)

Tested against `PER-10415-edtf-date-record-viewer` @ `7a71bfa` on app.dev.permanent.org (Dev build and deploy #1641, dispatched for this pass — dev is still serving this branch).

Scope: all scenarios in the PR description, not just the Jira QA comment. Both halves of the feature flag exercised.

**Priority scenarios (⭐) — confirmed in both surfaces**, viewer picker and file-list sidebar picker, as the PR asks:
- Month 13 / day 32 → inline field message + picker-level "The date entered is not valid. Please check the values and try again." + Save `disabled: true`. Clicking it and pressing Enter do nothing; dropdown stays open, no toast.
- 2021-02-29 → "That day does not exist in the selected month and year." inline, generic error above the footer, Save disabled.
- Time with no date → "A complete date is required when time is provided." with no inline field message, Save disabled.

**Also confirmed:** flag-on placement (picker above the table, no Date row inside); inline and modal saves with persistence; modal prefill from inline entry; qualifier round-trip (`1990-06-15~` → "Date • Approximate: June 15, 1990"); Cancel discarding an edit; cleared dates staying cleared with no displayDT fallback; five consecutive chevron/arrow-key moves with no stale value.

**The two fixes riding along both hold up.** With the update request blocked, the error banner renders fully on top of the full-screen view, and the picker re-syncs to the stored value instead of keeping the rejected entry — reproduced from both the inline picker and the modal.

**Flag OFF** (toggled in Retool, then restored): the legacy Date row returns inside the metadata table, no EDTF control above it, zero `pr-sidebar-date-picker` elements on the page. The old inline editor opens and still saves and persists — worth calling out, since this PR rewrote the shared commit path that editor also uses.

**Permissions:** picker inert under Viewer access, on an unlisted share link, and in a public archive. Curator was used as a control — pencil present and working — so the gating is role-driven rather than share-wide.

**One observation, not a finding against this PR:** a record carrying a qualified EDTF value renders "Invalid date" in the file-list Date column. Already tracked as PER-10655 and recorded in the PR-1056 UAT report (scenarios 55/56); noting only that it reproduces from the full-screen viewer too.

`QA Instructions / qa-instructions` is red — not Required, fails routinely, flagging for visibility rather than as a blocker.

PR #1067, opened 2026-06-18
PER-10637: Show displayTime in file list

cecilia-donnelly:
Change the date pipe to optionally take `displayTime` and fall back to `displayDT`. This depends on https://github.com/PermanentOrg/back-end/pull/889.

### DIFF
--- a/src/app/file-browser/components/file-list-item/file-list-item.component.html
+++ b/src/app/file-browser/components/file-list-item/file-list-item.component.html
@@ -90,14 +90,10 @@
 		<div class="second-row">
 			@if (!showAccess || !item.ShareArchiveVO) {
 				<span>
-					{{
-						item.displayTime || item.displayDT
-							| prDate: item.TimezoneVO : 'date'
-					}}
+					{{ startDisplayTime | prDate: item.TimezoneVO : 'date' }}
 					@if (item.dataStatus > 0) {
 						<span @ngIfFadeInAnimation>{{
-							item.displayTime || item.displayDT
-								| prDate: item.TimezoneVO : 'time'
+							startDisplayTime | prDate: item.TimezoneVO : 'time'
 						}}</span>
 					}
 				</span>
@@ -116,18 +112,12 @@
 		<div class="date">
 			@if (item.dataStatus > 0) {
 				<div class="date-top" @ngIfFadeInAnimation>
-					{{
-						item.displayTime || item.displayDT
-							| prDate: item.TimezoneVO : 'date'
-					}}
+					{{ startDisplayTime | prDate: item.TimezoneVO : 'date' }}
 				</div>
 			}
 			@if (item.dataStatus > 0) {
 				<div class="date-bottom" @ngIfFadeInAnimation>
-					{{
-						item.displayTime || item.displayDT
-							| prDate: item.TimezoneVO : 'time'
-					}}
+					{{ startDisplayTime | prDate: item.TimezoneVO : 'time' }}
 				</div>
 			}
 		</div>
--- a/src/app/file-browser/components/file-list-item/file-list-item.component.html
+++ b/src/app/file-browser/components/file-list-item/file-list-item.component.html
@@ -90,10 +90,14 @@
 		<div class="second-row">
 			@if (!showAccess || !item.ShareArchiveVO) {
 				<span>
-					{{ item.displayDT | prDate: item.TimezoneVO : 'date' }}
+					{{
+						item.displayTime || item.displayDT
+							| prDate: item.TimezoneVO : 'date'
+					}}
 					@if (item.dataStatus > 0) {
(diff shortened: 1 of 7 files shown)


aasandei-vsp:
image
image

aasandei-vsp (dismissed):
I cannot approve this PR, because I have found some issues. Could you please have a look and check if I am on the right track with these, please?
There is a bug on FE, which I fixed(see latest commit). If a folder is added, the displayTime will be an interval and the pipe does not know how to show it. What I did was to extract the start date from the interval and then feed it to the pipe, so it shows correctly. DisplayDT is also like the start date for folders, so this approach seems correct to me. See the bug in the app in the image bellow:

Image

The second issue I found is related to sorting by date, which I cannot fix on FE, as we get the sorting from /navigateLean. The problem is that the sorting is still happening by displayDT and not by displayTime. This seems a difficult one to solve, as displayDT and displayTime do not seem to be in sync and we are also not sure which one we'll be showing in the list. See the image bellow:

Image

cecilia-donnelly:
Thanks @aasandei-vsp, looks like a great implementation for ranges! I have a back-end PR in progress for sorting: https://github.com/PermanentOrg/back-end/pull/892.

slifty (approved):
This looks good -- copilot flagged an edge case that can result in an empty element that may create a visual effect; I defer to @cecilia-donnelly on whether we want to block on that, so I'm approving!

cecilia-donnelly:
@slifty when doing this I didn't test all the possible EDTF types because I expect we'll do a second round of formatting before lifting the feature flag and letting people add uncertain dates. Currently no one can create a date without a time until we release the full feature. I don't think this behaves particularly well for e.g. year-only dates, either. I saw this PR more as a bugfix for the current state (where the date is mysteriously different from one place to the next) and didn't intend it to treat all the ways we'll need to format EDTF dates. I will add a ticket specifically about making sure we're doing formatting well here to the epic, though, so we don't forget! Does that make sense?

cecilia-donnelly:
The promised ticket: https://permanent.atlassian.net/browse/PER-10655

cecilia-donnelly:
My take on asking Claude to write a nontechnical list. It doesn't know that there's no way to clear date or set a range yet, so I removed several of the suggested tests:

## Test Setup

Before running these tests, prepare the following items in your archive. You will need edit access to all of them.

**Files to prepare:**

| Item | What to set up |
|---|---|
| **File A** | Upload any file. Do not edit the date at all — leave it as-is. |
| **File C** | Upload any file. In the info panel, set the **Date** field to a specific date and time in the past (e.g., June 10, 2020 at 6:58 AM). Do not set an end date. |

**Folders to prepare:**

| Item | What to set up |
|---|---|
| **Folder A** | Create a folder. Do not edit the date. |
| **Folder B** | Create a folder. In the info panel, set the **Start date** field to a date in the past (e.g., May 20, 1985) and set the **End date** field to a later date (e.g., June 15, 1990). |
| **Folder C** | Create a folder. In the info panel, set only the **Start date** field (e.g., May 20, 1985). Leave the **End date** blank. |

---

## Tests

### Test 1: File with no custom date shows the upload date

**Item:** File A (no custom date set)

**Steps:**
1. Navigate to the folder containing File A.
2. Look at the date shown for File A in the file list.

**Expected result:** The date shown is the date and time when the file was uploaded to the archive. It should include a time of day.

---

### Test 3: File with a single date and time shows the date and time

**Item:** File C (custom date set with a specific time)

**Steps:**
1. Navigate to the folder containing File C.
2. Look at the date shown for File C in the file list.

**Expected result:** The date and time shown match what you entered in the info panel.

---

### Test 7: Folder with no custom date shows the creation date

**Item:** Folder A (no custom date set)

**Steps:**
1. Navigate to the location containing Folder A.
2. Look at the date shown for Folder A in the file list.

**Expected result:** A date is shown for the folder (it should reflect when the folder was created or its default date). It should not be blank.

---

### Test 8: The file list updates immediately after editing a date

**Item:** Any file or folder you have edit access to.

**Steps:**
1. Click on the file or folder to select it and open the info panel.
2. Edit the **Date** (or **Start date**) field and save a new date.
3. Look at the file list without navigating away.

**Expected result:** The date shown in the file list row updates to match the new date you just entered.

---

### Test 10: Date display is consistent in both list view and grid/thumbnail view

**Item:** File D (start date and end date both set)

**Steps:**
1. Switch to **list view** and check the date shown for File D.
2. Switch to **grid view** (thumbnail view) and check the date shown for File D.

**Expected result:** In both views, only the start date appears. It is the same date in both views.

cecilia-donnelly:
All of the above look correct to me and I also confirmed that the date is correct on first page load.

cecilia-donnelly:
@omnignorant verbally approved this.

PR #1070, opened 2026-06-30
[PER-10593] Reset scroll to top on public folder navigation

aasandei-vsp:
Steps to check in ticket: https://permanent.atlassian.net/browse/PER-10593

### DIFF
--- a/src/app/file-browser/components/file-list/file-list.component.spec.ts
+++ b/src/app/file-browser/components/file-list/file-list.component.spec.ts
@@ -168,6 +168,20 @@ describe('FileListComponent', () => {
 		});
 	}));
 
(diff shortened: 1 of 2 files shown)


cecilia-donnelly (approved):
Makes sense to me! I tried it and it worked.

omnignorant (approved):
UAT ✅ — verified on dev (Dev build and deploy #1596, branch PER-10593-gallery-scroll). Navigating into a public sub-folder from a scrolled-down position now resets scroll to the top of the page. Confirmed in both Chrome and Safari on current macOS — Safari being the environment the regression was originally reported in — repeated across several attempts, consistent. Approving from a product/UAT standpoint.

PR #1084, opened 2026-07-10
[PER-10633] Replace hardcoded colors in global styles and generic presentational component's styles

aasandei-vsp:
This PR only replaces hardcoded colors in global styles and generic presentational components with variables that have identical values, so no visible change.
What I did do was visually check the app where the hardcoded colors have been replaced by variables.
It is tedious and I do not believe it needs to be done multiple times.

No QA needed.


cecilia-donnelly (approved):
Thank you @aasandei-vsp !

PR #1085, opened 2026-07-13
[PER-10633] Replace hardcoded colors for file-browser components and generic elements from the same main screen

aasandei-vsp:
This PR only replaces hardcoded colors for file-browser components and generic elements from the same main screen with variables that have identical values, so no visible change.
What I did do was visually check the app where the hardcoded colors have been replaced by variables.
It is tedious and I do not believe it needs to be done multiple times.

No QA needed.

### DIFF
--- a/src/app/file-browser/components/sharing-dialog/sharing-dialog.component.scss
+++ b/src/app/file-browser/components/sharing-dialog/sharing-dialog.component.scss
@@ -14,9 +14,9 @@ select.form-select {
 
 .dialog-body {
 	overflow-y: visible !important;
-	background-color: white;
+	background-color: $PR-brand-white;
 	border-radius: 0.3rem;
-	box-shadow: 0 0.5rem 1rem #0000004d;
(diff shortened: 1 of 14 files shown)


src/app/file-browser/components/sharing-dialog/sharing-dialog.component.scss
```diff
@@ -14,9 +14,9 @@ select.form-select {
 
 .dialog-body {
 	overflow-y: visible !important;
-	background-color: white;
+	background-color: $PR-brand-white;
 	border-radius: 0.3rem;
-	box-shadow: 0 0.5rem 1rem #0000004d;
+	box-shadow: 0 0.5rem 1rem rgba($PR-brand-black, 0.3);
```
cecilia-donnelly, line 19: This is new to me! Thanks.

cecilia-donnelly (commented):
Looks good!

cecilia-donnelly (approved):
Officially approving, oops.

PR #1092, opened 2026-07-20
[PER-10476] [5] Replace navigateMin with getWithChildren in folder picker

aasandei-vsp:
Migrate the folder picker off /folder/navigateMin, its last consumer, and remove FolderRepo.navigate. The root.root folder stays on getRoot since Stela does not serve the virtual root.

Issue: PER-10476

# Should not be merged before #1091 

# Manual test cases — Replace navigateMin with getWithChildren

**Setup:** log in on an archive with nested folders and at least one record. The folder picker opens from an item's Move, Copy, and Choose-file actions.
**EXPECTED:** the folder picker now loads folders through the Stela `getWithChildren` endpoint, except the top-level root which still loads through `getRoot`.

---

## Move dialog

1. Select an item and choose Move.
   - **EXPECTED:** The picker opens at the root, listing My Files and Public (not Apps or Vault), and the header shows the root folder name.
2. Click into My Files, then into a subfolder.
   - **EXPECTED:** Each folder's subfolders load and the header name updates.
3. Press Back repeatedly up to the top.
   - **EXPECTED:** The root reloads with My Files and Public — no blank or errored panel.
4. Select a destination folder and confirm the Move.
   - **EXPECTED:** The item moves to the chosen folder and the dialog closes.

## Copy dialog

1. Select an item and choose Copy.
   - **EXPECTED:** The picker opens at the root with the same folders as above.
2. Navigate into a folder, select it, and confirm the Copy.
   - **EXPECTED:** The item is copied to the chosen folder and the dialog closes.

## Choose-file dialog

1. Open the record chooser (Choose file).
   - **EXPECTED:** Both folders and records are listed (records are not stripped).
2. Navigate into a folder and select a record.
   - **EXPECTED:** The record can be selected and confirmed.

## Filtered move

1. Move a folder into the tree.
   - **EXPECTED:** The folder being moved does not appear as a destination option.

## Error handling

1. With the picker open, go offline (DevTools > Network > Offline) and navigate into a folder.
   - **EXPECTED:** An error toast appears — no blank panel and no uncaught error in the console.

### DIFF
--- a/src/app/core/components/folder-picker/folder-picker.component.spec.ts
+++ b/src/app/core/components/folder-picker/folder-picker.component.spec.ts
@@ -4,14 +4,44 @@ import { cloneDeep, some } from 'lodash';
 
 import { DataService } from '@shared/services/data/data.service';
 import { ApiService } from '@shared/services/api/api.service';
+import { MessageService } from '@shared/services/message/message.service';
 import { FolderResponse } from '@shared/services/api/index.repo';
 import { SharedModule } from '@shared/shared.module';
 import { FolderVO } from '@root/app/models';
 import { HttpTestingController } from '@angular/common/http/testing';
 import { FolderPickerService } from '@core/services/folder-picker/folder-picker.service';
 import { DataStatus } from '@models/data-status.enum';
-import { of } from 'rxjs';
 import { FolderPickerComponent } from './folder-picker.component';
+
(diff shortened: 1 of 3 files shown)


aasandei-vsp:
This has been fixed on PR #1132

PR #1098, opened 2026-07-22
[PER-10569] Show downtime banner for scheduled maintenance

aasandei-vsp:
Add an announcement event to display a site-wide maintenance banner during the scheduled downtime on July 28th, 1:00-3:00 PM CDT. Reuses the existing time-window-driven Announcement mechanism, so the banner appears and disappears automatically for the window; no feature flag or component changes needed.

Issue: PER-10569

The way I tested this was just to modify the start/end dates locally in the code.

- I changed the start date to be 1 min before the current time, to see that the banner appears.

- Then I modified the end date to be 1 hour in the past, to make sure the banner won't appear after the window is finished.

It is worth mentioning that this banner only appears on a few screens:
image

For example, it does not appear when you go to an archive's profile:
image

### DIFF
--- a/src/app/announcement/data/events.ts
+++ b/src/app/announcement/data/events.ts
@@ -2,6 +2,12 @@ import { AnnouncementEvent } from '../models/announcement-event';
 
 export const ANNOUNCEMENT_EVENTS: AnnouncementEvent[] = [
 	{
+		start: new Date().getTime(),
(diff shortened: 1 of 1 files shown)


cecilia-donnelly (approved):
I changed the exact wording and the start time, @aasandei-vsp, so I will still leave it to you to have a final look before merging. Thank you for finding where this belongs and making the change! Sorry for the unclear ticket.

PR #1099, opened 2026-07-23
[PER-10667] Align public gallery date with detail view

aasandei-vsp:
The public gallery card rendered its date from displayDT (mapped from the API's displayDate), while the detail sidebar reads displayTime || displayDT. For items where those fields diverge, the gallery showed a different date than the detail panel.

Base the public-archive date on the existing startDisplayTime getter (displayTime with EDTF interval-start handling, falling back to displayDT) so the gallery matches the sidebar's field preference. Add spec coverage for the displayTime, displayDT-fallback, and EDTF-interval cases.

Issue: PER-10667

The dates I have found as being inconsistent with the rest of the app was in the gallery, so I have changed it to displayTime/displayDT:
image

The dates in the public folder behave as expected, like in private or shared.
Also, we are not actually supporting a friendly timezone as of now, but we will once the EDTF date will be released.

### DIFF
--- a/src/app/file-browser/components/file-list-item/file-list-item.component.spec.ts
+++ b/src/app/file-browser/components/file-list-item/file-list-item.component.spec.ts
@@ -389,4 +389,34 @@ describe('FileListItemComponent', () => {
 
 		expect(secondRowDate).toContain('2023-01-01T00:00:00.000Z');
 	});
+
+	it('should base the public-archive date on displayTime when it is set', async () => {
+		component.item.displayTime = '2020-06-10';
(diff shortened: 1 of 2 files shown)


cecilia-donnelly (approved):
Thank you!

omnignorant (approved):
UAT ✅ — verified on dev (Dev build and deploy #1597, branch PER-10667-gallery-items-date). Reproduced the pre-fix mismatch — the public gallery card showed a different date than the item's detail view once the editable date field diverged from the stored value — then confirmed after deploy that the gallery card date now matches the detail view. Approving from a product/UAT standpoint.

